PR #3, opened 2018-02-08
Exit with an error if an error occured

alexcarruthers:
For use in CI environments to make sure our swagger file actually generates.


gkoberger:
Whoops, I let this sit so long that everything was rewritten... sorry!

I've gone ahead and made the changes here: https://github.com/readmeio/oas/commit/8156788138e4e7eba94e1ca2c6a096217ac699cd

In the future, I promise I'll be better at getting to this sooner. Thanks @alexcarruthers !

PR #4, opened 2018-03-14
`generate` without colors

alansikora:
Hello!

I'm using this package and a colorized `generate` breaks my json files. And currently, it is the only way to regenerate a file without running `init` all the time.

So I added a new `print` method to pretty print the output and kept `generate` as simple as possible.

What do you think?


gkoberger:
I ran into this problem today! Let's do it!

PR #95, opened 2020-01-22
feature/metrics-log-annotation

gratcliff:
Log decoration - preliminary work:

- Building out 'findOperation' class method on OAS. Enables ingestion of an log, and combines it with the existing OAS spec to determine the target path

### DIFF
--- a/__tests__/fixtures/local-link.json
+++ b/__tests__/fixtures/local-link.json
@@ -1,0 +1,369 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Link Example",
+    "version": "1.0.0"
+  },
+  "servers": [
+    {
+      "url": "http://local-link.com"
+    }
+  ],
+  "paths": {
+    "/2.0/users/{username}": {
+      "get": {
+        "operationId": "getUserByName",
+        "security": [
+					{
+						"cookieAuth": []
+					},
+					{
+						"bearerAuth": []
+					}
+				],
+        "parameters": [
+          {
+            "name": "username",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "The User",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/user"
+                }
+              }
+            },
+            "links": {
+              "userRepositories": {
+                "$ref": "#/components/links/UserRepositories"
+              }
+            }
+          }
+        }
+      }
+    },
+    "/2.0/repositories/{username}": {
+      "get": {
+        "operationId": "getRepositoriesByOwner",
+        "parameters": [
+          {
+            "name": "username",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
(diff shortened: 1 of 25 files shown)


src/oas.js
```diff
@@ -101,6 +102,80 @@ function normalizedUrl(oas) {
   return ensureProtocol(url);
 }
 
+function normalizePath(path) {
+  const curlBacketMatch = /{(.*?)}/;
+
+  return path
```
dok: How about something like this?

```js
'/users/{user_id}/filter_sets/{filter_set_id}'.replace(/{(.*?)}/g, ':$1')
```

Then we can same a little bit of time on iteration

src/oas.js
```diff
@@ -128,6 +203,23 @@ class Oas {
     const operation = getPathOperation(this, { swagger: { path }, api: { method } });
     return new Operation(this, path, method, operation);
   }
+
+  findOperation(logInput) {
+    const { url, method } = logInput.request.log.entries[0].request;
```
dok: We're gonna move away from referencing `.request.log.entries` (we've "exploded" the request object in the HAR object to the root level of the mongo document)!  Let's use `logInput.url` & `logInput.method`

src/oas.js
```diff
@@ -101,6 +102,80 @@ function normalizedUrl(oas) {
...
+    .map(p => {
+      const pathMatch = curlBacketMatch.exec(p);
+      if (pathMatch) return `:${pathMatch[1]}`;
+      return p;
+    })
+    .join('/');
+}
+
+function generatePathMatches(paths, pathName, origin) {
+  return Object.keys(paths)
```
dok: What happens when there are multiple matches here? Can there be multiple matches in this step?

src/oas.js
gratcliff: Nice, that's a good improvement!

src/oas.js
gratcliff: Has that been updated in that repo yet?

src/oas.js
gratcliff: " it deals with multiple matches by prioritizing the match with the least path Parmas. Those wildcards really are the ones that would contribute to finding dups 
If there are other dups (like repeat paths) that would just be an incorrect oas file to begin with" via our slack

src/oas.js
gratcliff: Good catch, I'll swap that out! Thanks @dok

erunion: I'd also suggest supplying `findOperation` with the url and method upfront. This package shouldn't need to know about how we store logs in the DB, just that it needs to look for an operation that match a specific pattern.

gratcliff: Good call @erunion, i'll move it to the FE.

__tests__/fixtures/local-link.json
```diff
@@ -0,0 +1,369 @@
...
+    {
+      "url": "http://local-link.com"
+    }
+  ],
+  "paths": {
+    "/2.0/users/{username}": {
+      "get": {
+        "operationId": "getUserByName",
+        "security": [
+					{
```
erunion: Can you fix the weird indents going on here?

PR #96, opened 2020-01-23
feature/get-headers

gratcliff:
Functionality improvements for retrieving headers of a particular OAS file based on required instantiation arguments for the `Operation` class. 

Takes into account headers in security, request params, and response to return a holistic view of the headers in that file.

### DIFF
--- a/__tests__/oas.test.js
+++ b/__tests__/oas.test.js
@@ -3,280 +3,362 @@ const { Operation } = require('../src/oas');
 const petstore = require('./fixtures/petstore.json');
 const multipleSecurities = require('./fixtures/multiple-securities.json');
 
-describe('operation()', () => {
-  it('should return an operation object', () => {
-    const oas = { paths: { '/path': { get: { a: 1 } } } };
-    const operation = new Oas(oas).operation('/path', 'get');
-    expect(operation).toBeInstanceOf(Operation);
-    expect(operation.a).toBe(1);
-    expect(operation.path).toBe('/path');
-    expect(operation.method).toBe('get');
+describe('class.Oas', () => {
+  describe('operation()', () => {
+    it('should return an operation object', () => {
+      const oas = { paths: { '/path': { get: { a: 1 } } } };
+      const operation = new Oas(oas).operation('/path', 'get');
+      expect(operation).toBeInstanceOf(Operation);
+      expect(operation.a).toBe(1);
+      expect(operation.path).toBe('/path');
+      expect(operation.method).toBe('get');
+    });
+
+    it('should return a default when no operation', () => {
+      expect(new Oas({}).operation('/unknown', 'get')).toMatchSnapshot();
+    });
   });
 
-  it('should return a default when no operation', () => {
-    expect(new Oas({}).operation('/unknown', 'get')).toMatchSnapshot();
-  });
-});
-
-test('should remove end slash from the server URL', () => {
-  expect(new Oas({ servers: [{ url: 'http://example.com/' }] }).url()).toBe('http://example.com');
-});
-
-test('should default missing servers array to example.com', () => {
-  expect(new Oas({}).url()).toBe('https://example.com');
-});
-
-test('should default empty servers array to example.com', () => {
-  expect(new Oas({ servers: [] }).url()).toBe('https://example.com');
-});
-
-test('should default empty server object to example.com', () => {
-  expect(new Oas({ servers: [{}] }).url()).toBe('https://example.com');
-});
-
-test('should add https:// if url starts with //', () => {
-  expect(new Oas({ servers: [{ url: '//example.com' }] }).url()).toBe('https://example.com');
-});
-
-test('should add https:// if url does not start with a protocol', () => {
-  expect(new Oas({ servers: [{ url: 'example.com' }] }).url()).toBe('https://example.com');
-});
-
-describe('server variables', () => {
-  it('should use defaults', () => {
-    expect(
-      new Oas({
-        servers: [{ url: 'https://example.com/{path}', variables: { path: { default: 'path' } } }],
-      }).url()
-    ).toBe('https://example.com/path');
+  it('should remove end slash from the server URL', () => {
+    expect(new Oas({ servers: [{ url: 'http://example.com/' }] }).url()).toBe('http://example.com');
   });
 
-  it('should use user variables over defaults', () => {
-    expect(
-      new Oas(
-        {
-          servers: [{ url: 'https://{username}.example.com', variables: { username: { default: 'demo' } } }],
-        },
-        { username: 'domh' }
-      ).url()
-    ).toBe('https://domh.example.com');
+  it('should default missing servers array to example.com', () => {
+    expect(new Oas({}).url()).toBe('https://example.com');
   });
 
-  it('should fetch user variables from keys array', () => {
-    expect(
-      new Oas(
-        {
-          servers: [{ url: 'https://{username}.example.com', variables: { username: { default: 'demo' } } }],
-        },
-        { keys: [{ name: 1, username: 'domh' }] }
-      ).url()
-    ).toBe('https://domh.example.com');
+  it('should default empty servers array to example.com', () => {
+    expect(new Oas({ servers: [] }).url()).toBe('https://example.com');
   });
 
-  it.skip('should fetch user variables from selected app', () => {
-    expect(
-      new Oas(
-        {
-          servers: [{ url: 'https://{username}.example.com', variables: { username: { default: 'demo' } } }],
-        },
-        {
-          keys: [
-            { name: 1, username: 'domh' },
-            { name: 2, username: 'readme' },
-          ],
-        },
-        2
-      ).url()
-    ).toBe('https://readme.example.com');
+  it('should default empty server object to example.com', () => {
+    expect(new Oas({ servers: [{}] }).url()).toBe('https://example.com');
   });
 
-  // Test encodeURI
-  it('should pass through if no default set', () => {
-    expect(new Oas({ servers: [{ url: 'https://example.com/{path}' }] }).url()).toBe('https://example.com/{path}');
+  it('should add https:// if url starts with //', () => {
+    expect(new Oas({ servers: [{ url: '//example.com' }] }).url()).toBe('https://example.com');
   });
-});
 
-test('should be able to access properties on oas', () => {
-  expect(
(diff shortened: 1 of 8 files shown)


gratcliff:
> Can you add some tests?

Can do! I think I accidentally marked this as ready! Tests are currently a WIP.

__tests__/oas.test.js
```diff
@@ -3,280 +3,362 @@ const { Operation } = require('../src/oas');
...
-    expect(Object.keys(operation.prepareSecurity())).toHaveLength(0);
-  });
+      const { logOperation } = oas.findOperation(uri, method);
+      const operation = new Operation(oas, logOperation.url.path, logOperation.url.method, logOperation.operation);
 
-  it('should return empty if security scheme type doesnt exist', () => {
-    const operation = new Oas(multipleSecurities).operation('/unknown-auth-type', 'post');
-    expect(Object.keys(operation.prepareSecurity())).toHaveLength(0);
+      expect(operation.getHeaders()).toMatchObject({
+        request: ['apiKey'],
```
erunion: Based on the security scheme `type` and `scheme`, we should match them up to the header that would end up serving that (eg. `apiKey` could either match up with the `Authentication` header, or it could sometimes be supplied as a query param and shouldn't be returned here).

There's some work on handling the response from `prepareSecurity` in the explorer that might help:

* https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/Response.jsx
* https://github.com/readmeio/api-explorer/blob/master/packages/api-explorer/src/SecurityInput.jsx

The OAS docs on schemes might help as well https://swagger.io/docs/specification/authentication/#securitySchemes

dok:
LGTM! Thanks for the tests!

src/oas.js
```diff
@@ -137,34 +162,39 @@ function filterPathMethods(pathMatches, targetMethod) {
...
       return undefined;
     })
     .filter(p => p);
 }
 
-function findTargetPath(pathMatches) {
+function findTargetPath(pathMatches, index) {
   let minCount = Object.keys(pathMatches[0].url.slugs).length;
-  let candidate;
+  let logOperation;
```
erunion: Let's rename this to `operation` and not have any mention of logs in this package.

```suggestion
  let operation;
```

src/oas.js
```diff
@@ -137,34 +162,39 @@ function filterPathMethods(pathMatches, targetMethod) {
...
       minCount = paramCount;
-      candidate = selection;
+      logOperation = selection;
     }
   }
 
-  return candidate;
+  return {
+    logOperation,
+    index,
```
erunion: What's the index here for? Seems like it'll always be 0.

src/oas.js
```diff
@@ -195,20 +225,22 @@ class Oas {
...
-    const annotatedPaths = generatePathMatches(paths, pathname, origin);
+    const [, pathName] = url.split(targetServer.url);
+    const annotatedPaths = generatePathMatches(paths, pathName, targetServer.url);
     if (!annotatedPaths.length) return undefined;
 
     const includesMethod = filterPathMethods(annotatedPaths, method);
     if (!includesMethod.length) return undefined;
 
-    return findTargetPath(includesMethod);
+    return findTargetPath(includesMethod, index);
```
erunion: `findOperation` should probably pull back the spec of the operation instead.

src/oas.js
gratcliff: The index is passed in by our FE based on the loop it does to instantiate new instances of the `oas` class. Based on our FE filtering, we need to identify the target file where the path was found and use it for generating an `operation` instance. It defaults to 0.

__tests__/oas.test.js
```diff
@@ -3,280 +3,362 @@ const { Operation } = require('../src/oas');
...
+      expect(res).toBeUndefined();
+    });
+
+    it('should return a a result if found', () => {
+      const oas = new Oas(petstore);
+      const uri = `http://petstore.swagger.io/v2/pet/1`;
+      const method = 'DELETE';
+
+      const res = oas.findOperation(uri, method);
+      console.log(JSON.stringify(res));
```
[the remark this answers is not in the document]
gratcliff: Good catch!

__tests__/oas.test.js
```diff
@@ -3,280 +3,362 @@ const { Operation } = require('../src/oas');
...
+    it('should return undefined if no matching methods in path', () => {
+      const oas = new Oas(petstore);
+      const uri = `http://petstore.swagger.io/v2/pet/1`;
+      const method = 'PATCH';
+
+      const res = oas.findOperation(uri, method);
+      expect(res).toBeUndefined();
+    });
+
+    it('should return a a result if found', () => {
```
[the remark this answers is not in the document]
gratcliff: Good catch!

src/oas.js
gratcliff: I have mixed feelings about this, because we need to identify a target operation in our FE. Is there another agnostic name that you can suggest?

src/oas.js
gratcliff: Instead of returning the index and handling it in the front end, return the whole OAS file in this payload?

__tests__/oas.test.js
gratcliff: Thanks Jon, I'll put some work into this!

src/oas.js
erunion: I was originally thinking a few weeks ago that running `findOperation('/pets/1`, 'get')` would return back everything in the OAS under `paths[/pets].get` and the FE could use that.

gratcliff: It is doing that already! Per our discussion I've migrated the index to our main repo.

src/oas.js
gratcliff: I've ripped this out and we're maintaining it in our main repo.

src/oas.js
gratcliff: This isn't necessary anymore! It's been removed entirely.

__tests__/oas.test.js
gratcliff: @erunion I've updated this, please take another look. I'll finish tests for additions before merge.

__tests__/oas.test.js
```diff
@@ -3,280 +3,361 @@ const { Operation } = require('../src/oas');
...
+      const method = 'DELETE';
 
-  // TODO We dont currently support cookies?
-  it.todo('apiKey/cookie: should return with a type of Cookie');
+      const logOperation = oas.findOperation(uri, method);
+      const operation = new Operation(oas, logOperation.url.path, logOperation.url.method, logOperation.operation);
 
-  it.todo('should throw if attempting to use a non-existent scheme');
+      expect(operation.getHeaders()).toMatchObject({
+        request: ['api_key'],
```
erunion: This still returning `api_key` instead of `Authorization`?

gratcliff: Correct. Per their docs, it looks like scheme types of `apiKey` come in as unique headers: https://swagger.io/docs/specification/authentication/api-keys/

erunion: Shouldn't they still have an `in` that represents the location of where it's sent?

gratcliff: Well we're only concerned with headers, and it's being filtered as such. If it's added as a querystring that's not something should be returned or omitted from our log view.

erunion: I'm just confused why it's being returned as `api_key`. From my understanding of the FE implementation for this, we want to segment out headers in a response that aren't stored in their OAS, and to do that we'll need to know the name of the header (`Authorization` in this case). Pulling back `api_key` is going to require another lookup to see what `api_key` matches up to.

gratcliff: We should be returning the name of the api key, not `api_key` itself. There shouldn't be a need to find out what it's attached to after being served. I believe in relation to the test, it's named `api_key` in the petstore.json fixture for the `DELETE` method. Would you like me to update that value so that it doesn't seem conflated?

dok:
One other thing I noticed: 
Screen Shot 2020-01-24 at 9 22 42 AM

If there is a `/` at the end of a URL, it won't match!

gratcliff:
> One other thing I noticed:
> Screen Shot 2020-01-24 at 9 22 42 AM
> 
> If there is a `/` at the end of a URL, it won't match!

There were a couple issues with the base branch's matching. I've switched to a regex exec in this branch to cover that base. You do make a good point though! I'll need to write some more tests for the `getheaders` update. I'll cover this as well!

dok:
One other aspect to note: what if the base URL is different, this is the case for when someone toggles from live data to development data, the URL may shift from `metrics.readme.io` to `readme.local:3003`

gratcliff:
> One other aspect to note: what if the base URL is different, this is the case for when someone toggles from live data to development data, the URL may shift from `metrics.readme.io` to `readme.local:3003`

Good point. The base `oas` operation will return undefined as the servers do not match. As a result, we won't filter any headers in `readme`.

dok:
Thanks for looking into those. Because I'm actively using this now, I'm finding a few more issues that we should work through.

Screen Shot 2020-01-24 at 10 13 52 AM
I noticed that the `.origin` is returning a little extra piece of the path there

gratcliff:
> Thanks for looking into those. Because I'm actively using this now, I'm finding a few more issues that we should work through.
> 
> Screen Shot 2020-01-24 at 10 13 52 AM
> 
> I noticed that the `.origin` is returning a little extra piece of the path there

It depends on the server url: https://github.com/readmeio/oas/pull/96/files#diff-5081e0d03920d7a2e9a6b7e7cf2f179aR242

erunion:
@dok That's happening because `https://dash.readme.io/api/v1` is set as our server URL, so technically to OAS `/api-specification` is the path.

dok:
@gratcliff @erunion oh gotcha. Thanks for the clarification

src/oas.js
```diff
@@ -2,6 +2,7 @@
 const { pathToRegexp, match } = require('path-to-regexp');
 const getPathOperation = require('./lib/get-path-operation');
 const getUserVariable = require('./lib/get-user-variable');
+const getReference = require('./lib/get-reference');
```
erunion: You should be able to use `findSchemaDefinition` in `utils.js` for this instead. Takes two args:

```
findSchemaDefinition(ref, oas);
```

gratcliff: I saw that and it didn't look to map, but you are right, I can swap this out!

PR #105, opened 2020-01-29
Bug Fix - Query Params

gratcliff:
Query params were incorrectly being used to generate path match regex. Query strings are now being properly pruned for matching purposes


gratcliff:
> lgtm, but can you add a test that has a query param in the path?

Added!

PR #110, opened 2020-02-21
Updating our handling of JSON Schema to be compliant against allowEmptyValue

erunion:
This resolves some issues with how we're constructing JSON Schema with regard to empty `default` properties.

Previously when an empty `default` was added into a data schema, a bug would cascade up into `@readme/api-explorer` and our usage of `@readme/oas-to-har` and we'd create an API HAR of a payload consisting of empty strings.

Instead now, if a `default` evaluates to an empty string, and `allowEmptyValue: true` is present on that property, we'll let the empty value through, otherwise it'll get tossed.

Screen Shot 2020-02-21 at 12 03 18 PM

* [x] 🧰 Fixes to our construction of parameter and requestBody JSON Schema work to be compliant on `allowEmptyValue`. See https://swagger.io/specification/#parameterObject for details on the property.
* [x] 🚧 Rewrites handling of common common parameters to make the code less obtuse and easier to maintain.
* [x] 🌱 Added a shitload of JSON Schema unit tests.


packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -15,42 +17,110 @@ const types = {
...
+    if ('parameters' in pathOperation.oas.paths[path]) {
+      return pathOperation.oas.paths[path].parameters;
+    }
+  }
+
+  return [];
 }
 
 function getOtherParams(pathOperation, oas) {
-  let pathParameters = pathOperation.parameters || [];
```
erunion: Renamed this from `pathParameters` to `operationParameters` because in the context of what this is doing, it's building up an array of every parameter on the option; not an array of every **path** parameter on the option.

<sub>i spent a good number of minutes trying to hunt down some weird edge cases when writing tests before realizing this is what was going on 🙃 </sub>

erunion:
@rafegoldberg @gratcliff This is good for review.. Happy to go over it with you if you've got questions about WTF this is.

gratcliff:
@jon It's hard to review this without knowing exactly what the outcome of the data transformation should look like. 

There don't appear to be any glaring issues - data checks are present so nothing should fail. There is a lot of iteration and the recursion scares me a tad, but overall I think that's a necessity with the schemas you're trying to process.

PR #111, opened 2020-02-22
feat: migrating the repository into a lerna-powered monorepo architecture

erunion:


packages/cli/index.js
```diff
@@ -0,0 +1,3 @@
+const Oas = require('../tooling');
```
erunion: This should be updated to `@readme/oas-toolkit`.

packages/cli/utils.js
```diff
@@ -0,0 +1,3 @@
+const utils = require('../tooling/utils');
```
erunion: This should be updated to `@readme/oas-toolkit/utils`.

PR #114, opened 2020-02-27
fix: resolving a null assertion issue that broke jsonschema parsing

erunion:
Under certain circumstances where an object property would evaluate to `null`, we'd attempt to recurse down into it like an object because `typeof null === 'object`.

I've also added ESLint, Prettier, and our shared ESLint config into the root `package.json` because without it, VSCode wasn't able to find the executables.

### DIFF
--- a/packages/tooling/src/lib/parameters-to-json-schema.js
+++ b/packages/tooling/src/lib/parameters-to-json-schema.js
@@ -21,7 +21,11 @@ function getBodyParam(pathOperation, oas) {
 
   const cleanupSchemaDefaults = obj => {
     Object.keys(obj).forEach(prop => {
-      if (typeof obj[prop] === 'object') {
+      if (obj[prop] === null) {
+        // If the item is null, just carry on. Why do this in addition to `typeof obj[prop] == object`? Because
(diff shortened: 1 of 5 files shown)


packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -21,7 +21,11 @@ function getBodyParam(pathOperation, oas) {
 
   const cleanupSchemaDefaults = obj => {
     Object.keys(obj).forEach(prop => {
-      if (typeof obj[prop] === 'object') {
+      if (obj[prop] === null) {
+        // If the item is null, just carry on. Why do this in addition to `typeof obj[prop] == object`? Because
+        // `typeof null` equates to `object` for "legacy reasons" apparently.
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null
```
rafegoldberg: lololol, Javascript

> ...because `typeof null` equates to `object` [for "legacy reasons"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null)

🤣🔫

PR #116, opened 2020-02-27
feat: better support for polymorphism

erunion:
### find-schema-definition
* [x]  Slightly rewriting the tool so it can handle `$ref` pointers that are escaped with `~1` by just using the great [jsonpointer](https://www.npmjs.com/package/jsonpointer) library. 
  * This is mostly the same code that RJSF uses, just without the recursion to call `find-schema-definition` if we find a nested `$ref`. We handle that case in the code that handles `find-schema-definition` in the Explorer, and though they're probably ripe for a rewrite it's fine for now.

### flatten-schema
* [x] Added support for flattening a schema that contains `oneOf`, `allOf`, and `anyOf` polymorphism cases.
  * For `oneOf` and `anyOf` cases, since we can't really flatten down every option defined in those arrays into a single schema that would validate against the document, I'm choosing to just select the first defined object in the set and flatten that out.
* [x] Making a minor adjustment to only attempt to flatten an object if it's actually an object. Previously, we'd throw up a fatal exception on a schema that looks like this:

```json
"application/xml": {
  "schema": "~paths~/pet~post~requestBody~content~application/json~schema"
}
```

### parameters-to-json-schema
* [x] Made updates to stop storing our component schemas inside of a `definitions` object as a recent push to RJSF made this no longer a requirement.

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": ["@commitlint/config-conventional"]
+}
--- a/packages/tooling/__tests__/lib/find-schema-definition.test.js
+++ b/packages/tooling/__tests__/lib/find-schema-definition.test.js
@@ -1,4 +1,21 @@
+const petstore = require('@readme/oas-examples/3.0/json/petstore.json');
 const findSchemaDefinition = require('../../src/lib/find-schema-definition');
+
+test('should return a definition for a given ref', () => {
+  expect(findSchemaDefinition('#/components/schemas/Pet', petstore)).toStrictEqual(petstore.components.schemas.Pet);
+});
+
+test('should return a definition for a given ref that is escaped', () => {
+  expect(
+    findSchemaDefinition('#/components/schemas/Pet~1Error', {
+      components: {
+        schemas: {
+          'Pet/Error': petstore.components.schemas.Error,
+        },
+      },
+    })
+  ).toStrictEqual(petstore.components.schemas.Error);
+});
 
 test('should throw an error if there is an invalid ref', () => {
   expect(() => findSchemaDefinition('some-other-ref', {})).toThrow(/Could not find a definition for/);
@@ -6,4 +23,17 @@ test('should throw an error if there is an invalid ref', () => {
 
 test('should throw an error if there is a missing ref', () => {
   expect(() => findSchemaDefinition('#/components/schemas/user', {})).toThrow(/Could not find a definition for/);
+});
+
+test("should throw an error if an escaped ref isn't present its non-escaped format", () => {
+  expect(() => {
+    findSchemaDefinition('#/components/schemas/Pet~1Error', {
+      components: {
+        schemas: {
+          // This should be written in the schema as `Pet/Error`.
+          'Pet~1Error': {},
+        },
+      },
+    });
+  }).toThrow(/Could not find a definition for/);
 });
--- a/packages/tooling/__tests__/lib/flatten-schema.test.js
+++ b/packages/tooling/__tests__/lib/flatten-schema.test.js
@@ -1,6 +1,7 @@
 const flattenSchema = require('../../src/lib/flatten-schema');
 
 const petstore = require('../__fixtures__/petstore.json');
+const petstoreExpanded = require('@readme/oas-examples/3.0/json/petstore-expanded.json');
 
 test('should flatten schema to an array', () => {
   const schema = {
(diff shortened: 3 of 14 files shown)


packages/tooling/src/lib/flatten-schema.js
```diff
@@ -11,86 +11,107 @@ const getName = (parent, prop) => {
...
-}
+      return allof;
+    } else if ('oneOf' in obj) {
+      // We can't merge flatten objects in a `oneOf` representation into a single structure because that wouldn't
+      // validate against one of the objects, so let's just pick the first one present and flatten only that one.
+      return flattenSchema(obj.oneOf.shift());
+    } else if ('anyOf' in obj) {
+      // We can't merge flatten objects in an `anyOf` representation into a single structure because that wouldn't
+      // validate against one of the objects, so let's just pick the first one present and flatten only that one.
+      return flattenSchema(obj.anyOf.shift());
```
gratcliff: Can we add `TODO`s here so that we track support for the future?

erunion: Added a better comment with some links instead.

packages/tooling/__tests__/lib/find-schema-definition.test.js
```diff
@@ -1,9 +1,39 @@
+const petstore = require('@readme/oas-examples/3.0/json/petstore.json');
 const findSchemaDefinition = require('../../src/lib/find-schema-definition');
 
+test('should return a definition for a given ref', () => {
+  expect(findSchemaDefinition('#/components/schemas/Pet', petstore)).toStrictEqual(petstore.components.schemas.Pet);
+});
+
+test('should return a definition for a given ref that is escaped', () => {
+  expect(
+    findSchemaDefinition('#/components/schemas/Pet~1Error', {
```
domharrington: Is this pattern of escaping actually what happens in openapi? Is this supported? `Pet~1Error`

packages/tooling/__tests__/lib/flatten-schema.test.js
```diff
@@ -87,3 +88,166 @@ describe('$ref usages', () => {
     expect(flattenSchema(schema, oas)).toStrictEqual(expected);
   });
 });
+
+describe('polymorphism cases', () => {
+  describe('allOf', () => {
```
domharrington: So flattenSchema is for use in the response sidebar? 👌

packages/tooling/__tests__/lib/flatten-schema.test.js
```diff
@@ -87,3 +88,166 @@ describe('$ref usages', () => {
...
+        { name: 'tags[].id', type: 'Integer', description: undefined },
+        { name: 'tags[].name', type: 'String', description: undefined },
+        { name: 'status', type: 'String', description: 'pet status in the store' },
+        { name: 'id', type: 'Integer', description: undefined },
+      ]);
+    });
+  });
+
+  describe('anyOf', () => {
+    it('should flatten only the first schema listed', () => {
```
domharrington: This is good for now! We can work on this in future to allow selection in the UI.

packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -46,7 +46,7 @@ function getBodyParam(pathOperation, oas) {
     type,
     label: types[type],
     schema: oas.components
-      ? { definitions: { components: cleanupSchemaDefaults(oas.components) }, ...cleanupSchemaDefaults(schema.schema) }
+      ? { components: cleanupSchemaDefaults(oas.components), ...cleanupSchemaDefaults(schema.schema) }
```
domharrington: Is this how you're supposed to do it in OAS? Did I misread the spec?

.commitlintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["@commitlint/config-conventional"]
```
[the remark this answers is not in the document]
erunion: Helps conform commits to a common standard https://commitlint.js.org/. I've been manually adopting this in our OSS projects because it helps to keep changelogs consistent, this just formalizes that here.

packages/tooling/__tests__/lib/flatten-schema.test.js
erunion: Yep, that's the only place it's used.

packages/tooling/__tests__/lib/find-schema-definition.test.js
erunion: Yeah, to have a a schema, `Pet/Error`, with a slash in it you need to escape the `/` as `~1`, otherwise trying to access `#/components/schemas/Pet/Error` would try to load:

```
components: { schemas: { Pet: { Error: { ... } } } }
``` 

[jsonpointer](https://www.npmjs.com/package/jsonpointer) handles this all for us.

https://swagger.io/docs/specification/using-ref/#escape

packages/tooling/src/lib/parameters-to-json-schema.js
erunion: Nope! This change was needed because the version of your `#/definitions` handling in our fork got merged last week. https://github.com/rjsf-team/react-jsonschema-form/pull/1506

RJSF can now properly access `#/components/schemas/...` without trying to think that they should be located under `#/definitions/components/schemas/...`.

PR #141, opened 2020-03-27
feat: replacing the core parameters-to-json-schema lib

erunion:
## 🧰 What's being changed?
Since the release of `@readme/api-explorer` v6 we've been running into a truckload of JSON Schema issues with unsupported keywords like `example`. Instead of trying to put these fires out as they crop up, I'm swapping our our code OpenAPI parameter/request body -> JSON Schema conversion library with two packages:

* [@apidevtools/json-schema-ref-parser](https://www.npmjs.com/package/@apidevtools/json-schema-ref-parser)
* [@openapi-contrib/openapi-schema-to-json-schema](https://www.npmjs.com/package/@openapi-contrib/openapi-schema-to-json-schema)

With these two packages we can:

1. Dereference schemas before we process, eliminating the need to recur and resolve `$ref`'s as we go with our `find-schema-definition` library.
2. Remove a whole bunch of custom OpenAPI to JSON Schema conversion code that's handled much more gracefully, and with better test and spec coverage.

### DIFF
--- a/packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
+++ b/packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
@@ -4,10 +4,6 @@ const fixtures = require('../__fixtures__/lib/json-schema');
 
 const polymorphismScenarios = ['oneOf', 'allOf', 'anyOf'];
 
-const schemaDraftVersion = {
-  $schema: 'http://json-schema.org/draft-04/schema#',
-};
-
 test('it should return with null if there are no parameters', async () => {
   expect(await parametersToJsonSchema({ parameters: [] })).toBeNull();
   expect(await parametersToJsonSchema({})).toBeNull();
@@ -98,10 +94,7 @@ describe('parameters', () => {
             oas
           )
         )[0].schema.properties.param
-      ).toStrictEqual({
-        ...oas.components.parameters.Param.schema,
-        ...schemaDraftVersion,
-      });
+      ).toStrictEqual(oas.components.parameters.Param.schema);
     });
 
     it('should fetch parameters that have a child $ref', async () => {
@@ -166,7 +159,6 @@ describe('parameters', () => {
       ).toStrictEqual({
         param: {
           type: 'string',
-          ...schemaDraftVersion,
         },
       });
     });
@@ -306,7 +298,6 @@ describe('request bodies', () => {
         label: 'Body Params',
         type: 'body',
         schema: {
-          ...schemaDraftVersion,
           type: 'object',
           properties: {
             a: { type: 'string' },
@@ -339,7 +330,6 @@ describe('request bodies', () => {
         label: 'Form Data',
         type: 'formData',
         schema: {
-          ...schemaDraftVersion,
           type: 'object',
           properties: {
             a: { type: 'string' },
@@ -404,12 +394,8 @@ describe('request bodies', () => {
           type: 'body',
           label: 'Body Params',
           schema: {
-            ...schemaDraftVersion,
             ...oas.components.schemas.Pet,
-            components: {
-              ...schemaDraftVersion,
-              ...oas.components,
-            },
+            components: oas.components,
           },
         },
       ]);
@@ -433,7 +419,6 @@ describe('type', () => {
       expect((await parametersToJsonSchema({ parameters }))[0].schema).toStrictEqual({
         properties: {
           param: {
-            ...schemaDraftVersion,
             items: {},
             type: 'array',
           },
@@ -509,7 +494,6 @@ describe('enums', () => {
           type: 'object',
           properties: {
             Accept: {
-              ...schemaDraftVersion,
               type: 'string',
               enum: ['application/json', 'application/xml'],
             },
@@ -569,7 +553,6 @@ describe('descriptions', () => {
           type: 'object',
           properties: {
             Accept: {
-              ...schemaDraftVersion,
               description: 'Expected response format.',
               type: 'string',
             },
--- a/packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
+++ b/packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
@@ -4,19 +4,23 @@ const fixtures = require('../__fixtures__/lib/json-schema');
 
 const polymorphismScenarios = ['oneOf', 'allOf', 'anyOf'];
 
+const schemaDraftVersion = {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+};
+
 test('it should return with null if there are no parameters', async () => {
-  expect(parametersToJsonSchema({ parameters: [] })).toBeNull();
-  expect(parametersToJsonSchema({})).toBeNull();
+  expect(await parametersToJsonSchema({ parameters: [] })).toBeNull();
+  expect(await parametersToJsonSchema({})).toBeNull();
 });
 
 describe('parameters', () => {
   describe('type sorting', () => {
     const schema = {
       parameters: [
-        { in: 'path', name: 'path parameter' },
-        { in: 'query', name: 'query parameter' },
-        { in: 'header', name: 'header parameter' },
-        { in: 'cookie', name: 'cookie parameter' },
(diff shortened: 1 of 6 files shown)


packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -20,83 +27,45 @@ function getBodyParam(pathOperation, oas) {
...
-            break;
-
-          case 'type':
-            if (obj.type === 'array') {
-              if (!('items' in obj)) {
-                if ('properties' in obj) {
-                  // This is a fix to handle cases where someone may have typod `items` as `properties` on an array.
-                  // Since throwing a complete failure isn't ideal, we can see that they meant for the type to be
-                  // `object`, so we  can do our best to shape the data into what they were intendint it to be.
-                  // README-6R
```
erunion: Since `@openapi-contrib/openapi-schema-to-json-schema` ignores `properties` if the schema type is an array, I'm opting to no longer support this mangled OAS usecase. Instead, if we have a document with `type: array`, and it has `properties`, `@openapi-contrib/openapi-schema-to-json-schema` will return it as `{ type: array }` and then we'll add `item: {}` to fill it out as a mixed-content input.

packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -123,107 +92,31 @@ function getOtherParams(pathOperation, oas) {
...
-        // If we have `allowEmptyValue` present, and the default is actually an empty string, let it through as it's
-        // allowed.
-        schema.default = data.default;
-      } else if (data.default !== '') {
-        schema.default = data.default;
-      }
-    }
-
-    if ('maxLength' in data) schema.maximum = data.maxLength;
-    if ('minLength' in data) schema.minimum = data.minLength;
```
erunion: This was actually a large misunderstanding on my part with `maxLength` and `minLength`. OAS supports those two properties, as well as `minimum` and `maximum`. `maxLength` and `minLength` are to be used on strings, whereas `maximum` and `minimum` are for numbers. Mixing them like I was previously doing here wrong.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -5,26 +5,26 @@ const fixtures = require('../__fixtures__/lib/json-schema');
...
     const schema = {
       parameters: [
-        { in: 'path', name: 'path parameter' },
-        { in: 'query', name: 'query parameter' },
-        { in: 'header', name: 'header parameter' },
-        { in: 'cookie', name: 'cookie parameter' },
+        { in: 'path', name: 'path parameter', schema: { type: 'string' } },
+        { in: 'query', name: 'query parameter', schema: { type: 'string' } },
+        { in: 'header', name: 'header parameter', schema: { type: 'string' } },
+        { in: 'cookie', name: 'cookie parameter', schema: { type: 'string' } },
```
erunion: Added `schema` properties to these because it was invalid OAS and wouldn't pass `@openapi-contrib/openapi-schema-to-json-schema`.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -82,25 +82,26 @@ describe('parameters', () => {
...
       ).toStrictEqual(oas.components.parameters.Param.schema);
     });
 
-    it('should fetch parameters that have a child $ref', () => {
+    it('should fetch parameters that have a child $ref', async () => {
       const oas = {
         components: {
           schemas: {
             string_enum: {
-              name: 'string',
```
erunion: Removed `name` from this example because it's irrelvant..

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -335,43 +358,7 @@ describe('request bodies', () => {
   });
 
   describe('$ref support', () => {
-    it('should work for top-level request body $ref', () => {
```
erunion: I removed this test case because pulling a `$ref` for a `requestBody` that has no `content` or a mime type isn't actually validated OAS.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -429,37 +416,21 @@ describe('type', () => {
...
       ];
 
-      expect(parametersToJsonSchema({ parameters })[0].schema).toStrictEqual({
-        properties: { param: { items: {}, type: 'array' } },
-        required: [],
-        type: 'object',
-      });
-    });
-
-    it('should repair a malformed object that is typod as an array [README-6R]', () => {
```
erunion: `@openapi-contrib/openapi-schema-to-json-schema` ignores `properties` here because it's an array so this is no longer a case we need to try to repair.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -655,44 +585,36 @@ describe('additionalProperties', () => {
...
-      expect(parametersToJsonSchema({ parameters })[0].schema.properties.param.items).toStrictEqual({
+      expect((await parametersToJsonSchema({ parameters }))[0].schema.properties.param.items).toStrictEqual({
         additionalProperties: {
           type: 'string',
         },
         type: 'object',
       });
     });
-
-    it('should be ignored when set to `false`', () => {
```
erunion: This isn't a case we actually need to support or worry about.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -848,122 +764,30 @@ describe('defaults', () => {
...
-        it.each(schemaScenarios)(`scenario: %s`, scenario => {
+        it.each(schemaScenarios)(`scenario: %s`, async scenario => {
           const { requestBody, oas } = fixtures.generateRequestBodyDefaults(refType, scenario, fixtureOptions);
-          expect(parametersToJsonSchema({ requestBody }, oas)).toMatchSnapshot();
+          expect(await parametersToJsonSchema({ requestBody }, oas)).toMatchSnapshot();
         });
       });
     });
-
-    describe('should comply with the `allowEmptyValue` declarative when present', () => {
```
erunion: Since `allowEmptyValue` is actually only relevant to query params, and only when making an API request, it's not really relevant to support this kind of adhoc determination in this library, and instead this handling should happen further up the stack in `@readme/api-explorer` when we make an API request.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -848,122 +764,30 @@ describe('defaults', () => {
...
-
-      describe.each(polymorphismScenarios)('with usages of `%s`', mod => {
-        it.each(schemaScenarios)(`scenario: %s`, scenario => {
-          const { requestBody, oas } = fixtures.generateRequestBodyDefaults(mod, scenario, fixtureOptions);
-          expect(parametersToJsonSchema({ requestBody }, oas)).toMatchSnapshot();
-        });
-      });
-    });
-
-    describe('should not add a default when one is missing', () => {
```
erunion: This isn't a case we really need to support or worry about.

packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -848,122 +764,30 @@ describe('defaults', () => {
...
-        it.each(schemaScenarios)(`scenario: %s`, scenario => {
-          const { requestBody, oas } = fixtures.generateRequestBodyDefaults(mod, scenario, fixtureOptions);
-          expect(parametersToJsonSchema({ requestBody }, oas)).toMatchSnapshot();
-        });
-      });
-    });
-  });
-});
-
-describe('minLength / maxLength', () => {
```
erunion: See https://github.com/readmeio/oas/pull/141/files?diff=split&w=1#r399093091 for details on why this test case was removed.

erunion:
I ended up rewriting this code instead in https://github.com/readmeio/oas/pull/358.

PR #165, opened 2020-04-20
feat: look in request/response content when retrieving headers

kanadgupta:
Partially inspired by [`api-explorer`](https://github.com/readmeio/api-explorer/blob/d1df9260c6a7d9d106aa277716fb95d56e5a50fb/packages/oas-to-har/src/index.js#L244-L253).

In the `getHeaders()` function, I'm adding logic to also look in the [`requestBody`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#request-body-object) and [`response`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#responseObject) objects to see if there are `content` fields that contain [media type objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#mediaTypeObject). If these exist, we can infer `Content-Type` and `Accept` headers and add them to the result.

### DIFF
--- a/packages/tooling/src/oas.js
+++ b/packages/tooling/src/oas.js
@@ -114,6 +114,7 @@ class Operation {
       } else if (this.requestBody.content && Object.keys(this.requestBody.content))
         this.headers.request.push('Content-Type');
     }
+
     // This is a similar approach, but in this case if we check the response content
     // and prioritize the 'accept' request header and 'content-type' request header
     if (this.responses) {
--- a/packages/tooling/src/oas.js
+++ b/packages/tooling/src/oas.js
@@ -117,9 +117,11 @@ class Operation {
       }
       // This is a similar approach, but in this case if we check the response content
       // and prioritize the 'accept' request header and 'content-type' request header
-      if (!this.headers.response.includes('Content-Type') && this.responses) {
-        if (Object.keys(this.responses).some(response => !!this.responses[response].content))
-          this.headers.response.push('Content-Type');
+      if (this.responses) {
+        if (Object.keys(this.responses).some(response => !!this.responses[response].content)) {
+          if (!this.headers.request.includes('Accept')) this.headers.request.push('Accept');
+          if (!this.headers.response.includes('Content-Type')) this.headers.response.push('Content-Type');
+        }
       }
     }
 
--- a/packages/tooling/src/oas.js
+++ b/packages/tooling/src/oas.js
@@ -67,7 +67,7 @@ class Operation {
       }, {});
   }
 
-  getHeaders() {
+  getHeaders(includeContent = false) {
     this.headers = {
       request: [],
       response: [],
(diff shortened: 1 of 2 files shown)


packages/tooling/src/oas.js
```diff
@@ -104,6 +104,25 @@ class Operation {
...
+          const ref = findSchemaDefinition(this.requestBody.$ref, this.oas);
+          if (ref.content && Object.keys(ref.content)) this.headers.request.push('Content-Type');
+        } else if (this.requestBody.content && Object.keys(this.requestBody.content))
+          this.headers.request.push('Content-Type');
+      }
+      // This is a similar approach, but in this case if we check the response content
+      // and prioritize the 'accept' request header and 'content-type' request header
+      if (!this.headers.response.includes('Content-Type') && this.responses) {
+        if (Object.keys(this.responses).some(response => !!this.responses[response].content))
+          this.headers.response.push('Content-Type');
```
kanadgupta, line 122: I'm debating if we should also add an `Accept` request header to this block. We do something similar to that in the explorer: https://github.com/readmeio/api-explorer/blob/d1df9260c6a7d9d106aa277716fb95d56e5a50fb/packages/oas-to-har/src/index.js#L139-L142

erunion: +1 to adding an Accept header in there if already isn't one.

We should probably replace a lot of the work in oas-to-har with this.

kanadgupta: Added! https://github.com/readmeio/oas/pull/165/commits/046bd82d5500484b44f919d1355e12ee18ad8ec5

packages/tooling/src/oas.js
```diff
@@ -67,7 +67,7 @@ class Operation {
       }, {});
   }
 
-  getHeaders() {
+  getHeaders(includeContent = false) {
```
kanadgupta, line 70: @erunion only other thing I'm debating is how we should wrap this logic (if at all) — currently I just wrapped it in a flag. Let me know if you have any better ideas.

erunion: I've been trying to think of a scenario where we wouldn't want to return the accept/content-type, and I'm not sure. Maybe if you just wanted the explicitly-defined headers? But then again, request and response content-types are defined explicitly.

kanadgupta: Yeah I wasn't sure how we draw the line here... since this feels more like an inference than an explicitly-defined header

erunion: In metrics, if we're always going to do `includeContent=true` then we might as well just always return it.

kanadgupta: Cool, I'll revert the flag stuff and just make it so it always returns these.

packages/tooling/src/oas.js
```diff
@@ -104,6 +104,25 @@ class Operation {
...
+    // we check if the path operation request body contains content, which implies that
+    // we should also include the 'content-type' header.
+    if (!this.headers.request.includes('Content-Type') && this.requestBody) {
+      if (this.requestBody.$ref) {
+        const ref = findSchemaDefinition(this.requestBody.$ref, this.oas);
+        if (ref.content && Object.keys(ref.content)) this.headers.request.push('Content-Type');
+      } else if (this.requestBody.content && Object.keys(this.requestBody.content))
+        this.headers.request.push('Content-Type');
+    }
+    // This is a similar approach, but in this case if we check the response content
```
erunion, line 117:
```suggestion

    // This is a similar approach, but in this case if we check the response content
```

PR #167, opened 2020-04-21
fix: being more leanient with how we're constructing types

erunion:
If a schema doesn't have a defined `type`, instead of accepting that, we should attempt to do our best to give it a type so it's still valid JSON Schema.

### DIFF
--- a/packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
+++ b/packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
@@ -544,7 +544,7 @@ describe('type', () => {
       });
     });
 
-    it('should repair a malformed object that is typod as an array [README-6R]', () => {
+    it('should discard a malformed object that is typod as an array and treat it as a string [README-6R]', () => {
       const parameters = [
         {
           name: 'param',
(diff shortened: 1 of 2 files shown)


packages/tooling/__tests__/lib/parameters-to-json-schema.test.js
```diff
@@ -559,11 +559,30 @@ describe('type', () => {
       ];
 
       expect(parametersToJsonSchema({ parameters })[0].schema).toStrictEqual({
-        properties: { param: { type: 'array' } },
+        properties: { param: { type: 'object' } },
```
erunion, line 562: This test result should have always been `object`, but we were setting it to `array` due to a bug on line 242 where we'd overwrite the type we we were setting and re-using the originally supplied type (`array` in this case). 😬

domharrington, line 562: Is this a major breaking change? Or was this test case always broken?

erunion: Just added a comment on this. It was a broken test case.

PR #173, opened 2020-04-29
fix: bug where a string type would be added to anyOf, oneOf, and allOf

erunion:
This resolves a bug in our processing of `anyOf`, `oneOf` and `allOf` schemas where in specific cases, we'd add a `type: string` default to it if it were missing a type. This would result in the following broken UI in the API Explorer:

80650697-54d8ad80-8a29-11ea-9d27-e49836e2a98a

Turns out that trying to repair bad OAS schemas is difficult.

### DIFF
--- a/packages/tooling/src/lib/parameters-to-json-schema.js
+++ b/packages/tooling/src/lib/parameters-to-json-schema.js
@@ -195,9 +195,9 @@ function getOtherParams(pathOperation, oas) {
       !('type' in data) &&
       !('$ref' in data) &&
       !('allOf' in data) &&
-      !('oneOf' in data) &&
       !('anyOf' in data) &&
-      prevProp !== 'additionalProperties'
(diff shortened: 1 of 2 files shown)


packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -303,10 +303,15 @@ function getOtherParams(pathOperation, oas) {
 
     const properties = parameters.reduce((prev, current) => {
       const schema = {
-        type: 'string',
         ...(current.schema ? constructSchema(current.schema) : {}),
       };
 
+      // If we still don't have a `type` at the highest level of this schema, and it's not a polymorphism/inheritance
+      // model, add a `string` type so the schema will be valid.
```
rafegoldberg, line 310: ### 👍

PR #174, opened 2020-04-30
fix: edge case where component would sometimes be marked a string

erunion:
| [🎟 Asana](https://app.asana.com/0/681252538274980/1173503272587186/f) |
| :---  |

## 🆘 Problem

This resolves an edge case that was introduced in some logic I added to repair schemas without a `type` that was sometimes marking the schema object as a `string`.  This would result in the following "broken" UI:
 
Screen Shot 2020-04-29 at 10 42 36 PM

You can see it in action here:  http://bin.readme.com/s/5eaa651315dd18002423bd32

Now if we remove the `InputErrorResponse`, we'll see that it's resolved:

Screen Shot 2020-04-29 at 10 43 08 PM

Now what if you move the `InputErrorResponse` schema below `NewPreauthorization`?

Screen Shot 2020-04-29 at 10 46 16 PM

😬 

## 🚧 Resolution
So the reason this was happening was that since our `cleanupSchemaDefaults` method for `requestBody` and component schema objects is recursive, and we were invoking it against the `components` object, and not individual schemas, it was difficult for us to ascertain when we were **first** processing a component schema and when we should add a missing `type` at the top level of a schema if it didn't have one.

For this specific case, what was tripping up our logic was a few things:

* Both `InputErrorResponse` and `NewPreauthorization` schemas didn't have a defined `type`.
* Our code would correctly identify `InputErrorResponse` as an object because we hadn't seen a top-level `properties` property yet until that point, and when we did we immediately knew that we were processing an object, and could mark it as such.
* Once we hit the `NewPreauthorization` schema, and after we had already processed `InputErrorResponse`, we had the following data in our internal recursive `prevProps` memory state:

```
[
  "schemas", 
  "InputErrorResponse", 
  "properties", 
  "code", 
  "message", 
  "NewPreauthorization"
]
```

Since we retained knowledge of parsing `properties`, our code misinterpreted this as having already seen this schema before, is it not parsing properties again for it and since it's missing a `type` to add `string` so it's still valid JSON Schema. 

source

Resolving this involved a couple things:

* [x] Shift our recursive processing of component schemas to start at the schema itself, not the `components` level. This way we'll know exactly when we're **first** seeing a component schema.
* [x] Update our internal `prevProps` memory state to clean itself out if we don't have a `prevProp`. This will resolve the issue of `prevProps` retaining data from previously processed schemas.

### DIFF
--- a/packages/tooling/src/lib/parameters-to-json-schema.js
+++ b/packages/tooling/src/lib/parameters-to-json-schema.js
@@ -21,6 +21,11 @@ function getBodyParam(pathOperation, oas) {
 
   const cleanupSchemaDefaults = (obj, prevProp = false, prevProps = []) => {
     Object.keys(obj).forEach(prop => {
+      // Since this method is recursive, let's reset our states when we're first processing a new property tree.
+      if (!prevProp) {
+        prevProps = [];
+      }
+
       if (obj[prop] === null) {
         // If the item is null, just carry on. Why do this in addition to `typeof obj[prop] == object`? Because
         // `typeof null` equates to `object` for "legacy reasons" apparently.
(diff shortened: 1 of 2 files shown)


packages/tooling/src/lib/parameters-to-json-schema.js
```diff
@@ -77,15 +82,10 @@ function getBodyParam(pathOperation, oas) {
...
-              prevProp !== false &&
-              prevProps.includes('schemas') &&
-              'components' in oas &&
-              'schemas' in oas.components &&
-              prevProp in oas.components.schemas
-            ) {
-              // If we have a previous prop, but we're parsing the immediate schemas tree in the components object,
-              // then we're processing a title and description on a component schema.
+            } else if (prevProp !== false && prevProps.length === 1) {
+              // If we have a previous prop, but we're parsing the immediate schemas tree (we know this if prevProps
```
rafegoldberg, line 86: this stuff is crazy

erunion: All this just to remove this kind of stuff because RJSF renders it as an orphan block of text:

Screen Shot 2020-04-30 at 7 54 54 AM

Screen Shot 2020-04-30 at 7 56 11 AM

PR #184, opened 2020-05-11
feat: adding a getContentType method into the OAS Operation class

erunion:
This method exists currently in `@readme/oas-to-har` but it makes more sense for it to live here.

https://github.com/readmeio/api-explorer/blob/master/packages/oas-to-har/src/index.js#L32-L53

### DIFF
--- a/packages/tooling/__tests__/operation.test.js
+++ b/packages/tooling/__tests__/operation.test.js
@@ -4,10 +4,16 @@ const petstore = require('./__fixtures__/petstore.json');
 const multipleSecurities = require('./__fixtures__/multiple-securities.json');
 const referenceSpec = require('./__fixtures__/local-link.json');
 
+describe('#getContentType', () => {
+  it('should return the content type on an operation', () => {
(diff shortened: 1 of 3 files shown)


packages/tooling/__tests__/operation.test.js
```diff
@@ -4,10 +4,16 @@ const petstore = require('./__fixtures__/petstore.json');
 const multipleSecurities = require('./__fixtures__/multiple-securities.json');
 const referenceSpec = require('./__fixtures__/local-link.json');
 
+describe('#getContentType', () => {
+  it('should return the content type on an operation', () => {
```
domharrington, line 8: Are there any other tests in oas-to-har that we can pull over? Few little weird edge cases in that function! 

https://github.com/readmeio/api-explorer/blob/06aa341b629680b255e1469e427c76fe08e05bc3/packages/oas-to-har/__tests__/index.test.js#L1325-L1470

erunion: There were a couple I had missed.

PR #190, opened 2020-05-28
Bundle $refs, and other minor changes

gkoberger:


packages/cli/src/api.js
```diff
@@ -62,57 +45,6 @@ exports.api = function (args, opts) {
       let apiId = swagger.info.title ? uslug(swagger.info.title) : crypto.randomBytes(7).toString('hex');
 
       request.get(`${config.host.url}/check/${apiId}`, { json: true }, (err, check) => {
```
erunion, line 47: Just going to remove this API call since we aren't using `check` anymore.

PR #207, opened 2020-06-19
fix: issue where findOperation would fail on server with trailing slash

erunion:
This resolves a bug in `Oas.findOperation` and `Oas.getOperation` where if the defined server URL had a trailing slash, we wouldn't be able to locate an operation within the definition.

### DIFF
--- a/packages/tooling/__tests__/oas.test.js
+++ b/packages/tooling/__tests__/oas.test.js
@@ -151,6 +151,40 @@ describe('#findOperation()', () => {
     });
   });
 
+  it('should return result if server has a trailing slash', () => {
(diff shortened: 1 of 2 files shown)


erunion:
The part of code where this was happening was within `findOperation` where we were attempting to compare the origin `https://example.com`, against the server URL, `https://example.com/`.

https://github.com/readmeio/oas/blob/master/packages/tooling/src/oas.js#L151-L158

`targetServer` on this line would be falsy because it couldn't find `https://example.com` in the servers list, and the entire thing would bail out.

domharrington:
Makes perfect sense thanks for clarifying!

PR #223, opened 2020-07-16
fix: findOperation issues with "/" paths

kanadgupta:
For [Paths Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#pathsObject) that only contain `/`, `findOperation()` wasn't handling trailing slashes properly. For example, the ReadMe API has a `/` endpoint but `findOperation()` will return `undefined` for routes like https://dash.readme.com/api/v1 since it'll look for https://dash.readme.com/api/v1/.

Not sure if this is the best approach but I think it should be fine since [the spec dictates](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#user-content-patterned-fields:~:text=The%20field%20name%20MUST%20begin%20with%20a%20forward%20slash%20(%2F).) the following:
> The field name MUST begin with a forward slash (`/`)

meaning that all paths should contain a `/` at the bare minimum.

### DIFF
--- a/packages/tooling/__tests__/oas.test.js
+++ b/packages/tooling/__tests__/oas.test.js
@@ -217,6 +217,14 @@ describe('#findOperation()', () => {
       slugs: {},
       method: 'GET',
     });
+
+    expect(res.operation).toStrictEqual({
+      responses: {
+        '200': {
+          description: 'OK',
+        },
+      },
+    });
   });
 
   it('should return result if in server variable defaults', () => {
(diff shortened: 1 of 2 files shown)


packages/tooling/__tests__/oas.test.js
```diff
@@ -185,6 +185,40 @@ describe('#findOperation()', () => {
...
+          },
+        },
+      },
+    });
+
+    const uri = 'https://example.com';
+    const method = 'get';
+
+    const res = oas.findOperation(uri, method);
+    expect(res.url).toStrictEqual({
```
erunion, line 213: Just for sanity sake, can you also test that `res.operation` is filled with the right object?

PR #292, opened 2020-10-20
fix: swapping out the $ref parsing engine of flattenSchema

erunion:
## 🧰 What's being changed?

This completely swaps out the `$ref` parsing engine behind `flattenSchema` with [json-schema-ref-parser](https://github.com/APIDevTools/json-schema-ref-parser)  to allow us to properly, easily, handle circular references without infinitely recusing them.

How it works is we'll funnel an incoming schema and API definition into `json-schema-ref-parser` to dereference it but allowing it to ignore circular refs. If a `$ref` is circular, it'll remain in the schema as `$ref`, enabling us to quickly detect and ignore circular refs when flattening a schema.

With this change, `flattenSchema` is now an async function.

### DIFF
--- a/__tests__/tooling/lib/flatten-schema.test.js
+++ b/__tests__/tooling/lib/flatten-schema.test.js
@@ -187,6 +187,57 @@ describe('$ref usages', () => {
     ]);
   });
 
+  describe('external $refs', () => {
+    it('should ignore an external $ref inside of an array', async () => {
+      const externalSchema = {
+        properties: {
+          responses: {
+            type: 'array',
+            items: {
+              $ref: 'https://example.com/ApiResponse.yaml',
+            },
+          },
+          tag: {
+            $ref: '#/components/schemas/Tag',
+          },
+        },
+      };
+
+      expect(await flattenSchema(externalSchema, petstore)).toStrictEqual([
+        { name: 'responses', type: '[Circular]', description: undefined },
+        { name: 'tag', type: 'Object', description: undefined },
+        { name: 'tag.id', type: 'Integer', description: undefined },
+        { name: 'tag.name', type: 'String', description: undefined },
+      ]);
+    });
+
+    it('should ignore an external $ref on an object', async () => {
+      const externalSchema = {
+        properties: {
+          responses: {
+            type: 'array',
+            items: {
+              // $ref: 'https://example.com/ApiResponse.yaml',
+              $ref: '#/components/schemas/ApiResponse',
+            },
+          },
+          tag: {
+            $ref: 'https://example.com/Tag.yaml',
+            // $ref: '#/components/schemas/Tag',
+          },
+        },
+      };
+
+      expect(await flattenSchema(externalSchema, petstore)).toStrictEqual([
+        { name: 'responses', type: '[Object]', description: undefined },
+        { name: 'responses[].code', type: 'Integer', description: undefined },
+        { name: 'responses[].type', type: 'String', description: undefined },
+        { name: 'responses[].message', type: 'String', description: undefined },
+        { name: 'tag', type: 'Circular', description: undefined },
+      ]);
+    });
+  });
+
   describe('circular $ref', () => {
     const oas = {
       components: {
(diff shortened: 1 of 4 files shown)


dok:
How does this work in conjunction with https://github.com/readmeio/react-jsonschema-form? Are they related? 

I'm a little bit out of the loop on this one

__tests__/tooling/lib/flatten-schema.test.js
```diff
@@ -186,41 +186,89 @@ describe('$ref usages', () => {
...
+      };
+
+      expect(await flattenSchema(circularSchema, oas)).toStrictEqual([
+        { name: 'id', type: 'Number', description: undefined },
+        { name: 'fields', type: 'Circular', description: undefined },
+      ]);
+    });
+
+    it('should not recurse a circular $ref inside of an array', async () => {
+      const circularSchema = {
```
dok, line 227: Would we need to test $anyOf, $allOf and such as well?

erunion:
@dok This is completely unrelated to that. This `flattenSchema` method is only utilized when we flatten a response schema. 

Ideally we'd dereference these schemas before they're passed into the explorer from the beginning so we wouldn't need to have any `$ref` fetching or parsing at all, but we're ways away from there as it'll require some reconfiguration in our backend models and potentially how we're storing data.

__tests__/tooling/lib/flatten-schema.test.js
erunion: No we shouldn't need to since if a `$ref` is anywhere, and it's circular according to `json-schema-ref-parser` it'll remain as a `$ref: <pointer>` in the schema.

tooling/lib/flatten-schema.js
```diff
@@ -11,14 +11,28 @@ const getName = (parent, prop) => {
...
 
-module.exports = (schema, oas) => {
+module.exports = async (schema, oas) => {
+  const derefSchema = await $RefParser.dereference(
+    { ...schema, ...oas },
+    {
+      dereference: {
+        // If circular `$refs` are ignored they'll remain in `derefSchema` as `$ref: String`, otherwise `$ref‘ just
+        // won't exist. This allows us to do easy circular reference detection.
+        circular: 'ignore',
```
dok, line 21: What would be the way we would detect whether a dereferenced object after this function call? Whether a `$ref: String` exists within the dereferenced object?

dok: NVM I see it now: https://github.com/readmeio/oas/pull/292/files#diff-204487ea7066d929383376809d6e70ea36310eca01a89b9ef90cc8e7d366f150R35

erunion: If `$ref` exists in `derefSchema` at all, then it's circular.

dok: Gotcha. Are there any other times where a $ref cannot be satisfied?

dok:
@erunion oh I see. Thanks for the clarification. Is this easier to work with than the isCyclic functionality? It does seem to have a similar format to that (it looks recursive).

tooling/lib/flatten-schema.js
erunion: If it's an external ref (like a URL to another spec) it'll remain unsatisfied if it's inaccessible.

That said, we shouldn't be attempting to resolve external refs at this point in the application cycle so I'm going to add `resolve.external: false` into the options here to prevent that as well as a test for that.

erunion:
@dok Yeah this work could likely be built into either our form module, or when we build up JSON Schema with `tooling/lib/parametersToJsonSchema` to do this resolving work then.

Reworking that is all is a much larger project though and should probably be done further up the stack so we aren't processing the same `$ref` *n* times.

dok (approved):
Pre-approving with [this](https://github.com/readmeio/oas/pull/292/files#r508896765) in mind

tooling/lib/flatten-schema.js
```diff
@@ -44,10 +48,17 @@ module.exports = async (schema, oas) => {
         if (value.type === 'object') {
           array.push(flattenSchema(value, getName(parent, prop), level + 1));
         } else if (value.type === 'array' && value.items) {
```
dok, line 50: Will evaluating value.type === 'array' always result in value.items being an array? Another way to interpret whether value.items is an array would be `Array.isArray(value.items)`

erunion: `value.items` is always going to be an object as any other type attached to `items` is not valid against the OAS.

dok: Ok!

dok:
@erunion looks good!

PR #295, opened 2020-10-29
feat: ability to generate sample schemas for a request and response

erunion:
## 🧰 What's being changed?

* [x] Adds an async `getResponseExamples` method to the `Operation` class that can pull an array of examples that are available on it. This will replace the `get-response-examples` library currently in use within `@readme/api-explorer`.
  * Schema example generation code behind this has been pulled over from from `swagger-ui`, with full attribution, and in some cases been slightly modified. 
  * The code and tests behind this are largely the same as in `@readme/api-explorer`, but slightly refactored to reduce code bloat and supporting auto-generation when curated response examples (via the `example` or `examples` property at the media type, **not** schema, level) aren't present.
* [x] Also adds an async `getRequestBodyExamples` method that does the same as `getResponseExamples`, just instead returns request body examples.

### Automatic schema generation
For request bodies or responses that have no defined examples, we'll now start generating an example schema based off its schema. The code to accomplish this has been pulled from `swagger-ui`, and an example of it in action would be the following:

```json
"/emptyexample-with-schema": {
  "post": {
    "description": "This operation handles has no schemas, but empty `examples`.",
    "requestBody": {
      "required": true,
      "content": {
        "application/json": {
          "schema": {
            "$ref": "#/components/schemas/tag"
          },
          "examples": {}
        }
      }
    },
  }
}

...

"components": {
  "schemas": {
    "tag": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer",
          "format": "int64"
        },
        "name": {
          "type": "string"
        }
      }
    }
  }
}
```

As you can see this operation is referencing a `tag` schema, but neither the schema or the `application/json` media type have any examples. Running this code through `getRequestBodyExample` will return a generated schema like such:

```js
{
  id: 0,
  name: 'string',
}
```

Strings will be converted to a representation of its attached `format` (`email` will be `user@example.com`, `date` is a current date, `uuid` will be a uuid, etc.), numbers will be `0`, and booleans will be a boolean.

### getResponseExamples
Tons of unit tests were added, but the way that this'll work is you can run `operation.getResponseExamples()` and get back an array of the following sorts:

```js
[
  {
    languages: [
      {
        code: JSON.stringify([
          {
            id: 0,
            name: 'string',
          },
        ], undefined, 2),
        language: 'application/json',
        multipleExamples: false,
      },
    ],
    status: '200',
  },
]
```

This array can then be used in our API Explorer within our [ResponseExample](https://github.com/readmeio/api-explorer/blob/next/packages/api-explorer/src/ResponseExample.jsx) component that renders response examples.

### getRequestBodyExamples
Like with `getResponseExamples` this has a ton of tests, but if you run it on an operation that has a `requestBody` with a schema or example, you'll get something like this:

```js
[
  {
    code: 'string',
    mediaType: 'text/plain',
    multipleExamples: false,
  },
  {
    code: JSON.stringify({
      type: 'object',
      properties: {
        id: {
          type: 'number',
        },
        email: {
          type: 'string',
        },
        name: {
          type: 'string',
        },
      },
    }, undefined, 2),
    mediaType: 'application/json',
    multipleExamples: false,
  },
]
```


tooling/lib/get-mediatype-examples.js
```diff
@@ -0,0 +1,83 @@
...
+   */
+  getMediaTypeExample: mediaType => {
+    if (mediaType.example) {
+      return mediaType.example;
+    } else if (mediaType.examples) {
+      const examples = Object.keys(mediaType.examples);
+      if (examples.length) {
+        if (examples.length > 1) {
+          // Since we're trying to return a single example with this method, but have multiple present,
+          // return `false` so `getMultipleExamples` will pick up this response instead later.
```
darrenyong, line 21: `getMultipleExamples` is [here](https://github.com/readmeio/api-explorer/blob/ad91b045c493c92b1ae61102e4a25aa150e51fda/packages/api-explorer/src/lib/get-response-examples.js#L118) and [here](https://github.com/readmeio/api-explorer/blob/ad91b045c493c92b1ae61102e4a25aa150e51fda/packages/api-explorer/src/lib/get-response-examples.js#L55) right?

erunion: @darrenyong yep

erunion:
Approved in https://github.com/readmeio/api-explorer/pull/1027

PR #300, opened 2020-11-04
feat: add scope support to `oas generate` command

kschiu:
## What's being changed?

This PR passes in the `scope` argument to `swagger-inline` so that generated swagger files can be properly scoped.

As per https://github.com/readmeio/swagger-inline#available-options, swagger files can be generated by matching the scope field declared.

## Usage of scope
No scope
`oas generate VetStore.yaml`

With scope
`oas generate VetStore.yaml --scope public`

## Testing
Please see repo for example:
https://github.com/kschiu/oas_test

In the `src/` directory,  ran: `oas generate VetStore.yaml > ../no_scope.json`, which generated:
https://github.com/kschiu/oas_test/blob/master/no_scope.json

In the `src/` directory,  ran: `oas generate VetStore.yaml --scope public > ../public_scope.json`, which generated:
https://github.com/kschiu/oas_test/blob/master/public_scope.json

### DIFF
--- a/cli/lib/utils.js
+++ b/cli/lib/utils.js
@@ -25,17 +25,28 @@ exports.config = function (env) {
 
 exports.findSwagger = function (info, cb) {
   const base = exports.isSwagger(_.last(info.args)) ? _.last(info.args) : undefined;
-  const scope = info.opts.scope ? info.opts.scope : undefined;
 
   swaggerInline('**/*', {
     format: '.json',
     metadata: true,
+    scope: info.opts.scope,
     base,
-    scope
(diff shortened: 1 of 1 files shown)


kschiu:
Hey @erunion, I seem to be unable to add a reviewer - could you take a look and let me know what you think? The use case here is being able to pass in `scope` when using the `oas generate` command.

Thanks in advance!

erunion (approved):
looks good to me!

erunion:
Published to 5.0.1, thanks @kschiu!

kschiu:
@erunion Thanks for the speedy review!

PR #301, opened 2020-11-05
refactor: pull change how schema is stored in the operation class

erunion:
## 🧰 What's being changed?

* [x] This surfaces the `parametersToJsonSchema` library we have directly onto the Operation class with a new `getParametersAsJsonSchema` method. Does the same thing, it's just now easier to use.
* [x] Refactors how we store operation data within the Operation class by no longer replacing the `this` scope with the operation schema, instead rather storing the operation schema to a new `schema` property.
  * **This is a breaking change.**


darrenyong (approved):
Looks good 😄

PR #302, opened 2020-11-06
feat: ship a compiled tooling lib dist

erunion:
## 🧰 What's being changed?

The `tooling` sublibrary in this package now ships with some async methods within the `Operation` class, and we've having trouble getting Webpack to compile it properly within our setup. So instead of messing with it further, we're going to instead ship a compiled dist of that library in the published package.

Kind of a pain in the ass, but it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ This library also exposes a set of tooling to help you manage OpenAPI definition
 require('oas/tooling')
 ```
 
+> To use a compiled version of this offering within a browser, you can load `oas/.tooling`
+
 Components available are the following:
 
 * `oas/tooling`: An OAS class can help you discover operations within a large OpenAPI file.
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   },
   "scripts": {
     "lint": "eslint .",
+    "prepare": "webpack",
     "pretest": "npm run lint && npm run prettier",
     "prettier": "prettier --list-different --write \"./**/**.js\"",
     "test": "jest --coverage",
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -65,6 +65,8 @@ This library also exposes a set of tooling to help you manage OpenAPI definition
 require('oas/tooling')
 ```
 
+> To use a compiled version of this offering within a browser, you can load `oas/.tooling`
```
rafegoldberg, line 68: ### 👍

package.json
```diff
@@ -32,6 +32,7 @@
   },
   "scripts": {
     "lint": "eslint .",
+    "prepare": "webpack",
```
rafegoldberg, line 35: Is there a reason to call this "prepare"? Maybe we can use "build" to be a little more conventional?

```suggestion
    "build": "webpack",
```

webpack.config.js
```diff
@@ -0,0 +1,30 @@
...
+  entry: ['./tooling/index.js'],
+  mode: 'production',
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            extends: './.babelrc',
```
rafegoldberg, line 14: So much config all for one little line...

rafegoldberg (approved):
Looks good to me.

package.json
```diff
@@ -78,12 +78,14 @@
...
     "coveragePathIgnorePatterns": [
+      "<rootDir>/.tooling",
       "<rootDir>/node_modules",
       "<rootDir>/__tests__/(.*)/__fixtures__/"
     ],
     "testPathIgnorePatterns": [
       "<rootDir>/__tests__/(.*)/__fixtures__/"
-    ]
+    ],
+    "transform": {}
```
rafegoldberg, line 88: Oh yeah, this... _great_ defaults Jest...

package.json
erunion: npm runs `prepare` by default when you publish a package. I can keep `build` but just have `prepare` run that.

PR #327, opened 2020-12-08
feat: addition of handling for dereferencing an oas definition

erunion:
## 🧰 What's being changed?

* [x] Implements a new `dereference` method into the `Oas` class that will dereference the OAS definition using [@apidevtools/json-schema-ref-parser](https://www.npmjs.com/package/@apidevtools/json-schema-ref-parser).
* [x] Removes the same `dereference` method from the `Operation` class. Since that work has been bubbled up into the parent OAS, and doesn't need to be run on individual operations, it didn't make sense to keep it around.
  * This'll require a major version bump since this is a breaking change.
* [x] Depromisified `Operation.getRequestBodyExamples` and  `Operation.getResponseExamples`. Since these methods no longer need to independently do their own OAS dereferencing, they no longer need to be promises.
* [x] Adds unit tests for our handling of circular schemas.


Dashron:
No real concerns with the code at first glance, but what was the purpose behind this change? What was wrong before this change, and fixed with this code?

Dashron:
talked with jon, this is a stepping stone to fix https://app.asana.com/0/1177947654981875/1199076686753793/f and some oas performance issues. PR looks fine to me.

Dashron (approved):

PR #336, opened 2020-12-15
feat: flattenSchema now expects a fully dereferenced schema

erunion:
## 🧰 What's being changed?

* [x] Revises the `flattenSchema` library to now expect a fully dereferenced schema (that can be done with `Oas.dereference()`. With this `flattenSchema` now only accepts a single parameter and also is no longer a promise.


Dashron (approved):
nothing jumps out at me, hard to test and verify without seeing the input, when we get everything hooked together we can test again.

PR #344, opened 2020-12-22
feat: implementing blocking on `oas.dereference` to prevent multiple executions

erunion:
## 🧰 What's being changed?

* [x] Implements some Promise blocking on `Oas.dereference()` to prevent multiple calls of `oas.dereference()` that are happening at the same time from running on top of each other.
* [x] Adds some state handling into the `Oas` class so it has awareness if `oas.dereference()` has been run and if `oas.dereference()` is called again it won't re-run the process.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,10 @@
   "root": true,
   "rules": {
     "no-console": "off",
-    "no-param-reassign": "off"
+    "no-param-reassign": "off",
+
+    "no-underscore-dangle": ["error", {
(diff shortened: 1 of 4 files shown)


erunion:
Initial testing with a 200KB OAS that uses `$ref` for just about everything and the performance returns on this are pretty crazy.

Before:
before

After:
after

erunion:
Merging this work into the `build/v9` branch so I can tag RC's for NPM off it.

https://github.com/readmeio/oas/pull/345

PR #345, opened 2020-12-22
feat: oas tooling process performance improvements

erunion:
## 🧰 What's being changed?

* [x] Implements some Promise blocking on `Oas.dereference()` to prevent multiple calls of `oas.dereference()` that are happening at the same time from running on top of each other. https://github.com/readmeio/oas/pull/344
* [x] Adds some state handling into the `Oas` class so it has awareness if `oas.dereference()` has been run and if `oas.dereference()` is called again it won't re-run the process. https://github.com/readmeio/oas/pull/344
* [x] Wraps our schema sample generation code in [memoizee](https://www.npmjs.com/package/memoizee) so identical schemas will only get processed once. https://github.com/readmeio/oas/pull/346
* [x] Adds handling to sample generation for malformed `array` schemas that are missing `items`. https://github.com/readmeio/oas/pull/345/commits/517c38045690ba0bf349176bd92f2f488544ac5a

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,10 @@
   "root": true,
   "rules": {
     "no-console": "off",
-    "no-param-reassign": "off"
+    "no-param-reassign": "off",
+
+    "no-underscore-dangle": ["error", {
+      "allow": ["_dereferencing", "_promises"]
+    }]
   }
 }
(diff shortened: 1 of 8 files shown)


erunion:
I've manually merged this work into `master`.

PR #358, opened 2021-01-12
feat: json schema examples

erunion:
## 🧰 What's being changed?

This adds support for pulling defined examples in OAS parameters and request bodies into our the JSON Schema that we generate with `oas.getParametersAsJsonSchema()`. 

https://swagger.io/docs/specification/adding-examples

> ⚠️  &nbsp; Since JSON Schema doesn't support the `example` prop that OAS has, we need to remap it as `examples`. So if something has an `example`, if that example is a primitive value it'll get remapped to `examples[example]`.

* Added support for parameter example
  * [x] Supports `example`
  * [x] Supports `examples`
* Added support for request body examples
  * Within a request body schema
    * [x] Supports `example`
    * [x] Supports `examples`
  * Defined at the request body alongside the `schema` object
    * [x] Supports `example`
    * [x] Supports `examples`

To facilitate this work I've also (mostly) rewritten `get-parameters-as-json-schema` to:

* [x] Make it easier to maintain. Previously we had individual JSON Schema parsing engines for parameters and request bodies/components. There is now just a single engine. Additionally instead of this new engine building up a schema from scratch as the previous ones did, this new one now adjusts the incoming schema to be JSON Schema-compatible. The end result is cleaner and more forgiving code.
* [x] Removed some funky handling and removal of `title` and `description` properties within some schemas. We had these exceptions in place due to how some of them look when used within `@readme/oas-form`, but it doesn't make sense to do that work here as all this library should care about is constructing valid JSON Schema. Once this work is complete I'll follow up another batch of work within `@readme/oas-form` to adjust how it handles these cases.
* [x] Remove handling of `$ref` pointers. Since we now dereference OAS definitions where we're using this tooling, we no longer need to worry about parsing these pointers, and not doing so allows us to better handle circular references without falling into an infinite loop. If a `$ref` is still present in the schema, we'll leave it but clean up the schema it's pointing to so if something using this library can handle circular refs (`@readme/oas-form` for example) it'll have the data necessary to do so.
  * I've left some existing test cases that surrounded parsing of `$ref` pointers for now, just added some `oas.dereference()` work around them to simulate a real-world case. Test expectations remain unchanged.

## 🧪 Testing

The `@readme/oas-form` part of this work to support these newly discovered examples is still forthcoming, but there are a **lot** of tests in this PR that should more than compensate for the multitude of ways you can define an example.


__tests__/tooling/__fixtures__/lib/json-schema.js
```diff
@@ -67,6 +67,14 @@ function buildSchemaDefault(opts) {
     props.default = opts.default ? opts.default : false;
```
erunion, line 67: This entire fixture file is still a huge mess and I want to at some point overhaul it into a general OAS fixture library but alas time.

__tests__/tooling/__snapshots__/operation.test.js.snap
```diff
@@ -66,10 +66,6 @@ Array [
               },
             },
             "type": "object",
-            "xml": Object {
```
erunion, line 69: `xml` was removed from this snapshot because we're now filtering out `xml` properties from our schemas.

tooling/lib/get-schema.js
```diff
@@ -9,11 +9,17 @@ module.exports = function getSchema(pathOperation, oas) {
   try {
     if (pathOperation.requestBody.content) {
       const type = Object.keys(pathOperation.requestBody.content)[0];
-      return { type, schema: pathOperation.requestBody.content[type].schema };
+
+      return {
+        type,
+        schema: pathOperation.requestBody.content[type],
```
erunion, line 15: Had to slightly change how this library works because it was discarding `example` and `examples` by just returning the content type schema.

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -15,173 +15,280 @@ const types = {
...
+  'example', // OpenAPI supports `example`, but we're mapping it to `examples` below.
+  'deprecated',
+];
+
+function encodePointer(str) {
+  // These characters need to be encoded to be allowed within a JSON pointer.
+  return str.replace('~', '~0').replace('/', '~1');
+}
 
-  const cleanupSchemaDefaults = (originType, obj, prevProp = false, prevProps = []) => {
```
erunion, line 22: This file is so much more readable now, good lord.

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -15,173 +15,291 @@ const types = {
...
-            // Only add a type if the previous prop isn't also named `properties`!
+    if (example !== undefined) {
+      break;
+    }
+  }
+
+  return example;
+}
+
+function constructSchema(data, prevSchemas = [], currentLocation = '') {
```
Dashron, line 98: Can you add some comments to this function describing what it does, and why we want to do it? This file is pretty dense and I think more comments will help clear it up.

Dashron (changes requested):
just could use some more comments in the get-parameters-as-json-schema. That file is pretty dense, even as an improvement to the old version.

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -15,173 +15,347 @@ const types = {
...
+ * Given a JSON pointer and an array of examples do a reverse search through them until we find the JSON pointer, or
+ * part of it, within the array.
+ *
+ * This function will allow you to take a pointer like `/tags/name` and return back `buster` from the following array:
+ *
+ * ```
+ *  [
+ *    {id: 20},
+ *    {tags: {name: 'doggie'}}
+ *  ]
```
Dashron, line 74: 🎆 🎆 🎆 examples! 🎆 🎆 🎆

Dashron: Where does the key `buster` come from in the above example?

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -15,173 +15,347 @@ const types = {
...
+    } else if ('items' in schema) {
+      schema.type = 'array';
+    } else {
+      // If we're processing a schema that has no types, no refs, and is just a lone schema, we should treat it at the
+      // bare minimum as a simple string so we make an attempt to generate valid JSON Schema.
+      schema.type = 'string';
+    }
+  }
+
+  // JSON Schema doesn't support OpenAPI-style examples so we need to reshape them a bit.
```
Dashron, line 177: I like this approach a lot, preferring json schema as the format of truth (which is the direction oas is going), and reworking oas examples into that format.

tooling/operation/get-parameters-as-json-schema.js
erunion: Just updated this example because it was totally wrong.

Dashron: The pointer should return `doggie` right, not buster?

tooling/operation/get-parameters-as-json-schema.js
erunion: Yeahhhh previously this library would take the OpenAPI-style and completely rebuild it into JSON Schema which caused us to have to add support for things we didn't account for. I much prefer to just assume it's okay and then add/remove whatever we need from there for stuff we can't handle or support.

tooling/operation/get-parameters-as-json-schema.js
erunion: Yeah my bad, will update it again.

Dashron:
The comments clear up this code so much, super easy to understand now. Looks great.

tooling/operation/get-parameters-as-json-schema.js
erunion: I found a bug in particular while rewriting this where we'd completely toss `properties` from an object under some specific circumstances. 😬

Dashron (approved):

PR #369, opened 2021-02-03
fix: issue where flattened array descriptions wouldn't be retained

erunion:
## 🧰 What's being changed?

This fixes a bug in our `flattenSchema` library where if an array description was located inside of the `items` object, we wouldn't retain it.

Fixes RM-232

## 🧪 Testing

```js
{
  type: 'object',
  properties: {
    products: {
      type: 'array',
      items: {
        description: 'product schema description', // This description wouldn't be retained and `products` 
                                                   // would be set with an `undefined` description.
        type: 'object',
        properties: {
          cash: {
            description: 'cash schema description',
            type: 'object',
            properties: {
              accounts: {
                type: 'array',
                items: {
                  description: 'account schema description',
                  type: 'object',
                  properties: {
                    id: {
                      type: 'string',
                      description: 'account id',
                    },
                  },
                },
              },
            },
          },
        },
      },
    },
  },
}
```

### DIFF
--- a/__tests__/tooling/lib/flatten-schema.test.js
+++ b/__tests__/tooling/lib/flatten-schema.test.js
@@ -523,3 +523,57 @@ describe('polymorphism cases', () => {
     });
   });
 });
+
+test('should retain descriptions on objects', () => {
+  const schema = {
+    type: 'object',
+    properties: {
+      products: {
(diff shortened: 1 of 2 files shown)


Dashron (approved):
makes sense

PR #370, opened 2021-02-05
fix: prevent us from crashing if `examples` is an empty object

erunion:
## 🧰 What's being changed?

Fixes a bug in our examples lookup work where if `examples: {}` was present we'd crash because we're expecting `value` to be inside one of the keyed properties of it.


rafegoldberg (approved):
This makes sense to me after playing around with the error!

domharrington (commented):
LGTM

PR #372, opened 2021-02-12
fix: adding support for parameter `content` schemas

erunion:
## 🧰 What's being changed?

With the rewrite of `getParametersAsJsonSchema()` #358 we lost support for parameter `content` schemas, something we never actually officially supported or had tests for.

* [x] Adds support for `content` within parameters
* [x] Fixes an edge case where if a `type` couldn't be determined for a schema (maybe `schema` or `content` aren't present for whatever reason, or aren't in the right shape), then we'll default to treating the schema as a `string`.

```json
{
  "name": "groupId",
  "description": "Filter the data by groupId",
  "in": "query",
  "required": false,
  "content": { "application/json": { "schema": { "type": "string" } } }
}
```

Fixes [RM-311](https://linear.app/readme-io/issue/RM-311/we-are-trying-to-render-params-we-dont-understand)

## 🧪 Testing

I've added in a bunch of different test cases, but the ways we support this and pick up schemas within `content` are as follows:

* If there is just a single content type in `content` we'll pick that up and use the schema inside it.
* If there are two non-`application/json` content types in `content` we'll use the first in the list.
* If there are two content types in `content`, and one of them is JSON-like (`application/json`, `text/json`, `application/vnd.something+json`) we'll use that over the other.

### DIFF
--- a/__tests__/tooling/operation/get-parameters-as-json-schema.test.js
+++ b/__tests__/tooling/operation/get-parameters-as-json-schema.test.js
@@ -255,6 +255,105 @@ describe('parameters', () => {
     expect(oas.operation('/', 'get').getParametersAsJsonSchema()[0].schema.properties.options.type).toBe('array');
   });
 
+  it('should set a type to `string` if neither `schema` or `current` are present', () => {
+    const oas = createOas({
+      parameters: [
+        {
+          name: 'userId',
+          in: 'query',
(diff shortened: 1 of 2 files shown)


Dashron (approved):

Dashron (approved):
tests & code look good

PR #373, opened 2021-02-12
feat: populate params jwt defaults

darrenyong:
## 🧰 What's being changed?
I've added a field on the Operation class called `jwtDefaults` that will store the parameter defaults that's sent from JWT Variables.

Then, we're going to pass the object through to `getParametersAsJsonSchema()` and subsequently, `constructSchema()`. In `constructSchema()`, we'll do the checking via `jsonPointer` and if the parameters exist, we're going to add it on `schema.default`.

Describe in detail what this PR is for.
This feature will allow users to add parameter defaults in the JWT User Data object.

## 🧪 Testing
Check that parameters in the JWT User Data object are added to endpoints that have the same parameters. If the parameters don't exist on the endpoint, it shouldn't be added as a default!


tooling/operation.js
```diff
@@ -8,11 +8,12 @@ const getResponseExamples = require('./operation/get-response-examples');
 const matchesMimeType = require('./lib/matches-mimetype');
 
 class Operation {
-  constructor(oas, path, method, operation) {
+  constructor(oas, path, method, operation, jwtDefaults) {
```
erunion, line 11: It feels weird tying our JWT naming into this. Maybe this parameter and `this.jwtDefaults` should be called `globalDefaults` instead? Or instead of attaching it to the `Operation` class, it should just be passed into `parametersAsJsonSchema(jwtDefaults)`?

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -322,11 +328,20 @@ function constructSchema(data, prevSchemas = [], currentLocation = '') {
...
     if (polyType in schema && Array.isArray(schema[polyType])) {
       schema[polyType].forEach((item, idx) => {
-        schema[polyType][idx] = constructSchema(item, prevSchemas, `${currentLocation}/${idx}`);
+        schema[polyType][idx] = constructSchema(item, prevSchemas, `${currentLocation}/${idx}`, jwtDefaults);
       });
     }
   });
 
+  // Check to see if defaults and currentLocation exist and it's not just an empty object and blank
+  // If currentLocation is blank, jsonPointer returns the whole object so it will shove everything into formData
```
erunion, line 337: Care to slightly rewrite this to explain JWT defaults in here and why we might want to add them as a schema default? Also if you can wrap variables in backticks like I've done in other code comments in this file that'd be awesome.

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -322,11 +328,20 @@ function constructSchema(data, prevSchemas = [], currentLocation = '') {
...
+  if (jwtDefaults && Object.keys(jwtDefaults).length > 0 && currentLocation) {
+    const userJwtDefault = jsonpointer.get(jwtDefaults, currentLocation);
```
erunion, line 339: There's some cases where `jsonpointer.get()` will throw an exception so you should catch that here and just not add a default if it happens. Check out where I'm doing this in `searchForExampleByPointer`, there's some comments explaining why it might happen.

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -365,7 +380,7 @@ function getRequestBody(operation, oas) {
     examples.push({ examples: requestBody.examples });
   }
 
-  const cleanedSchema = constructSchema(requestBody.schema, examples);
+  const cleanedSchema = constructSchema(requestBody.schema, examples, undefined, jwtDefaults);
```
erunion, line 383: `currentLocation` should probably always just be an empty string.

```suggestion
  const cleanedSchema = constructSchema(requestBody.schema, examples, '', jwtDefaults);
```

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -375,7 +390,12 @@ function getRequestBody(operation, oas) {
         }
 
         Object.keys(oas.components[componentType]).forEach(schemaName => {
-          components[componentType][schemaName] = constructSchema(oas.components[componentType][schemaName]);
+          components[componentType][schemaName] = constructSchema(
+            oas.components[componentType][schemaName],
+            undefined,
```
erunion, line 395: The `prevSchemas` argumet defaults to an empty array so you should pass that in here instead of `undefined`.

```suggestion
           [],
```

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -375,7 +390,12 @@ function getRequestBody(operation, oas) {
...
+            undefined,
```
erunion, line 396: `currentLocation` should probably always be a string.

```suggestion
           '',
```

tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -433,7 +453,7 @@ function getParameters(path, operation, oas) {
 
     const properties = parameters.reduce((prev, current) => {
       const schema = {
-        ...(current.schema ? constructSchema(current.schema) : {}),
+        ...(current.schema ? constructSchema(current.schema, undefined, undefined, jwtDefaults) : {}),
```
erunion, line 456:
```suggestion
        ...(current.schema ? constructSchema(current.schema, [], '', jwtDefaults) : {}),
```

erunion (changes requested):
Some minor nits but otherwise this looks great!

tooling/operation.js
darrenyong: I just extracted it out of the Operation class and decided to pass it through `getParametersAsJsonSchema()`!

tooling/operation.js
```diff
@@ -234,8 +234,8 @@ class Operation {
    *
    * @return {array}
    */
-  getParametersAsJsonSchema() {
-    return getParametersAsJsonSchema(this.path, this.schema, this.oas);
+  getParametersAsJsonSchema(globalDefaults) {
```
erunion, line 237: Can you add this parameter, and a short description of what it's for, into the docblock above?

erunion (approved):
🥳

tooling/operation.js
darrenyong: done!

Dashron:
Code looks fine, do you have an example we could test?

And note for the future, I know we update the user provided schema in other locations, but I wonder if that's going to bite us in the future.

__tests__/tooling/operation/get-parameters-as-json-schema.test.js
```diff
@@ -1304,6 +1304,36 @@ describe('defaults', () => {
...
+      it('should not add jwtDefaults if there are no matches', async () => {
+        const schema = new Oas(petstore);
+        await schema.dereference();
+        const operation = schema.operation('/pet', 'post');
+        operation.jwtDefaults = {
+          fakeParameter: {
+            id: 4,
+            name: 'Testing',
+          },
+        };
```
erunion, line 1316: This test needs to get updated to not set `jwtDefaults` on the `operation`.

darrenyong: updated!

darrenyong:
> Code looks fine, do you have an example we could test?
> 
> And note for the future, I know we update the user provided schema in other locations, but I wonder if that's going to bite us in the future.

@Dashron, I'll provide test cases when I open the PR in the api-explorer repo, since this stuff requires some of the code from the API Explorer to test properly!

PR #374, opened 2021-02-18
fix: edge case where allOfs would receive a `type` when one wasn't present

erunion:
## 🧰 What's being changed?

@rafegoldberg stumbled on this in the [AWS Workspaces](https://preview.readme.io/?selected=https%3A%2F%2Fapi.apis.guru%2Fv2%2Fspecs%2Famazonaws.com%2Fworkspaces%2F2015-04-08%2Fopenapi.json) definition while testing out https://github.com/readmeio/api-explorer/pull/1179, but we have an edge case right now where if you're using `allOf` to dynamically add a `description`, and only that, to another schema, we'll add `type: string` to the description object.

For example:

```js
allOf: [
  { 
    type: 'array', 
    items: { 
      type: 'string' 
    }
  }, 
  { description: 'Parameter description' } // This ends up receiving `type: string`.
],
```

Reason why this ends up failing in the API Explorer is because when `@readme/oas-form` attempts to merge the schemas in the `allOf`, `array` and `string` collide and [json-schema-merge-allof](https://www.npmjs.com/package/json-schema-merge-allof) throws an error and instead of merging the schemas together `@readme/oas-form` instead returns the original schemas sans the `allOf` -- resulting in a `type`-less schema.

I continue to be eternally grateful that I took the time to rewrite `getParametersAsJsonSchema` in December because fixing this kind of thing in the old code would have been a Hot Mess.

## 🧪 Testing

See attached unit tests, but here's the before and after of this fix in the API Explorer:

#### Before
Screen Shot 2021-02-17 at 6 40 43 PM

#### After
Screen Shot 2021-02-17 at 6 41 31 PM

### DIFF
--- a/tooling/operation/get-parameters-as-json-schema.js
+++ b/tooling/operation/get-parameters-as-json-schema.js
@@ -178,10 +178,17 @@ function searchForExampleByPointer(pointer, examples = []) {
  * @link https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md
  * @param {Object} data
  * @param {Object[]} prevSchemas
- * @param {string} currentLocation
+ * @param {String} currentLocation
  * @param {Object} globalDefaults
+ * @param {Boolean} isPolymorphicAllOfChild
  */
-function constructSchema(data, prevSchemas = [], currentLocation = '', globalDefaults) {
+function constructSchema(
+  data,
+  prevSchemas = [],
+  currentLocation = '',
+  globalDefaults,
+  isPolymorphicAllOfChild = false
+) {
   const schema = { ...data };
 
   // If this schema contains a `$ref`, it's circular and we shouldn't try to resolve it. Just return and move along.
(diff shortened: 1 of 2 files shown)


tooling/operation/get-parameters-as-json-schema.js
```diff
@@ -197,6 +204,15 @@ function constructSchema(data, prevSchemas = [], currentLocation = '', globalDef
       schema.type = 'object';
     } else if ('items' in schema) {
       schema.type = 'array';
+    } else if (isPolymorphicAllOfChild) {
```
erunion, line 207: Since this shapeless case only happens with an `allOf` where each schema within that block is merged together, we shouldn't apply this logic to `oneOf` and `anyOf` polymorphic schemas.

rafegoldberg (approved):
Damn, the turnaround between my comment and this PR was quick af. I haven't run this code, but it code reads well and test cases check out per your screenshots/description of the issue!￼

tooling/operation/get-parameters-as-json-schema.js
Dashron: am I reading this right that we're just leaving this with an empty if block?

Is this preferred over switching the final else to else if not polymporphic? I'm not sure I have an opinion just yet.

erunion: Yeah this is just a no-op. Figured it was cleaner and easier to read this way.

PR #387, opened 2021-03-11
feat: support polymorphistic response examples

darrenyong:
## 🧰 What's being changed?
| [Linear Ticket](https://linear.app/readme-io/issue/CX-183/allof-property-causes-display-issue) |
|------------------------------------------------------------------------------------------------|

This PR adds support to render responses that are currently using polymorphism. 

The operation with the response:
image

The respective components:
image

We currently render the response example in the above operation like so:
image

As you can see, the `reach` parameter is using `allOf`, although it is a weird use of `allOf`, but we don't render the object that `reach` is referencing. 

With these changes, we're correctly grabbing the `properties` object from the schema to show the following:
image

## 🧪 Testing
1. Use the OA file in the Linear ticket or the one on my [test project](https://darren.readme.io/openapi/60497c0392b39a0065e98136) for a condensed down version. 
2. Upload the file
3. You should see the updated response example when the package has been bumped in the explorer.

I did run into an issue where `mergeAllOf` couldn't merge the `allOf` schema so I am making it return the first item in the polymorphism array. Here's an example of an `allOf` schema that was having trouble:
image

### DIFF
--- a/tooling/samples/index.js
+++ b/tooling/samples/index.js
@@ -42,23 +42,21 @@ const sampleFromSchema = (schema, config = {}) => {
   let { type } = objectifySchema;
 
   const hasPolymorphism = usesPolymorphism(objectifySchema);
-  if (hasPolymorphism) {
-    if (hasPolymorphism === 'allOf') {
-      try {
-        return sampleFromSchema(
-          mergeAllOf(objectifySchema, {
-            resolvers: {
-              // Ignore any unrecognized OAS-specific keywords that might be present on the schema (like `xml`).
-              defaultResolver: mergeAllOf.options.resolvers.title,
-            },
-          })
-        );
-      } catch (error) {
-        return undefined;
-      }
-    } else {
-      return sampleFromSchema(objectifySchema[hasPolymorphism]);
+  if (hasPolymorphism === 'allOf') {
+    try {
+      return sampleFromSchema(
+        mergeAllOf(objectifySchema, {
+          resolvers: {
+            // Ignore any unrecognized OAS-specific keywords that might be present on the schema (like `xml`).
+            defaultResolver: mergeAllOf.options.resolvers.title,
+          },
+        })
+      );
+    } catch (error) {
+      return undefined;
     }
+  } else if (hasPolymorphism) {
+    return sampleFromSchema(objectifySchema[hasPolymorphism]);
   }
 
   const { example, additionalProperties, properties, items } = objectifySchema;
--- a/tooling/samples/index.js
+++ b/tooling/samples/index.js
@@ -5,7 +5,7 @@
  * @link https://github.com/swagger-api/swagger-ui/blob/master/src/core/plugins/samples/fn.js
  */
 
-const { objectify, hasPolymorphism, isFunc, normalizeArray, deeplyStripKey } = require('./utils');
+const { objectify, usesPolymorphism, isFunc, normalizeArray, deeplyStripKey } = require('./utils');
 const memoize = require('memoizee');
 const mergeAllOf = require('json-schema-merge-allof');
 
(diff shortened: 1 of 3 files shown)


tooling/samples/index.js
```diff
@@ -39,8 +39,18 @@ const primitive = schema => {
...
+
+  const usesPolymorphism = hasPolymorphism(objectifySchema);
+  if (usesPolymorphism === 'allOf') {
+    const mergedAllOf = mergeAllOf(objectifySchema);
+    properties = mergedAllOf.properties;
+    items = mergedAllOf.items;
+  } else if (usesPolymorphism) {
+    properties = objectifySchema[usesPolymorphism][0].properties;
+    items = objectifySchema[usesPolymorphism][0].items;
+  }
```
erunion, line 53: Weird, we already have handling for `allOf` a few lines down https://github.com/readmeio/oas/blob/master/tooling/samples/index.js#L61-L74

How come that code wasn't getting picked up?

darrenyong: Ah, if I turn:
```
"reach": {
  "allOf": [
    {
      "$ref": "#/components/schemas/SegmentReach"
    }
  ],
  "type": "object"
},
```
to
```
"reach": {
  "allOf": [
    {
      "$ref": "#/components/schemas/SegmentReach"
    }
  ]
},
```
it correctly renders.

Dashron: Interesting! I looked around and the presence of a type property as a sibling to a ref (or an all of) is completely valid.

So we need to support this!

Maybe we can consolidate this logic with the existing allOf logic though?

erunion: Yeah after looking at this a little closer I think it might be more clear what's going on if when we detect polymorphism we recurse into that object instead of overwriting `items` and `properties`, similar to how we're already handling `allOf` when there's no `type` present. Something like:

```js
if (hasPolymorphism) {
  if (hasPolymorphism === 'allOf') {
    try {
      return sampleFromSchema(
        mergeAllOf(objectifySchema, {
          resolvers: {
            // Ignore any unrecognized OAS-specific keywords that might be present on the schema (like `xml`).
            defaultResolver: mergeAllOf.options.resolvers.title,
          },
        })
      );
    } catch (err) {
      // Unable to merge the schema for whatever reason so let's just ignore it and do a no-op here.
      return undefined
    }
  } else {
    return sampleFromSchema(objectifySchema[hasPolymorphism]);
  }
}
```

darrenyong: updated!

tooling/samples/index.js
```diff
@@ -40,8 +40,28 @@ const primitive = schema => {
 const sampleFromSchema = (schema, config = {}) => {
   const objectifySchema = objectify(schema);
   let { type } = objectifySchema;
-  const { example, properties, additionalProperties, items } = objectifySchema;
 
+  const hasPolymorphism = usesPolymorphism(objectifySchema);
+  if (hasPolymorphism) {
```
Dashron, line 45: I don't think you need this check, it seems to be covered by the following if.

tooling/samples/index.js
Dashron: I don't think we need this and the allOf handling from line 81 to 94. Worth testing, and deleting if we can.

Dashron (changes requested):

tooling/samples/index.js
darrenyong: @Dashron, I added Line 45 in case there's `anyOf` or `oneOf` and there is a `type` present on the schema. The if on Line 46 will only check if there's an `allOf` and the else if will grab the rest!

Although, now that I'm typing this out, I'm wondering if I can add an `or` on the `(!type)` portion 🤔

darrenyong: hey @Dashron, I updated this portion! GitHub isn't letting me respond to [this comment](https://github.com/readmeio/oas/pull/387/files/b23220047ef59525ab40a2061daf9211df70c0c5#discussion_r595424390), but I've refactored it so it handles schemas with `anyOf` or `oneOf` and a type.

tooling/samples/index.js
darrenyong: When I initially wrote this, I thought this portion would handle the `anyOf` and `oneOf` schemas with `type` cases, but I didn't correctly test it. I updated it and confirmed it works for those cases and added a test for it.

Dashron (approved):
👌👀👌👀👌👀👌👀👌👀 good shit go౦ԁ sHit👌 thats ✔ some good👌👌shit right👌👌there👌👌👌 right✔there ✔✔if i do ƽaү so my self 💯 i say so 💯 thats what im talking about right there right there (chorus: ʳᶦᵍʰᵗ ᵗʰᵉʳᵉ) mMMMMᎷМ💯 👌👌 👌НO0ОଠOOOOOОଠଠOoooᵒᵒᵒᵒᵒᵒᵒᵒᵒ👌 👌👌 👌 💯 👌 👀 👀 👀 👌👌Good shit

PR #388, opened 2021-03-17
feat: add new methods for server variable support

domharrington:
## 🧰 What's being changed?

Oas class changes to support editable server urls component

## 🧪 Testing

See unit tests

### DIFF
--- a/__tests__/tooling/index.test.js
+++ b/__tests__/tooling/index.test.js
@@ -62,7 +62,7 @@ describe('#splitUrl()', () => {
       }).splitUrl()
     ).toStrictEqual([
       { key: 'https://example.com/-0', type: 'text', value: 'https://example.com/' },
-      { key: 'path-1', type: 'variable', value: 'path' },
+      { key: 'path-1', type: 'variable', value: 'path', description: undefined, enum: undefined },
     ]);
   });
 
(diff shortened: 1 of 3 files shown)


__tests__/tooling/index.test.js
```diff
@@ -42,6 +42,138 @@ describe('#url()', () => {
...
+      new Oas({
+        servers: [{ url: 'https://example.com/{path}', variables: { path: { enum: ['v1', 'v2'] } } }],
+      }).splitUrl()[1].enum
+    ).toStrictEqual(['v1', 'v2']);
+  });
+});
+
+describe('#variables([selected])', () => {
+  it('should return with list of variables', () => {
+    const variables = { path: { description: 'path description' } };
```
erunion, line 115: Can you add `type: int` to these variable examples so they look more real?

erunion (approved):
Minor test comment but otherwise looks good! I can get you a build first thing Monday morning.

Dashron (approved):

__tests__/tooling/index.test.js
domharrington: I dont think they have types? https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#serverVariableObject Unless i'm misunderstanding here.

erunion: Ah no you're right, I was looking at this on mobile and was thinking I was looking at path params, not path vars.

erunion:
@domharrington Published to 10.3.0

PR #406, opened 2021-04-01
refactor: extract part of prepareSecurity() function into a new function

kanadgupta:
## 🧰 What's being changed?

As part of the changes I made in https://github.com/readmeio/readme/pull/3811, I wrote a little helper function [here](https://github.com/readmeio/readme/blob/b2b01ecbfe0a3c82c847aefb3ce0eaf711afe7b3/packages/react/src/ui/APIAuth/index.jsx#L70-L72) that uses some of the logic we use in [`operation.prepareSecurity()`](https://github.com/readmeio/oas/blob/7d5c1575ef1dcf6b9913657c89f3c5dbaf404bb6/tooling/operation.js#L122) for validating and setting types for security schemes (while maintaining the array structure so multiple auth grouping is preserved). That function should probably live in this library instead.

I simply extracted that part of the `prepareSecurity()` function into its own function (and added a tiny flag for filtering out junk)—hence I'm considering this a refactor, but it could arguably be called a feature?

We might be able to deprecate `operation.prepareSecurity()` because I'm not sure what that's used for now, but I'm leaving it in here since it feels unnecessary to introduce breaking changes.

## 🧪 Testing

Added some test coverage for the new function, confirmed that the changes don't break `operation.prepareSecurity()`.

### DIFF
--- a/tooling/operation.js
+++ b/tooling/operation.js
@@ -66,6 +66,11 @@ class Operation {
     return matchesMimeType.xml(this.getContentType());
   }
 
+  /**
+   * @returns The securities associated with this operation.
+   * If none are defined at an operation level, the securities for the entire `oas` are returned
+   * (with an empty array as a fallback).
+   */
   getSecurity() {
     if (!('components' in this.oas) || !('securitySchemes' in this.oas.components)) {
       return [];
@@ -74,6 +79,14 @@ class Operation {
     return this.schema.security || this.oas.security || [];
   }
 
+  /**
+   * @param {boolean} filter Optional flag that filters out invalid/nonexistent security schemes,
+   * rather than returning `false`.
+   * @returns An array of arrays of objects of grouped security schemes.
+   * The inner array determines and-grouped security schemes, the outer array determines or-groups.
+   *
+   * See https://swagger.io/docs/specification/authentication/#multiple
+   */
   getSecurityWithTypes(filter = false) {
     const securityRequirements = this.getSecurity();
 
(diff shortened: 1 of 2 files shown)


kanadgupta:
@erunion this will be far easier to review if you hide whitespace changes lol

tooling/operation.js
```diff
@@ -74,60 +79,80 @@ class Operation {
...
-  prepareSecurity() {
+  /**
+   * @param {boolean} filter Optional flag that, when set to `true`,
+   * filters out invalid/nonexistent security schemes, rather than returning `false`.
+   * @returns An array of arrays of objects of grouped security schemes.
+   * The inner array determines and-grouped security schemes, the outer array determines or-groups.
+   *
+   * See https://swagger.io/docs/specification/authentication/#multiple
+   */
+  getSecurityWithTypes(filter = false) {
```
erunion, line 90:
```suggestion
  getSecurityWithTypes(filterInvalid = false) {
```

tooling/operation.js
```diff
@@ -74,60 +79,80 @@ class Operation {
     return this.schema.security || this.oas.security || [];
   }
 
-  prepareSecurity() {
+  /**
+   * @param {boolean} filter Optional flag that, when set to `true`,
```
erunion, line 83:
```suggestion
   * @param {boolean} filterInvalid Optional flag that, when set to `true`,
```

erunion (approved):
LGTM

PR #407, opened 2021-04-06
Adds the ability to extract response data similar to getParametersAsJsonSchema

Dashron:
## 🧰 What's being changed?

We're hoping to render response schemas more similarly to request schemas, so this adds the ability to extract response data that looks like our request schemas.

## 🧪 Testing

See the tests!

### DIFF
--- a/__tests__/tooling/operation.test.js
+++ b/__tests__/tooling/operation.test.js
@@ -727,6 +727,11 @@ describe('#getResponseByStatusCode()', () => {
 describe('#getAllResponseStatusCodes()', () => {
   it('should return all valid status codes for a response', () => {
     const operation = new Oas(petstore).operation('/pet/findByStatus', 'get');
-    expect(operation.getAllResponseStatusCodes()).toStrictEqual(['200', '400']);
+    expect(operation.getResponseStatusCodes()).toStrictEqual(['200', '400']);
+  });
+
+  it('should return an empty array if there are no responses', () => {
+    const operation = new Oas(petstore).operation('/pet/findByStatus', 'doesnotexist');
+    expect(operation.getResponseStatusCodes()).toStrictEqual([]);
   });
 });
--- a/__tests__/tooling/operation.test.js
+++ b/__tests__/tooling/operation.test.js
@@ -724,7 +724,7 @@ describe('#getResponseByStatusCode()', () => {
   });
 });
 
-describe('#getAllResponseStatusCodes()', () => {
+describe('#getResponseStatusCodes()', () => {
   it('should return all valid status codes for a response', () => {
     const operation = new Oas(petstore).operation('/pet/findByStatus', 'get');
     expect(operation.getResponseStatusCodes()).toStrictEqual(['200', '400']);
--- a/__tests__/tooling/operation.test.js
+++ b/__tests__/tooling/operation.test.js
@@ -723,3 +723,10 @@ describe('#getResponseByStatusCode()', () => {
     });
   });
 });
+
+describe('#getAllResponseStatusCodes()', () => {
+  it('should return all valid status codes for a response', () => {
+    const operation = new Oas(petstore).operation('/pet/findByStatus', 'get');
+    expect(operation.getAllResponseStatusCodes()).toStrictEqual(['200', '400']);
+  });
+});
--- a/tooling/operation.js
+++ b/tooling/operation.js
@@ -3,6 +3,7 @@ const kebabCase = require('lodash.kebabcase');
 
 const findSchemaDefinition = require('./lib/find-schema-definition');
 const getParametersAsJsonSchema = require('./operation/get-parameters-as-json-schema');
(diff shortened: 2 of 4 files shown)


tooling/operation.js
```diff
@@ -265,6 +266,24 @@ class Operation {
...
+  getResponseAsJsonSchema(statusCode) {
+    return getResponseAsJsonSchema(this, this.oas, statusCode);
+  }
+
+  /**
+   * Get an array of all valid response status codes for this operation
+   * @param {*} statusCode
+   * @returns
+   */
+  getAllResponseStatusCodes() {
```
erunion, line 283: Just so it's a little more consistent with the other `getResponse*` methods:

```suggestion
  getResponseStatusCodes() {
```

tooling/operation/get-response-as-json-schema.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {String} statusCode
+ * @returns Array<{schema: Object, type: string, label: string}>
+ */
+module.exports = function getResponseAsJsonSchema(operation, oas, statusCode) {
+  const response = operation.getResponseByStatusCode(statusCode);
+  const jsonSchema = [];
+
+  // TODO: This lookup should be in the operation
+  if (response.content?.['application/json'].schema) {
+    const actualSchema = response?.content?.['application/json'].schema;
```
erunion, line 53: This is going to lose a lot of the special cases and `example` and `examples` handling that we've built in `get-parameters-as-json-schema`. The core `constructSchema` method in that library is already being exported so you should be able to pull that in and run:

```
builtSchema = constructSchema(response.content['application/json'].schema)
```

tooling/operation/get-response-as-json-schema.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {Oas} oas
+ * @param {String} statusCode
+ * @returns Array<{schema: Object, type: string, label: string}>
+ */
+module.exports = function getResponseAsJsonSchema(operation, oas, statusCode) {
+  const response = operation.getResponseByStatusCode(statusCode);
+  const jsonSchema = [];
+
+  // TODO: This lookup should be in the operation
+  if (response.content?.['application/json'].schema) {
```
erunion, line 52: Hardcoding to `application/json` will prevent us from supporting vendor mimes like `application/vnd.github.v3.star+json`.

tooling/operation/get-response-as-json-schema.js
Dashron: So I've wrapped `response?.content?.['application/json'].schema;` in constructSchema, but I'm not sure having builtSchema equal that is correct. 

It might just be a naming issue, but I'm not really sure what to call builtschema, since it's mirroring this format: https://github.com/readmeio/oas/blob/master/tooling/operation/get-parameters-as-json-schema.js#L539. That is schema-like, but has a schema sub-section? 

For now I renamed it schema wrapper like I have in my tests.

__tests__/tooling/operation.test.js
```diff
@@ -723,3 +723,10 @@ describe('#getResponseByStatusCode()', () => {
     });
   });
 });
+
+describe('#getAllResponseStatusCodes()', () => {
```
erunion, line 727:
```suggestion
describe('#getResponseStatusCodes()', () => {
```

__tests__/tooling/operation.test.js
```diff
@@ -723,3 +723,10 @@ describe('#getResponseByStatusCode()', () => {
...
+  it('should return all valid status codes for a response', () => {
+    const operation = new Oas(petstore).operation('/pet/findByStatus', 'get');
+    expect(operation.getAllResponseStatusCodes()).toStrictEqual(['200', '400']);
```
erunion, line 730:
```suggestion
    expect(operation.getResponseStatusCodes()).toStrictEqual(['200', '400']);
```

tooling/operation.js
```diff
@@ -265,6 +266,24 @@ class Operation {
...
+    return getResponseAsJsonSchema(this, this.oas, statusCode);
+  }
+
+  /**
+   * Get an array of all valid response status codes for this operation
+   * @param {*} statusCode
+   * @returns
+   */
+  getResponseStatusCodes() {
+    return Object.keys(this.schema.responses);
```
erunion, line 284: Just in case there's no `responses`:

```suggestion
    return this.schema.responses ? Object.keys(this.schema.responses) : [];
```

tooling/operation/get-response-as-json-schema.js
erunion: Ok I think how you have it now is good.

erunion:
Not sure how comprehensive we want these tests to be since `constructSchema` is pretty well covered but I think what you've got here so far is a great start at least. Happy to tag a release after these other couple issues are sorted.

erunion (approved):

PR #421, opened 2021-05-04
fix: edgecase where server urls would get corrupted with normalized versions

erunion:
## 🧰 What's being changed?

This was being surfaced in our `APIServers` component where because we do `findOperation` lookups in a handful of spots to be able to handle servers with variables, our `findOperationMatches` code was actually overwriting the core `servers` data within the OAS, resulting in complete data loss of these servers variables. 

Because apparently of this:

```js
const servers = [{ url: 'https://example.com' }];
const found = servers.find(server => server.url === 'https://example.com');
found.url = 'https://google.com';

console.log(servers[0].url); // https://google.com
```

🤡 

## 🧪 Testing

See the unit test I added.

### DIFF
--- a/__tests__/tooling/index.test.js
+++ b/__tests__/tooling/index.test.js
@@ -510,6 +510,25 @@ describe('#findOperation()', () => {
       },
     });
   });
+
+  it('should not overwrite the servers in the core OAS while looking for matches', () => {
+    const oas = new Oas(serverVariables);
+    const uri = 'https://demo.example.com:443/v2/post';
+    const method = 'post';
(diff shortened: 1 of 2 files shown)


rafegoldberg (approved):
So not to say I understand the *implications* of this issue, but to clarify: essentially this was a "set by reference" issue? Either way, the code + test case seem sound to me!

domharrington (approved):
This seems great, and nice test case. So findOperation was clobbering the `url` property of the actual Oas.server object? Makes sense.

PR #422, opened 2021-05-06
fix: quirks with operation discovery on servers that have variables

erunion:
## 🧰 What's being changed?

This is a weird one!

Say you have an OAS that has a server URL of `https://{region}.node.example.com/v14` and a logged in user JWT with `region` set to `us`. `region` is an enum that can be either `us` or `eu`. Now say you're trying to find an operation for https://eu.node.example.com/v14/api/esm. Won't work.

The way we do operation discovery (with `getOperation()` and `findOperation()`) is we run through a checklist of things to look for:

1. First we replace the server URL with the users variables so https://{region}.node.example.com/v14 becomes https://us.node.example.com/v14
2. Then we look at the origin of the URL we're looking for: https://us.node.example.com
3. If that fails we fallback to the hostname we're looking for: us.node.example.com
4. If we've got a match, split our found server URL off the URL we're looking for. So https://us.node.example.com/v14 is removed from https://us.node.example.com/v14/api/esm resulting in the path we should pull operations for: `/api/esm`.

Now where this breaks is that if either the user is logged out (no JWT variables), or you're looking at an old log you want to pull records for (like in metrics) where you might have https://eu.node.example.com/v14/api/esm. Since `eu` **isn't** the default value for `region` we'll never be able to find `/api/esm` without regexing `https://(.*).node.example.com/v14` out.

That's what this does. 😰 

## 🧪 Testing

I reorganized some tests around in here, but peep the ones I added to the `getOperation()` block.

### DIFF
--- a/__tests__/tooling/index.test.js
+++ b/__tests__/tooling/index.test.js
@@ -53,7 +53,7 @@ describe('#url([selected])', () => {
     expect(new Oas({ servers: [{ url: 'https://example.com' }] }).url(10)).toBe('https://example.com');
   });
 
-  describe('variable replacement', () => {
+  describe('server variables', () => {
     const oas = new Oas({
       servers: [
         {
@@ -81,6 +81,69 @@ describe('#url([selected])', () => {
       expect(oas.url(0, { basePath: 'v3', name: 'subdomain', port: '8080' })).toBe(
         'https://subdomain.example.com:8080/v3'
       );
+    });
+
+    it('should use defaults', () => {
+      expect(
+        new Oas({
+          servers: [{ url: 'https://example.com/{path}', variables: { path: { default: 'path' } } }],
+        }).url()
+      ).toBe('https://example.com/path');
+    });
+
+    it('should use user variables over defaults', () => {
+      expect(
+        new Oas(
+          {
+            servers: [{ url: 'https://{username}.example.com', variables: { username: { default: 'demo' } } }],
+          },
+          { username: 'domh' }
+        ).url()
+      ).toBe('https://domh.example.com');
+    });
+
+    it('should fetch user variables from keys array', () => {
+      expect(
+        new Oas(
+          {
+            servers: [{ url: 'https://{username}.example.com', variables: { username: { default: 'demo' } } }],
+          },
+          { keys: [{ name: 1, username: 'domh' }] }
+        ).url()
+      ).toBe('https://domh.example.com');
+    });
+
+    it('should look for variables in selected server', () => {
+      expect(
+        new Oas({
+          servers: [
+            { url: 'https://{username1}.example.com', variables: { username1: { default: 'demo1' } } },
+            { url: 'https://{username2}.example.com', variables: { username2: { default: 'demo2' } } },
+          ],
+        }).url(1)
+      ).toBe('https://demo2.example.com');
+    });
+
+    it.skip('should fetch user variables from selected app', () => {
+      expect(
+        new Oas(
+          {
+            servers: [{ url: 'https://{username}.example.com', variables: { username: { default: 'demo' } } }],
+          },
+          {
+            keys: [
+              { name: 1, username: 'domh' },
+              { name: 2, username: 'readme' },
+            ],
+          },
+          2
+        ).url()
+      ).toBe('https://readme.example.com');
+    });
+
+    // Test encodeURI
+    it('should pass through if no default set', () => {
+      expect(new Oas({ servers: [{ url: 'https://example.com/{path}' }] }).url()).toBe('https://example.com/{path}');
     });
   });
 });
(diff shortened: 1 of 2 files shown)


__tests__/tooling/index.test.js
```diff
@@ -82,6 +82,69 @@ describe('#url([selected])', () => {
...
+        new Oas({
+          servers: [
+            { url: 'https://{username1}.example.com', variables: { username1: { default: 'demo1' } } },
+            { url: 'https://{username2}.example.com', variables: { username2: { default: 'demo2' } } },
+          ],
+        }).url(1)
+      ).toBe('https://demo2.example.com');
+    });
+
+    it.skip('should fetch user variables from selected app', () => {
```
ilias-t, line 127: skip?

__tests__/tooling/index.test.js
```diff
@@ -82,6 +82,69 @@ describe('#url([selected])', () => {
         'https://subdomain.example.com:8080/v3'
       );
     });
+
+    it('should use defaults', () => {
+      expect(
+        new Oas({
+          servers: [{ url: 'https://example.com/{path}', variables: { path: { default: 'path' } } }],
```
ilias-t, line 89: do we handle the case for when we have a variable-url but no actual variables—I only ask because I remember we was that recently (wouldn't have figured it was valid otherwise)

__tests__/tooling/index.test.js
```diff
@@ -573,70 +636,90 @@ describe('#getOperation()', () => {
...
-        },
-        2
-      ).url()
-    ).toBe('https://readme.example.com');
-  });
+      const method = source.method.toLowerCase();
+      const oas = new Oas(apiDefinition, { region: 'eu' });
+      const operation = oas.getOperation(source.url, method);
+
+      expect(operation).not.toBeUndefined();
```
ilias-t, line 676: nit: is there a `toBeDefined` ?

tooling/index.js
```diff
@@ -255,26 +269,61 @@ class Oas {
...
+          if (!found) {
+            return false;
+          }
+
+          return {
+            matchedServer: server,
+            pathName: url.split(new RegExp(rgx)).slice(-1).pop(),
+          };
+        })
+        .filter(Boolean);
```
ilias-t, line 305: nice

tooling/index.js
```diff
@@ -255,26 +269,61 @@ class Oas {
...
+
+      if (!matchedServerAndPath.length) {
         return undefined;
```
ilias-t, line 308: I know this code wasn't changed in this PR but this could just become `return`

tooling/index.js
```diff
@@ -255,26 +269,61 @@ class Oas {
...
-    // happen.
-    const targetServer = {
-      ...matchedServer,
-      url: this.replaceUrl(matchedServer.url, matchedServer.variables || {}),
-    };
+      pathName = matchedServerAndPath[0].pathName;
+      targetServer = {
+        ...matchedServerAndPath[0].matchedServer,
+      };
+    } else {
```
ilias-t, line 315: this is a pointless `else` given there's a `return` in the if above, right?

domharrington (approved):
Code seems good! Read through all of the test cases and it seems sound.

tooling/index.js
```diff
@@ -255,26 +269,61 @@ class Oas {
...
+    //
+    // So what this does is transform `https://{region}.node.example.com/v14` into
+    // `https://([-_a-zA-Z0-9[\\]]+).node.example.com/v14`, and from there we'll be able to match
+    // https://eu.node.example.com/v14/api/esm and ultimately find the operation matches for `/api/esm`.
+    if (!matchedServer) {
+      const matchedServerAndPath = servers
+        .map(server => {
+          const rgx = transformUrlIntoRegex(server.url);
+          const found = new RegExp(rgx).exec(url);
+          if (!found) {
```
ilias-t, line 296: nit this could become a ternary `return found ? { ... } : false`

ilias-t (approved):

__tests__/tooling/index.test.js
erunion: This was already being skipped, I just moved the test case into another `describe` block.

__tests__/tooling/index.test.js
erunion: So this?

```
servers: [
  { url: 'https://{region}.node.example.com/v14' }
}
```

I can add some test cases.

tooling/index.js
erunion: No, because we're setting a specific `targetServer` object and `pathName` with this regex work, otherwise we should default to the found `matchedServer` from the origin and hostname lookups.

__tests__/tooling/index.test.js
ilias-t: yes!

tooling/index.js
ilias-t: okie

__tests__/tooling/index.test.js
erunion: Looks like this PR fixes that up as well.

tooling/index.js
erunion: ESLint doesn't like when you mix `return` and `return something` in the same function. https://eslint.org/docs/rules/consistent-return

PR #432, opened 2021-05-11
feat: addition of a `splitVariables` method for extracting server variables from a URL

erunion:
## 🧰 What's being changed?

I'm working on adding support for alternate servers and server variables to [api](https://npm.im/api) and because that receives full HAR objects I don't have knowledge there of where in the HAR's URL server variables are (or which server from the OAS that I'm looking at).

With this new method, `oas.splitVariables()`, we can now pass in a fully-composed server URL like https://buster.example.com:3000/pet and receive back an object informing where in the `servers` array this server exists, as well as an object of each variable that is being used within the URL (and its data).

For example, on `oas.splitVariables('https://buster.example.com:3000/pet')` I might receive back:

```js
{ 
  selected: 0, 
  variables: { 
    name: 'buster', 
    port: 3000, 
    basePath: 'pet' 
  }
}
```

The shape of this data exactly matches what we accept in `oas.url()` so doing the following also fully works (and is also backed by new unit tests):

```js
const split = oas.splitVariables('https://buster.example.com:3000/pet');
console.log(oas.url(split.selected, split.variables)); // https://buster.example.com:3000/pet
```

## 🧪 Testing

See tests.

### DIFF
--- a/tooling/index.js
+++ b/tooling/index.js
@@ -56,7 +56,7 @@ function normalizedUrl(oas, selected) {
  * @returns {String}
  */
 function transformUrlIntoRegex(url) {
-  return stripTrailingSlash(url.replace(/{([-_a-zA-Z0-9[\]]+)}/g, () => '([-_a-zA-Z0-9[\\]]+)'));
+  return stripTrailingSlash(url.replace(/{([-_a-zA-Z0-9[\]]+)}/g, '([-_a-zA-Z0-9[\\]]+)'));
 }
 
 function normalizePath(path) {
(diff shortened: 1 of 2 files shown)


tooling/index.js
```diff
@@ -209,6 +209,49 @@ class Oas {
...
+   * `https://buster.example.com:3000/pet` to this function, you'll get back the following:
+   *
+   *    { selected: 0, variables: { name: 'buster', port: 3000, basePath: 'pet' } }
+   *
+   * Re-supplying this data to `oas.url()` should return the same URL you passed into this method.
+   *
+   * @param {String} baseUrl
+   * @returns {Object|Boolean}
+   */
+  splitVariables(baseUrl) {
```
erunion, line 226: I named this to match `splitUrl` because it's doing similar work `serverVariables` just deals with cases where we don't yet know what the variables in the server URL are. I'm open to suggestions if you have a better idea for a different name.

domharrington (approved):
LGTM!

PR #451, opened 2021-06-10
feat: use null instead of empty-string for empty state

ilias-t:
## 🧰 What's being changed?

Allows consumers of this package to more easy distinguish between an empty oas file and empty user input.

## 🧪 Testing

Updated the tests so they pass.

## ⚠️ Potentially Breaking Changes

- Whenever we integrate this into the `api-explorer` package, there are some [tests](https://github.com/readmeio/api-explorer/blob/main/packages/api-explorer/__tests__/lib/get-auth.test.js) that may need to be updated. I don't see any code in that project that seems like this it will be affected by this change.
- There are some places in the ReadMe project that explicitly check for empty strings that likely need to updated.
  - [SelectWidget](https://github.com/readmeio/readme/blob/next/packages/schema-form/src/Form/components/widgets/SelectWidget.js#L103)
  - [BaseInput](https://github.com/readmeio/readme/blob/next/packages/schema-form/src/Form/components/widgets/BaseInput.js#L72)
  - [TextAreaWidget](https://github.com/readmeio/readme/blob/next/packages/schema-form/src/Form/components/widgets/TextareaWidget.js#L7)
  - Maybe for some of the empty-string checks in [Form/utils](https://github.com/readmeio/readme/blob/next/packages/schema-form/src/Form/utils.js)
  - Maybe [MultiSchemaField](https://github.com/readmeio/readme/blob/next/packages/schema-form/src/form-components/MultiSchemaField.jsx#L36)

### DIFF
--- a/__tests__/lib/get-auth.test.js
+++ b/__tests__/lib/get-auth.test.js
@@ -8,15 +8,15 @@ const oas = new Oas(multipleSecurities);
 test('should fetch all auths from the OAS files', () => {
   expect(oas.getAuth({ oauthScheme: 'oauth', apiKeyScheme: 'apikey' })).toStrictEqual({
     apiKeyScheme: 'apikey',
-    apiKeySignature: '',
+    apiKeySignature: null,
     basicAuth: {
-      pass: '',
-      user: '',
+      pass: null,
+      user: null,
     },
-    httpBearer: '',
-    oauthDiff: '',
+    httpBearer: null,
+    oauthDiff: null,
     oauthScheme: 'oauth',
-    unknownAuthType: '',
+    unknownAuthType: null,
   });
 });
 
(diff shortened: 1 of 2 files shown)


Dashron:
are we able to lint this? I'm strongly anti-empty strings and am in favor of this change.

Dashron:
As for the possible updates that check for empty strings, I don't think select, base, text area or multischema will ever get the data you changed to null.

erunion (approved):
talked yesterday offline about this and im good with it. if you don't mind, after i publish a new release of `oas` can you update it in the `api-explorer` repo and PR that for me?

Dashron:
I would love it if I never saw a hardcoded empty string in anything ever, and overrides required comment overrides. nice to have for the future.

Dashron (approved):

erunion:
@ilias-t Oof I didn't think about that. I just went through dependency hell upgrading everything we use from oas11 to oas12.

ilias-t:
@erunion dang, sorry if this is a headache!

PR #452, opened 2021-06-23
fix: update getParameters to pass parameter name to constructSchema

darrenyong:
## 🧰 Changes

Currently, for parameters, we don't pass in `current.name` as `currentLocation` to `constructSchema`. This makes it so that when we get into `consructSchema`, `currentLocation` will be blank for the first iteration.

This change makes it so `currentLocation` will be correct for each iteration of parameters.

## 🧬 QA & Testing

This is mostly to fix user defined JWT defaults not working for path, query and header parameters. I can provide testing steps when I open the PR in the main repo.

### DIFF
--- a/src/operation/get-parameters-as-json-schema.js
+++ b/src/operation/get-parameters-as-json-schema.js
@@ -505,7 +505,7 @@ function getParameters(path, operation, oas, globalDefaults) {
           if (typeof current.content[contentType] === 'object' && 'schema' in current.content[contentType]) {
             schema = {
               ...(current.content[contentType].schema
-                ? constructSchema(current.content[contentType].schema, [], '', globalDefaults)
+                ? constructSchema(current.content[contentType].schema, [], `/${current.name}`, globalDefaults)
                 : {}),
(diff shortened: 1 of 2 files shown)


src/operation/get-parameters-as-json-schema.js
```diff
@@ -483,7 +483,7 @@ function getParameters(path, operation, oas, globalDefaults) {
       let schema = {};
       if ('schema' in current) {
         schema = {
-          ...(current.schema ? constructSchema(current.schema, [], '', globalDefaults) : {}),
+          ...(current.schema ? constructSchema(current.schema, [], `/${current.name}`, globalDefaults) : {}),
```
erunion, line 486: Might need to do the same fix on lien 508?

darrenyong: Tested it with a file to make sure that line needed to be updated too and it did! Updated!

erunion (approved):

Dashron (approved):

PR #454, opened 2021-06-28
fix: quirk where a path with a query string would break matching

erunion:
## 🧰 Changes

This fixes a quirk in our `findOperation()` method where if a path in the API definition had a query param (like `?param=123` it would completely kill the method because the `?` was being interpreted as a regex modifier for [path-to-regexp](http://npm.im/path-to-regexp). To resolve this, I've loosened up matching on these kinds of paths so it'll look on `/path` instead of `/path?param=123`.

Fixes RM-1396

## 🧬 QA & Testing

I've added test cases for the initial regression as well as some other potential variants of this quirk.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -74,7 +74,13 @@ function normalizePath(path) {
   // variable starts.
   //
   // For example if the URL is `/post/:param1::param2` we'll be escaping it to `/post/:param1\::param2`.
-  return path.replace(/{(.*?)}/g, ':$1').replace(/::/, '\\::');
+  return (
+    path
+      .replace(/{(.*?)}/g, ':$1')
+      .replace(/::/, '\\::')
+      // Need to escape question marks too because they're treated as regex modifiers in `path-to-regexp`
+      .split('?')[0]
+  );
 }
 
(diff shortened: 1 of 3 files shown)


mjcuva (approved):
Code seems good to me!

src/index.js
```diff
@@ -74,7 +74,13 @@ function normalizePath(path) {
   // variable starts.
   //
   // For example if the URL is `/post/:param1::param2` we'll be escaping it to `/post/:param1\::param2`.
-  return path.replace(/{(.*?)}/g, ':$1').replace(/::/, '\\::');
+  return (
```
domharrington, line 77: Wonder if we could utilize the URL constructor here?


```
> new URL('/rating_stats?listing_ids[]=1234567', 'https://example.com').pathname
'/rating_stats'
> new URL('/people/GWID:3', 'https://example.com').pathname
'/people/GWID:3'
```

domharrington (approved):

src/index.js
erunion: That's not a bad idea. Will look into that after I get this fix up to the preview site.

PR #455, opened 2021-06-28
fix: path matching quirk in oas files missing servers

erunion:
## 🧰 Changes

This fixes a bug with `findOperation` where if an API definition didn't have any defined `servers` we wouldn't be able to match any operation within it.

Fixes RM-1425

## 🧬 QA & Testing

See attached test.


thenaomiday (approved):
Looks solid and test passes 🎉

PR #456, opened 2021-06-30
fix: bug where variable-casing in servers and urls would break path matching

erunion:
## 🧰 Changes

This fixes a bug in `Oas.findOperation()` where if a defined API definition server url had variable casing we wouldn't be able to match it if the `uri` being passed into `findOperation` was all lowercased.

## 🧬 QA & Testing

I've added tests for this original case of the server not matching the incoming URL, but also the reverse situation.

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -698,7 +698,7 @@ describe('#findOperation()', () => {
       });
     });
 
-    it('should return a match if but the uri has variable casing', () => {
+    it("should return a match if the uri has variable casing but the defined server doesn't", () => {
       const oas = new Oas({
         servers: [{ url: 'https://api.example.com/' }],
         paths: {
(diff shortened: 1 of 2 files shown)


__tests__/index.test.js
```diff
@@ -673,6 +673,56 @@ describe('#findOperation()', () => {
...
+      expect(res.url).toStrictEqual({
+        origin: 'https://api.Example.com',
+        path: '/anything',
+        nonNormalizedPath: '/anything',
+        slugs: {},
+        method: 'GET',
+      });
+    });
+
+    it('should return a match if but the uri has variable casing', () => {
```
erunion, line 701: i can write clear sentences yes

domharrington (approved):
LGTM! Thanks Jon.

PR #470, opened 2021-07-28
feat: writing the return signature of the `get*Examples` methods to be more flexible

erunion:
Work in support of RM-1201.

## 🐳  Problems

Our existing `operation.getRequestBodyExamples()` and `operation.getResponseExamples()` both produce return signatures that are inflexible, have two internal properties where examples are stored, and aren't clear about what data you should and need to be looking for.

For example if you run `operation.getResponseBodyExamples()` on an operation that uses the `example` property for defining a single example on a single media type (`application/json`) you'll get something like this:

```json
[
  {
    "code": "{\n  \"id\": 12343354,\n  \"email\": \"test@example.com\",\n  \"name\": \"Test user name\"\n}",
    "mediaType": "application/json",
    "multipleExamples": false
  }
]
```

Run it on something that uses `example` on `text/plain` request bodies and `examples` on `application/json` and you'll get something **wildly** different:

```json
[
  {
    "code": "OK",
    "mediaType": "text/plain",
    "multipleExamples": false
  },
  {
    "code": false,
    "mediaType": "application/json",
    "multipleExamples": [
      {
        "code": "{\n  \"name\": \"Fluffy\",\n  \"petType\": \"Cat\"\n}",
        "label": "cat"
      },
      {
        "code": "{\n  \"name\": \"Puma\",\n  \"petType\": \"Dog\"\n}",
        "label": "dog"
      }
    ]
  }
]
```

For responses it's even more confusing:

```json
[
  {
    "status": "200",
    "languages": [
      {
        "code": "{\n  \"id\": 12343354,\n  \"email\": \"test@example.com\",\n  \"name\": \"Test user name\"\n}",
        "language": "*/*",
        "multipleExamples": false
      }
    ]
  }
]
```

1. What is `code` and why is it always a string? Where's my example?
2. Why do I need to look at `code` and `multipleExamples` to see if I have multiple examples present?
3. Can `code` and `multipleExamples` both be present at the same time?
4. What about example `summary` and `description` fields? Where are those?

🤷

## 👨‍🔬  Solutions

This work aims to clean up these signatures and make these APIs cleaner (a lot cleaner), and expose example summaries and descriptions if present.

The new signature for these two accessors this the following:

```json
// Request bodies
[
  {
    "mediaType": "application/json",
    "examples": [
      {
        "value": {
          "id": 12343354,
          "email": "test@example.com",
          "name": "Test user name"
        }
      }
    ]
  }
]

// Responses
[
  {
    "status": "200",
    "mediaTypes": {
      "*/*": [
        {
          "summary": "Example user",
          "value": {
            "id": 12343354,
            "email": "test@example.com",
            "name": "Test user name"
          }
        }
      ]
    }
  }
]
```

## 🧰 Changes

* [x] `languages` has been renamed to `mediaTypes` to better reflect what it actually is, and be more akin to what this data is referred to as within the OpenAPI Specification.
* [x] `code` has been fully renamed to `value`.
  * `code` was holdover from our Manual API editor where we store these manual examples as `code` in the database and the `get*Examples()` methods here had handling for both that and native OpenAPI examples. Since examples created with our Manual API editor are now being inserted into API definitions for our new Reference Guide natively within the `examples` property, we no longer need this special handling here.
    * `@readme/api-explorer` still needs this `code` handling, and will continue to do so, but thankfully because that source is compiled as a dist unless we upgrade `oas` there these changes will not affect it.
* [x] `code` is no longer always a string.
  * Reason we were always converting it to a string is because Codemirror requires that code you pass to it be stringified, so instead of doing that work where we use Codemirror it was bundled into this library. If you need to stringify an example you'll need to do so yourself now.
* [x] `code` and `multipleExamples` have both been squashed into a single `examples` array. If you need to know if you have multiple examples present, the length of that array will tell you.

## 🧬 QA & Testing

A lot of changes will be necessary to our `Request` and `Response` React components to incorporate these changes, but if tests pass here and you're good with the API changes then 👍


src/operation/get-requestbody-examples.js
```diff
@@ -1,30 +1,4 @@
...
- * This code is identical to `get-response-examples` except that this returns the media type as `mediaType` instead of
- * `language`.
- *
- * @link https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#mediaTypeObject
- * @param {string} mediaType
- * @param {object} mediaTypeObject
- * @param {(object|false)} example
- * @returns {(object|false)}
- */
-function constructMediaType(mediaType, mediaTypeObject, example) {
```
erunion, line 16: This `constructMediaType` method in the response and request body example generators was similar enough that I was able to consolidate all of this logic into a single `getMediaTypeExamples` library that can handle both as they're both now just OAS [Media Type Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#mediaTypeObject).

src/lib/get-mediatype-examples.js
```diff
@@ -1,92 +1,77 @@
...
+    return [
+      {
+        value: mediaTypeObject.example,
+      },
+    ];
+  } else if (mediaTypeObject.examples) {
+    const { examples } = mediaTypeObject;
+    const multipleExamples = Object.keys(examples)
+      .map(key => {
+        let summary = key;
```
kanadgupta, line 26: I feel like there's value in preserving both the `key` here **and** a hypothetical `example.summary`. What about storing this as a `title` value or something? So like in this example:
```yml
foo:
  summary: A foo example
  value: {"foo": "bar"}
bar:
  summary: A bar example
  value: {"bar": "baz"}
```
we'd set the `multipleExamples` array to something like:
```json
[
  {
    "summary": "A foo example",
    "title": "foo",
    "value": { "foo": "bar" }
  },
  {
    "summary": "A bar example",
    "title": "bar",
    "value": { "bar": "baz" }
  }
]
```
 Based on [the spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#example-object), it seems like the `summary` contains descriptive information that seems like it's playing a different role than the key of the object.

kanadgupta (commented):
I think this is overall a great change (and I appreciate you writing such a detailed rationale about these changes 💙)! I like how well this better conforms to the spec rather than trying to reinvent the wheel and I always appreciate when we can consolidate logic! I'm assuming these changes are considered breaking?

One question about our `summary` logic below, but otherwise LGTM!

src/lib/get-mediatype-examples.js
erunion: I like this.

erunion: Added in https://github.com/readmeio/oas/pull/470/commits/2327feec10347b82a964b15ff4de31bbb5e18a8d

kanadgupta (approved):

domharrington (approved):
Code changes LGTM! The old object was kinda weird for sure.

PR #475, opened 2021-08-03
feat: improved handling of same-level properties and oneOf's

erunion:
## 🧰 Changes

This updates our `getParametersAsJsonSchema` library to improve some of its handling with `properties` and polymorphic `oneOf` schemas that are at the same level. That's a lot of goobledegook, let me explain.

See the following schema:

Screen Shot 2021-08-03 at 1 39 18 PM

In this there's a `properties` object and a `oneOf` array at the same level within the `exclusiveRequirements` object. Currently when we process this schema the `properties` object is left alone because it's fully formed but we add a `type: string` into each of the options within the `oneOf` because we think that they're malformed schemas. The result of this when we render it within ReadMe is this:

Screen Shot 2021-08-03 at 1 41 39 PM

Instead what this schema is actually trying to tell us is that `exclusiveRequirements` can be either the `properties` object with `requiredOne` as required or the same but with `requiredTwo` required.

Because attempting to fix this kind of thing within our JSON Schema form work is an **incredibly** daunting task it's much easier for us to guide its hand a bit by reframing what this schema looks like into something that'll be handled as it's intended to be rendered. Automatically.

Screen Shot 2021-08-02 at 4 52 54 PM

With this work if there's a `properties` or `oneOf`/`anyOf`/`allOf` at the same level within an object, that polymorphic part of the equation will be transposed into a nested `allOf` array including the `properties` schema. So in this case the `{ title: '...', required: [...] }` object will now look like:

```
{
  "allOf": [
    {
      "title": "...",
      "required": [...]
    },
    {
      "properties": {
        "requiredOne": { ... },
        "requiredTwo": { ... },
        "optional": { ... }
      }
    }
  ]
}
```

Now when our JSON Schema form handles this it'll run it through [json-schema-merge-allof](https://www.npmjs.com/package/json-schema-merge-allof) and generate something that looks like this:

Screen Shot 2021-08-03 at 1 46 56 PM

Fix in support of RM-1499.

## 🧬 QA & Testing

See attached unit tests.

> ⚠️  Note that the screenshots above are a bit misleading and show "Option 1" where it should say "One data member is required.", as well as the duplicate descriptions, are bugs within our form system and will be addressed within that codebase.

### DIFF
--- a/src/operation/get-parameters-as-json-schema.js
+++ b/src/operation/get-parameters-as-json-schema.js
@@ -53,6 +53,10 @@ function isPolymorphicSchema(schema) {
   return 'allOf' in schema || 'anyOf' in schema || 'oneOf' in schema;
 }
 
+function cloneObject(obj) {
+  return JSON.parse(JSON.stringify(obj));
+}
+
 /**
  * Determine if a given schema looks like a `requestBody` schema and contains the `content` object.
  *
@@ -212,21 +216,13 @@ function constructSchema(data, opts = {}) {
           isPolymorphicAllOfChild: polyType === 'allOf',
         };
 
-        // When `properties` are present alongside a polymorphic schema instead of letting whatever JSON Schema
-        // interpreter is handling these constructed schemas we can guide its hand a bit by manually transforming it
-        // into an inferred `allOf` of the `properties` + the polymorph schema.
+        // When `properties` or `items` are present alongside a polymorphic schema instead of letting whatever JSON
+        // Schema interpreter is handling these constructed schemas we can guide its hand a bit by manually transforming
+        // it into an inferred `allOf` of the `properties` + the polymorph schema.
         if ('properties' in schema) {
-          schema[polyType][idx] = constructSchema(
-            {
-              allOf: [
-                item,
-                {
-                  properties: schema.properties,
-                },
-              ],
-            },
-            polyOptions
-          );
+          schema[polyType][idx] = constructSchema({ allOf: [item, { properties: schema.properties }] }, polyOptions);
+        } else if ('items' in schema) {
+          schema[polyType][idx] = constructSchema({ allOf: [item, { items: schema.items }] }, polyOptions);
         } else {
           schema[polyType][idx] = constructSchema(item, polyOptions);
         }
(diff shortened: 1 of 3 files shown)


src/operation/get-parameters-as-json-schema.js
```diff
@@ -177,19 +177,21 @@ function searchForExampleByPointer(pointer, examples = []) {
...
+ * @param {Boolean} opts.isPolymorphicAllOfChild - Is this schema the child of a polymorphic `allOf` schema?
  */
-function constructSchema(
-  data,
-  prevSchemas = [],
-  currentLocation = '',
-  globalDefaults,
-  isPolymorphicAllOfChild = false
-) {
+function constructSchema(data, opts = {}) {
```
erunion, line 186: Reworked this argument structure into an `opts` object because the arguments list was getting a little out of control with half of them being optional.

src/operation/get-parameters-as-json-schema.js
```diff
@@ -198,6 +200,40 @@ function constructSchema(
...
+          prevSchemas,
+          currentLocation: `${currentLocation}/${idx}`,
+          globalDefaults,
+          isPolymorphicAllOfChild: polyType === 'allOf',
+        };
+
+        // When `properties` are present alongside a polymorphic schema instead of letting whatever JSON Schema
+        // interpreter is handling these constructed schemas we can guide its hand a bit by manually transforming it
+        // into an inferred `allOf` of the `properties` + the polymorph schema.
+        if ('properties' in schema) {
```
Dashron, line 218: should we do this with `items` too?

Dashron (approved):

julshotal (approved):
Tests lgtm 👍

src/operation/get-parameters-as-json-schema.js
erunion: yeah

src/operation/get-parameters-as-json-schema.js
```diff
@@ -415,7 +444,10 @@ function getRequestBody(operation, oas, globalDefaults) {
     examples.push({ examples: requestBody.examples });
   }
 
-  const cleanedSchema = constructSchema(requestBody.schema, examples, '', globalDefaults);
+  // We're cloning the request, and further below the component, schema because we've had issues with request schemas
+  // that were dereferenced being processed multiple times because their component is also processed.
+  const requestSchema = cloneObject(requestBody.schema);
```
erunion, line 449: @Dashron @julshotal I discovered a bug while adding support for `items` to this work, and changing the example OAS I wrote to use `$ref` pointers, where schemas were sometimes being processed multiple times depending if they were used in a `$ref`. I've added these `cloneObject()` methods around in a few places where we're initially calling `constructSchema` to fix it.

PR #476, opened 2021-08-04
fix: automatic merging of dynamically generated `allOf` schemas

erunion:
## 🧰 Changes

This is a slight modification to #475 where instead of migrating `properties` and/or `items` into a newly created `allOf` we're now going to dynamically merge those objects if possible.

Doing this because there's some quirks with our form generator where if you give it a `oneOf` where each option contains an `allOf` it can't pick up the `title` from inside of the `allOf` and honestly it's just easier to fix this by doing it in here instead of there.

## 🧬 QA & Testing

See updated tests.

### DIFF
--- a/__tests__/operation/get-parameters-as-json-schema.test.js
+++ b/__tests__/operation/get-parameters-as-json-schema.test.js
@@ -513,12 +513,8 @@ describe('parameters', () => {
             polymorphicParam: {
               type: 'object',
               oneOf: [
-                {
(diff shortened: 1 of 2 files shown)


Dashron (approved):
Bummer we couldn't keep the allof trick, it feels far safer

erunion:
Going to fix this instead in the form code as I like the `allOf` work.

PR #478, opened 2021-08-09
fix: bug where request or response example fetching would update the operation schema

erunion:
## 🧰 Changes

This fixes a bug in our core `sampleFromSchema` handling for request and response example generation where if you ran one before the other, schemas that are either `readOnly` or `writeOnly` would be excluded when they should normally be present. This was happening because the object we were passing into `sampleFromSchema` was being passed as a reference and that code updates the schema object, which was then propagating back upstream to the `Operation` class instance.

I'm also adding support for `format: YYYY-MM-DD` because I noticed a spec we have has that instead of the more common `format: date`.

Fix in support of RM-1765.

## 🧬 QA & Testing

See added tests.

### DIFF
--- a/__tests__/__datasets__/readonly-writeonly.json
+++ b/__tests__/__datasets__/readonly-writeonly.json
@@ -44,6 +44,63 @@
             }
           }
         }
+      },
+      "post": {
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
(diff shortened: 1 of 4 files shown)


darrenyong (approved):
Awesome, thanks for fixing this so quickly. Code looks good, will check it in the API Explorer in the readme repo! :smile:

PR #479, opened 2021-08-11
perf: only include components if circular $ref pointers are in the JSON Schema

erunion:
## 🧰 Changes

This makes a few minor, but important, changes to `Operation.getParametersAsJsonSchema()` with regard to how it handles `$ref` pointers and some problems it's got therein:

* If a `parameter` schema utilized a `$ref` pointer, and that pointer remained after dereferencing (implying that it's circular), we would previously not copy that component into the resulting JSON Schema leading to a JSON Schema to at has a `$ref` pointer to something that doesn't exist.
* If a `requestBody` exists, and there are `components` in the API definition we would always copy those components into the JSON Schema for that request body. Always.

With this work we now only include `components` if `$ref` pointers are discovered while constructing JSON Schema for parameters, request bodies, and responses.

What's this mean?

* If a circular `$ref` exists anywhere in a parameter or request body it will now still be accessible (irregardless if we actually want to process that circular pointer or not).
* Since we're now only including `components` if `$ref` pointers are in the JSON Schema we're constructing for +95% of the JSON Schema we construct will now be **dramatically** smaller.

Using the [petstore](https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore.json) example on the old code generating JSON Schema for `POST /pet` would produce a JSON object of 4627 characters. Now? 

618 characters. **An 87% reduction**. 

giphy

Work in support of CX-499.

## 🧬 QA & Testing

See new unit tests I added and also the smaller snapshots from data no longer being unnecessarily dumped into the schema.

### DIFF
--- a/__tests__/lib/openapi-to-json-schema.test.js
+++ b/__tests__/lib/openapi-to-json-schema.test.js
@@ -738,6 +738,7 @@ describe('`example` / `examples` support', () => {
 
       const schema = operation.getParametersAsJsonSchema()[0].schema;
 
+      expect(schema.components).toBeUndefined();
       expect(schema.properties.id.examples).toStrictEqual([20]);
 
       // Not `buster` because `doggie` is set directly alongside `name` in the definition.
@@ -749,11 +750,6 @@ describe('`example` / `examples` support', () => {
           examples: ['https://example.com/dog.png'],
         },
       });
-
-      // `Pet` schema `id` example should not be present because that `id` was set against the `requestBody`, not the
-      // component.
-      expect(schema.components.schemas.Pet.properties.id.examples).toBeUndefined();
-      expect(schema.components.schemas.Pet.properties.name.examples).toStrictEqual(['doggie']);
     });
   });
 
--- a/src/lib/openapi-to-json-schema.js
+++ b/src/lib/openapi-to-json-schema.js
@@ -266,6 +266,7 @@ function toJSONSchema(data, opts = {}) {
         const example = schema.examples[name];
         if ('$ref' in example) {
           // no-op because any `$ref` example here after dereferencing is circular so we should ignore it
+          refLogger(example.$ref);
         } else if ('value' in example) {
           if (isPrimitive(example.value)) {
             examples.push(example.value);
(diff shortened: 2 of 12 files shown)


src/lib/openapi-to-json-schema.js
```diff
@@ -168,23 +168,27 @@ function searchForExampleByPointer(pointer, examples = []) {
  * @link https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md
  * @param {Object} data
  * @param {Object} opts
- * @param {Object[]} opts.prevSchemas - Array of parent schemas to utilize when attempting to path together examples.
  * @param {String} opts.currentLocation - Current location within the schema -- this is a JSON pointer.
  * @param {Object} opts.globalDefaults - Object containing a global set of defaults that we should apply to schemas that match it.
  * @param {Boolean} opts.isPolymorphicAllOfChild - Is this schema the child of a polymorphic `allOf` schema?
+ * @param {Object[]} opts.prevSchemas - Array of parent schemas to utilize when attempting to path together examples.
+ * @param {Function} opts.refLogger - A function that's called anytime a (circular) `$ref` is found.
```
gratcliff, line 175: Is `opts.refLogger` supposed to be a configurable hook?

gratcliff (approved):

src/lib/openapi-to-json-schema.js
erunion: Yeah we supply that option within in `get-parameters-as-json-schema` so we can optionally add `components` into each schema group that needs it.

__tests__/lib/openapi-to-json-schema.test.js
```diff
@@ -738,6 +738,7 @@ describe('`example` / `examples` support', () => {
 
       const schema = operation.getParametersAsJsonSchema()[0].schema;
 
+      expect(schema.components).toBeUndefined();
```
erunion, line 741: Components is/should be undefined here because the OAS it's testing has been dereferenced and there aren't any circular `$ref`s in it.

src/lib/openapi-to-json-schema.js
gratcliff: Do you think this needs to be documented further?

erunion: I think it's ok

PR #481, opened 2021-08-16
feat: preserving ref schema names during dereferencing

erunion:
## 🧰 Changes

With the work I did in #479, I accidentaly broke some functionality in our form generator where for an `anyOf` or `oneOf` schema we use the name of the component schema if there isn't a `title` present. Since those schemas are no longer always present we need an alternative way to expose what the schema *was* called, and we're doing that now with this `x-readme-ref-name` extension.

Pre-dereferencing we now run through all component schemas and note what their name is so that when we dereference that metadata will be preserved wherever that schema `$ref` is utilized.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -495,6 +495,15 @@ class Oas {
     // Extract non-OAS properties that are on the class so we can supply only the OAS to the ref parser.
     const { _dereferencing, _promises, user, ...oas } = this;
 
+    // Because referencing will eliminate any lineage back to the original `$ref`, information that we might need at
+    // some point, we should run through all available component schemas and denote what their name is so that when
+    // dereferencing happens below those names will be preserved.
+    if (oas && oas.components && oas.components.schemas && typeof oas.components.schemas === 'object') {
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -495,6 +495,15 @@ class Oas {
     // Extract non-OAS properties that are on the class so we can supply only the OAS to the ref parser.
     const { _dereferencing, _promises, user, ...oas } = this;
 
+    // Because referencing will eliminate any lineage back to the original `$ref`, information that we might need at
+    // some point, we should run through all available component schemas and denote what their name is so that when
+    // dereferencing happens below those names will be preserved.
+    if (oas && oas.components && oas.components.schemas && typeof oas.components.schemas === 'object') {
+      Object.keys(oas.components.schemas).forEach(schemaName => {
+        oas.components.schemas[schemaName]['x-readme-ref-name'] = schemaName;
```
Dashron, line 503: clean and simple, love it.

Dashron (approved):

PR #482, opened 2021-08-20
fix: quirk with request/response examples where readonly/writeonly was ignored in allOf

erunion:
## 🧰 Changes

Fixes a quirk with `readOnly` and `writeOnly` where if they were within a schema that was in an `allOf` we wouldn't include them in generated examples.

Fixes RM-1765

## 🧬 QA & Testing

See tests.


Dashron (approved):
tests make sense, code looks fine. I'll test more thoroughly in the next pr on site.

PR #499, opened 2021-09-14
feat: adding a new `getTags` method to the operation class

erunion:
## 🧰 Changes

Adds a new `.getTags()` method to the `Operation` class that will pull back tags, and their OAS-level defined metadata, for the current operation.

## 🧬 QA & Testing

See tests.


julshotal (approved):
lgtm! Thank you!

PR #501, opened 2021-10-01
feat: exposting deprecated parameters

julshotal:
## 🧰 Changes
WIP
Exposing the deprecated property on schemas for use in the reference section, and removing it from our unsupported property list. I have been using this via npm link to test adding the deprecated badge to the schema forms on the other side

## 🧬 QA & Testing
- [ ] Is this a good way to expose this property or should it be done more similarly to required? 
- [ ] Did I miss anything important?
- [ ] Tests (that I need to write) are passing


erunion (approved):
added some more tests cases but looks good!

PR #532, opened 2021-11-01
feat: migrating more of the library to typescript

erunion:
## 🧰 Changes

In addition to implementing [eslint-plugin-jsdoc](https://npm.im/eslint-plugin-jsdoc) to enforce better code documentation this migrates the following pieces of code over to Typescript:

* `lib/`
  * [x] `get-user-variable`
  * [x] `get-schema`
  * [x] `get-media-type-examples`
  * [x] `get-user-variable`
  * [x] `matches-mimetype`
* `operation/`
  * [x] `get-callback-examples`
  * [x] `get-requestbody-examples`
  * [x] `get-response-examples`
* [x] `samples/*`
* [x] `Oas`
* [x] `Operation`
  * Pulled across from https://github.com/readmeio/oas/pull/530

All that's left to get migrated over to TS is being done in https://github.com/readmeio/oas/pull/530 and should not require a breaking change release (as this work here is due to some changes in how `Oas` stores its data, see #536).

## 🧬 QA & Testing

If all tests are still passing then 👍


__tests__/index.test.ts
```diff
@@ -21,76 +20,79 @@ test('should export utils', () => {
...
 describe('#getVersion()', () => {
-  it('should be able to identify a Swagger definition', () => {
-    expect(new Oas(swagger).getVersion()).toBe('2.0');
-  });
-
   it('should be able to identify an OpenAPI definition', () => {
-    expect(new Oas(petstore).getVersion()).toBe('3.0.0');
-    expect(new Oas(webhooks).getVersion()).toBe('3.1.0');
+    expect(new Oas(petstore as RMOAS.OASDocument).getVersion()).toBe('3.0.0');
+    expect(new Oas(webhooks as RMOAS.OASDocument).getVersion()).toBe('3.1.0');
```
erunion, line 30: I loathe having to type these specs but there's some weird quirk with `openapi-types` and the petstore that we've got (that's a valid spec):

```
Argument of type '{ openapi: string; servers: { url: string; }[]; info: { description: string; version: string; title: string; termsOfService: string; contact: { email: string; }; license: { name: string; url: string; }; }; externalDocs: { ...; }; tags: ({ ...; } | { ...; })[]; paths: { ...; }; components: { ...; }; }' is not assignable to parameter of type 'OASDocument'.
  Type '{ openapi: string; servers: { url: string; }[]; info: { description: string; version: string; title: string; termsOfService: string; contact: { email: string; }; license: { name: string; url: string; }; }; externalDocs: { ...; }; tags: ({ ...; } | { ...; })[]; paths: { ...; }; components: { ...; }; }' is not assignable to type 'Omit<Omit<Document<{}>, "paths" | "components">, "paths" | "components" | "info" | "servers" | "webhooks" | "jsonSchemaDialect"> & { info: InfoObject; jsonSchemaDialect?: string; servers?: ServerObject[]; } & Pick<...> & Omit<...>'.
    Type '{ openapi: string; servers: { url: string; }[]; info: { description: string; version: string; title: string; termsOfService: string; contact: { email: string; }; license: { name: string; url: string; }; }; externalDocs: { ...; }; tags: ({ ...; } | { ...; })[]; paths: { ...; }; components: { ...; }; }' is not assignable to type 'Pick<PathsWebhooksComponents<{}>, "components">'.
      Types of property 'components' are incompatible.
        Type '{ schemas: { Order: { type: string; properties: { id: { type: string; format: string; }; petId: { type: string; format: string; }; quantity: { type: string; format: string; }; shipDate: { type: string; format: string; }; status: { ...; }; complete: { ...; }; }; xml: { ...; }; }; ... 4 more ...; ApiResponse: { ...; }...' is not assignable to type 'ComponentsObject'.
          Type '{ schemas: { Order: { type: string; properties: { id: { type: string; format: string; }; petId: { type: string; format: string; }; quantity: { type: string; format: string; }; shipDate: { type: string; format: string; }; status: { ...; }; complete: { ...; }; }; xml: { ...; }; }; ... 4 more ...; ApiResponse: { ...; }...' is not assignable to type '{ schemas?: Record<string, SchemaObject>; responses?: Record<string, ReferenceObject | ResponseObject>; parameters?: Record<...>; ... 6 more ...; pathItems?: Record<...>; }'.
            Types of property 'schemas' are incompatible.
              Type '{ Order: { type: string; properties: { id: { type: string; format: string; }; petId: { type: string; format: string; }; quantity: { type: string; format: string; }; shipDate: { type: string; format: string; }; status: { ...; }; complete: { ...; }; }; xml: { ...; }; }; ... 4 more ...; ApiResponse: { ...; }; }' is not assignable to type 'Record<string, SchemaObject>'.
                Property '"Order"' is incompatible with index signature.
                  Type '{ type: string; properties: { id: { type: string; format: string; }; petId: { type: string; format: string; }; quantity: { type: string; format: string; }; shipDate: { type: string; format: string; }; status: { type: string; description: string; enum: string[]; }; complete: { ...; }; }; xml: { ...; }; }' is not assignable to type 'SchemaObject'.
                    Type '{ type: string; properties: { id: { type: string; format: string; }; petId: { type: string; format: string; }; quantity: { type: string; format: string; }; shipDate: { type: string; format: string; }; status: { type: string; description: string; enum: string[]; }; complete: { ...; }; }; xml: { ...; }; }' is not assignable to type 'MixedSchemaObject'.
                      Types of property 'type' are incompatible.
                        Type 'string' is not assignable to type '("array" | NonArraySchemaObjectType)[]'.ts(2345)
```

__tests__/index.test.ts
```diff
@@ -122,7 +125,13 @@ describe('#url([selected])', () => {
...
       expect(
         new Oas({
+          openapi: '3.0.0',
+          info: {
+            title: 'testing',
+            version: '1.0.0',
+          },
           servers: [{ url: 'https://example.com/{path}', variables: { path: { default: 'path' } } }],
+          paths: {},
         }).url()
```
erunion, line 135: Fleshed these test specs out so they're valid and we don't need to force type them as `RMOAS.OASDocument`.

__tests__/index.test.ts
```diff
@@ -152,34 +173,25 @@ describe('#url([selected])', () => {
...
           servers: [
             { url: 'https://{username1}.example.com', variables: { username1: { default: 'demo1' } } },
             { url: 'https://{username2}.example.com', variables: { username2: { default: 'demo2' } } },
           ],
+          paths: {},
         }).url(1)
       ).toBe('https://demo2.example.com');
     });
 
-    it.skip('should fetch user variables from selected app', () => {
```
erunion, line 163: Removed this test because it's been skipped for 2+ years now.

__tests__/index.test.ts
```diff
@@ -335,12 +364,15 @@ describe('#splitVariables()', () => {
       },
     });
 
+    // @ts-expect-error This is an annoying exclusion but `split` is a known object at this point.
```
erunion, line 367: Would love to know if there's a better way to handle this.

__tests__/index.test.ts
```diff
@@ -990,12 +1055,11 @@ describe('#getOperation()', () => {
     it('should be able to find an operation if there are no user variables present', () => {
       const source = {
         url: 'https://eu.node.example.com/v14/api/esm',
-        method: 'put',
+        method: 'put' as RMOAS.HttpMethods,
```
erunion, line 1058: Kind of annoying.

__tests__/index.test.ts
```diff
@@ -1151,20 +1214,22 @@ describe('#dereference()', () => {
...
 
     expect(
-      oas.api.paths['/multischema/of-everything'].post.requestBody.content['application/json'].schema[
-        'x-readme-ref-name'
-      ]
+      (
+        (oas.api.paths['/multischema/of-everything'].post.requestBody as RMOAS.RequestBodyObject).content[
+          'application/json'
+        ].schema as RMOAS.SchemaObject
+      )['x-readme-ref-name']
```
erunion, line 1225: Kind of annoying.

__tests__/samples/index.test.ts
```diff
@@ -202,93 +189,81 @@ describe('sampleFromSchema', () => {
...
-    expect(sampleFromSchema(definition, { includeWriteOnly: true })).toStrictEqual(expected);
+    });
   });
 
   describe('primitive type handling', () => {
     it('should handle when an unknown type is detected', () => {
-      const definition = {
+      const definition: SchemaObject = {
         type: 'array',
+        // @ts-expect-error We're testing the failure case for `png` not being a valid type.
```
erunion, line 205: Hate having to do this but it's in a test and we're testing a failure case.

package.json
```diff
@@ -49,6 +49,8 @@
   },
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.6",
+    "@types/json-schema-merge-allof": "^0.6.1",
+    "@types/memoizee": "^0.4.6",
```
erunion, line 53: Should these actually be dev deps?

src/rmoas.types.ts
```diff
@@ -0,0 +1,67 @@
+import { OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
+// import Oas from '.';
+
+export type OASDocument = OpenAPIV3.Document | OpenAPIV3_1.Document;
+export type OperationObject = OpenAPIV3.OperationObject | OpenAPIV3_1.OperationObject;
+
+export type HttpMethods =
+  | (OpenAPIV3.HttpMethods | OpenAPIV3_1.HttpMethods)
+  | ('get' | 'put' | 'post' | 'delete' | 'options' | 'head' | 'patch' | 'trace');
```
erunion, line 9: Had to add these so we could use them as valid methods because the enum from the `*.HttpMethods` types wasn't doing it.

src/samples/index.ts
```diff
@@ -22,23 +22,26 @@ const primitives = {
...
+  boolean: (schema: SchemaObject): boolean => (typeof schema.default === 'boolean' ? schema.default : true),
 };
 
-const primitive = schema => {
+const primitive = (schema: SchemaObject) => {
   schema = objectify(schema);
   const { type, format } = schema;
 
+  // @ts-expect-error Suppressing `any` errors here because if what we're looking for isn't in `primitives` we throw.
   const fn = primitives[`${type}_${format}`] || primitives[type];
```
erunion, line 33: The error I'm suppressing here:

> Element implicitly has an 'any' type because expression of type '`${string}_${string}`' can't be used to index type '{ string: () => string; string_email: () => string; 'string_date-time': () => string; string_date: () => string; 'string_YYYY-MM-DD': () => string; string_uuid: () => string; string_hostname: () => string; string_ipv4: () => string; ... 4 more ...; boolean: (schema: SchemaObject) => boolean; }'.ts(7053)

Is there a better way to handle this?

src/types.ts
```diff
@@ -0,0 +1,15 @@
+import { OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
```
erunion, line 1: // todo merge these with `rmoas.types`

tsconfig.json
```diff
@@ -4,7 +4,6 @@
     "baseUrl": "./src/",
     "declaration": true,
     "declarationDir": "./@types",
-    "downlevelIteration": true, // @todo remove this option when the entire library has been ported to TS
```
erunion, line 7: This was in here because we were converting a regex iterator to an array in the `Oas` class using the spread operator but I was able to remove this setting by rewriting that to use `Array.from()` instead.

tsconfig.test.json
```diff
@@ -2,7 +2,8 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "module": "es6",
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "noImplicitAny": false,
```
erunion, line 6: Added this because it was really annoying fighting with it in unit tests.

src/operation.ts
```diff
@@ -254,27 +305,24 @@ class Operation {
...
 
-    let oasTags = new Map();
+    const oasTags = new Map();
     if ('tags' in this.api) {
       this.api.tags.forEach(tag => {
         oasTags.set(tag.name, tag);
       });
     }
 
-    oasTags = Object.fromEntries(oasTags);
```
erunion, line 271: Removed this nonsense to just use `Map.has()` and `Map.get()` instead like a non-galaxy brained person.

<sub>what the heck was i thinking</sub>

src/index.ts
```diff
@@ -1,12 +1,31 @@
...
+
+import $RefParser from '@apidevtools/json-schema-ref-parser';
+import { pathToRegexp, match } from 'path-to-regexp';
+import getAuth from './lib/get-auth';
+import getUserVariable, { UserVariable } from './lib/get-user-variable';
+import Operation, { Callback, Webhook } from './operation';
+import utils from './utils';
+
+// @todo replace `pathToRegexp_MatchResult` with the `MatchResult` type that `path-to-regexp` has.
+type pathToRegexp_MatchResult = { index: number; params: Record<string, string>; path: string };
```
erunion, line 12: This is really annoying but the `MatchResult` type that `path-to-regexp` exports was fighting me in the `generatePathMatches` method with this:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
  No index signature with a parameter of type 'string' was found on type '{}'.ts(7053)
```

src/index.ts
```diff
@@ -149,9 +170,24 @@ function findTargetPath(pathMatches) {
...
+export default class Oas {
+  api: RMOAS.OASDocument;
+
+  user: User;
+
+  // @todo These are always `Promise.resolve` and `Promise.reject`. Make these types more specific than `any`.
+  _promises: Array<{
+    resolve: any;
+    reject: any;
+  }>;
```
erunion, line 182: Annoying but I couldn't get the `Promise<any>` to work here instead.

src/index.ts
```diff
@@ -267,8 +307,8 @@ class Oas {
         // instead of doing that we need to re-regex the server URL, this time splitting on the path parameters -- this
         // way we'll be able to extract the parameter names and match them up with the matched server that we obtained
         // above.
-        const variables = {};
-        [...server.url.matchAll(/{([-_a-zA-Z0-9[\]]+)}/g)].forEach((variable, y) => {
```
erunion, line 271: Converted this to use `Array.from()` instead because Typescript was requiring a down-level transpiling in order to get its iterator working with the spread operator. Doesn't need that with `Array.from()`. 🤷

src/index.ts
```diff
@@ -149,9 +170,24 @@ function findTargetPath(pathMatches) {
...
+
+  user: RMOAS.User;
+
+  // @todo These are always `Promise.resolve` and `Promise.reject`. Make these types more specific than `any`.
+  _promises: Array<{
+    resolve: any;
+    reject: any;
+  }>;
+
+  _dereferencing: {
```
erunion, line 184: Would love to make these `private` properties but we have some tests that's looking at this data and I don't want to add a method to expose it just for that test.

src/index.ts
```diff
@@ -149,9 +170,24 @@ function findTargetPath(pathMatches) {
...
+export default class Oas {
+  api: RMOAS.OASDocument;
+
+  user: RMOAS.User;
+
+  // @todo These are always `Promise.resolve` and `Promise.reject`. Make these types more specific than `any`.
+  _promises: Array<{
+    resolve: any;
+    reject: any;
+  }>;
```
erunion, line 182: Annoying but I couldn't get the `Promise<any>` to work here instead.

src/index.ts
erunion, line 184: Would love to make this and `_promises` private properties but we have some tests looking at this data to make sure it's working, and I don't want to add a getter just for that test.

src/lib/get-user-variable.ts
```diff
@@ -0,0 +1,21 @@
+import * as RMOAS from '../rmoas.types';
+
+export type UserVariable = unknown | null;
+
+function getKey(key: RMOAS.User | Record<string, primitive>, property: string): UserVariable {
```
ilias-t, line 5: nit: generally I'm against return types as all it really does is create redundant code that at best does nothing but at worst widens the type definition specificity

ilias-t: obviously there are exceptions to this rule like for callback functions or something where the function ends up being defined elsewhere and it then needs to be consumed in some place and that consumer needs to guard against that return type

erunion: Aren't return types the strongest part of TS?

ilias-t: How so? The params are usually the most useful aspect when defining types. Try removing all the return types in this PR and you'll see that TS will be able to derive them all from the code itself rather than some override that may or may not actually widen the specificity.

ilias-t: one way to think about explicit vs implicit types as just with regular variables. there's no need to write `isThing: boolean = true` or `key: string = 'foo'`, TS understands what types they are implicitly and it's the same concept here.

erunion: Huh yeah ok. I didn't realize that TS was smart enough to infer returned types without having them be explicit. I'll refactor this.

erunion: @ilias-t Thanks for the feedback, I've gone through and removed most of the types that I could that TS wasn't able to automatically infer.

package.json
ilias-t: yeah I think that makes more sense

src/samples/index.ts
ilias-t: hard to say without playing with the code directly, here's a related [stack post](https://stackoverflow.com/questions/57086672/element-implicitly-has-an-any-type-because-expression-of-type-string-cant-b). I would just make sure `primitives` is typed properly and if that doesn't work you can always `as` it

ilias-t (approved):

src/samples/index.ts
erunion: Was able to fix this by typing `primitives` to the following:

```
Record<string, (arg: void | RMOAS.SchemaObject) => primitive | boolean>
```

src/samples/index.ts
```diff
@@ -22,23 +22,35 @@ const primitives = {
...
 };
 
-const primitive = schema => {
+const primitive = (schema: RMOAS.SchemaObject) => {
   schema = objectify(schema);
   const { type, format } = schema;
 
-  const fn = primitives[`${type}_${format}`] || primitives[type];
-
+  const fn = primitives[`${type}_${format}`] || primitives[type as string];
```
Dashron, line 32: I'm surprised you have to type `type` here? shouldn't it know from the schema?

src/operation/get-response-examples.ts
```diff
@@ -0,0 +1,51 @@
...
+        if (!mediaType) return;
+
+        const mediaTypeObject = response.content[mediaType];
+        const examples = getMediaTypeExamples(mediaType, mediaTypeObject, {
+          includeReadOnly: true,
+          includeWriteOnly: false,
+        });
+
+        if (examples) {
+          mediaTypes[mediaType] = examples as MediaTypeExample[];
```
Dashron, line 37: does `getMediaTypeExamples` not already return a `MediaTypeExample[]`?

Dashron (changes requested):
phew, that was a long one. some possible typing improvements, but they might not all be accurate. looks good otherwise, if tests pass I trust ya on execution.

tsconfig.test.json
erunion: That's what this is doing because it's in `tsconfig.test.json` unless you're referring to something else.

__tests__/lib/get-auth.test.ts
```diff
@@ -35,15 +38,15 @@ test('should not error if oas.components is not set', () => {
   const user = { oauthScheme: 'oauth', apiKeyScheme: 'apikey' };
 
   expect(() => {
-    new Oas().getAuth(user);
+    new Oas({} as unknown as RMOAS.OASDocument).getAuth(user);
```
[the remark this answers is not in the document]
erunion: well... this code wasn't working yesterday without `as unknown` there but it's working now so i'm removing those.

🤷‍♂️

src/global.d.ts
```diff
@@ -0,0 +1 @@
+type primitive = string | number;
```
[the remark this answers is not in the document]
erunion: yeah

src/cli/lib/utils.js
```diff
@@ -101,7 +101,7 @@ exports.fileExists = function (file) {
 /**
  * Really simple way at guessing the language that their API might be written in. If we're wrong, it's not a big deal!
  *
- * @returns {String}
+ * @returns string
```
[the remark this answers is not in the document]
erunion: Made this because of `eslint-plugin-jsdoc` requirements but I later added exclusions for `src/cli` into our config. Reverting this.

src/index.ts
erunion: Was able to fix this with this:

Screen Shot 2021-11-04 at 4 46 00 PM

src/index.ts
```diff
@@ -149,29 +209,58 @@ function findTargetPath(pathMatches) {
...
+   * Internal storage array that the library utilizes to keep track of its `dereferencing` state so it doesn't initiate
+   * multiple dereferencing processes.
+   */
+  protected dereferencing: {
+    processing: boolean;
+    complete: boolean;
+  };
+
+  constructor(oas: RMOAS.OASDocument, user?: RMOAS.User) {
+    this.api = oas || ({} as RMOAS.OASDocument);
```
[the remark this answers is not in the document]
erunion: Yeah this throws because things that are required for a valid OAS aren't present in `{}`:

```
Type 'OASDocument | {}' is not assignable to type 'OASDocument'.
  Type '{}' is not assignable to type 'OASDocument'.
    Type '{}' is not assignable to type 'Omit<Omit<Document<{}>, "paths" | "components">, "servers" | "paths" | "components" | "info" | "webhooks" | "jsonSchemaDialect"> & { info: InfoObject; jsonSchemaDialect?: string; servers?: ServerObject[]; } & Pick<...> & Omit<...>'.
      Property 'openapi' is missing in type '{}' but required in type 'Omit<Omit<Document<{}>, "paths" | "components">, "servers" | "paths" | "components" | "info" | "webhooks" | "jsonSchemaDialect">'.ts(2322)
index.d.ts(153, 9): 'openapi' is declared here.
```

Typecasting it directly to `as RMOAS.OASDocument` fixes that. I don't really want to define `this.api` as `Partial<RMOAS.OASDocument>` because that's going to make everything in the OAS that's required as not.

src/index.ts
```diff
@@ -297,26 +386,28 @@ class Oas {
...
-  replaceUrl(url, variables = {}) {
+  replaceUrl(url: string, variables: Variables = {}) {
     // When we're constructing URLs, server URLs with trailing slashes cause problems with doing lookups, so if we have
     // one here on, slice it off.
     return stripTrailingSlash(
-      url.replace(/{([-_a-zA-Z0-9[\]]+)}/g, (original, key) => {
-        if (getUserVariable(this.user, key)) {
-          return getUserVariable(this.user, key);
+      url.replace(/{([-_a-zA-Z0-9[\]]+)}/g, (original: string, key: string) => {
+        const userVariable = getUserVariable(this.user, key);
```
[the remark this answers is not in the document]
erunion: `this.user` is coming from our JWT so all we really know about it is that it's an object. https://docs.readme.com/docs/passing-data-to-jwt

src/index.ts
```diff
@@ -432,29 +532,32 @@ class Oas {
...
+  findOperation(url: string, method: RMOAS.HttpMethods): PathMatch {
     const annotatedPaths = this.findOperationMatches(url);
     if (!annotatedPaths) {
       return undefined;
     }
 
-    const includesMethod = filterPathMethods(annotatedPaths, method);
-    if (!includesMethod.length) return undefined;
-    return findTargetPath(includesMethod);
+    const matches = filterPathMethods(annotatedPaths, method) as Array<{
```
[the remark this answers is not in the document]
erunion: This is happening because of some funky stuff with `PathMatch.p` sometimes being a `PathsObject` and other times being an `OperationObject`. Since all the tests here still work I'm going to follow up this PR with a better fix for this.

src/index.ts
```diff
@@ -531,14 +635,15 @@ class Oas {
...
   getWebhooks() {
-    const webhooks = {};
+    const webhooks: Record<string, Record<RMOAS.HttpMethods, Webhook>> = {};
+    const api = this.api as OpenAPIV3_1.Document;
 
-    Object.keys(this.api.webhooks ? this.api.webhooks : []).forEach(id => {
-      webhooks[id] = {};
-      Object.keys(this.api.webhooks[id]).forEach(method => {
+    Object.keys(api.webhooks ? api.webhooks : []).forEach(id => {
+      webhooks[id] = {} as Record<RMOAS.HttpMethods, Webhook>;
```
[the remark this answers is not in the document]
erunion: This is to initialize `webhooks[id]` so we can do `webhooks[id][method] = this.operation(...)` on the line below it.

src/lib/get-auth.ts
```diff
@@ -44,11 +29,24 @@ function getKey(user: User, scheme: SecurityScheme): authKey {
...
+ * database that were ingested before we had good validation in place.
+ *
+ * @param user User
+ * @param scheme Security scheme to get auth keys for.
+ * @param selectedApp The user app to retrieve an auth key for.
+ * @returns The found auth key for this security scheme.
+ */
+export function getByScheme(
+  user: RMOAS.User,
+  scheme = <RMOAS.KeyedSecuritySchemeObject | any>{},
```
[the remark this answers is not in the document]
erunion: Added that so that these tests could still function where we're testing a case where an unknown schema type is present:

```
expect(getByScheme(topLevelUser, { type: 'unknown' })).toBeNull();
expect(getByScheme(keysUser, { type: 'unknown' })).toBeNull();
expect(getByScheme(keysUser, { type: 'unknown' }, 'app-2')).toBeNull();
```

Without `any` there:

> Type '"unknown"' is not assignable to type '"oauth2" | "http" | "apiKey" | "openIdConnect"'.ts(2322)
> index.d.ts(415, 9): The expected type comes from property 'type' which is declared here on type 'KeyedSecuritySchemeObject'

There a better way to handle this without `any`, or are these tests just not really necessary anymore?

src/lib/get-schema.ts
```diff
@@ -0,0 +1,56 @@
...
+  api?:
+    | RMOAS.OASDocument
+    | {
+        [schemaType: string]: {
+          [key: string]: RMOAS.SchemaObject;
+        };
+      }
+): {
+  type: string;
+  schema: any; // @fixme give this a better type
```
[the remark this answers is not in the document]
erunion: I don't remember actually but I don't like this entire file and want to rewrite how it's used in `get-parameters-as-json-schema` to use a generalized `Operation.getRequestBody()` method. Going to leave this as-is and revisit in a later PR.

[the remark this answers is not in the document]
erunion: I think it's because `schema` can sometimes be a `RequestBodyObject` and other times it can be a `SchemaObject`? It's a goddamn mess.

src/lib/get-user-variable.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  if ('keys' in user && Array.isArray(user.keys) && user.keys.length) {
+    if (selectedApp) {
+      key = user.keys.find(k => k.name === selectedApp);
+    } else {
+      key = user.keys[0];
+    }
+  }
+
+  return key[property] || null;
```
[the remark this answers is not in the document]
erunion: Not sure I follow.

PR #537, opened 2021-11-05
feat: addition of an `init` and methods for pulling spec extensions

erunion:
## 🧰 Changes

* [x] Creates a static `Oas.init()` method that will do type transformation on raw JSON objects before it's sent into the `Oas` constructor. This helps so that we don't need to export `RMOAS.OASDocument` and you can do `new Oas(petstore)` out of the box.
* [x] Adds `hasExtension()` and `getExtension()` methods to both `Oas` and `Operation` for [spec extensions](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#specificationExtensions) lookups. We can use this in [@readme/oas-extensions](https://npm.im/@readme/oas-extensions) to clean up some work there.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -1411,3 +1411,23 @@ describe('#getTags()', () => {
     });
   });
 });
+
+describe('#getExtension()', () => {
+  it('should return the extension if it exists', () => {
+    const oas = Oas.init({
+      ...petstore,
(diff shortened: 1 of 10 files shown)


domharrington (approved):
Don't fully understand why it's necessary but code LGTM!

PR #540, opened 2021-11-16
fix: upgrading eslint to eslint 8

erunion:
## 🧰 Changes

* [x] Upgrades ESLint to v8
* [x] Upgrades `@readme/eslint-config` to v8
* [x] Fixes a handful of new issues that popped up.


src/operation.ts
```diff
@@ -566,7 +566,3 @@ export class Callback extends Operation {
 }
 
 export class Webhook extends Operation {}
-
-module.exports = Operation;
-module.exports.Callback = Callback;
-module.exports.Webhook = Webhook;
```
erunion, line 572: Not sure what happened here but some changes to the [import/no-import-module-exports](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-import-module-exports.md) rule caught it.

src/operation/get-callback-examples.ts
```diff
@@ -18,7 +18,6 @@ export default function getCallbackExamples(operation: RMOAS.OperationObject) {
   const ret: CallbackExamples = [];
 
   // spreads the contents of the map for each callback so there's not nested arrays returned
-  // eslint-disable-next-line sonarjs/no-empty-collection
```
erunion, line 21: I've removed `eslint-plugin-sonarjs` from our config because it doesn't support ESLint 8 yet. TBD if it'll come back because it hasn't really offered us much.

Dashron (approved):

PR #563, opened 2021-12-23
fix: improved support for common params

erunion:
## 🧰 Changes

* [x] Updates our `getCallbacks()` accessor to stop including non-method items (like common parameters) from being returned as a callback.
* [x] Updates our `getSummary()` and `getDescription()` accessors to allow commonly set summary and descriptions to override those at the path item level.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/__tests__/operation.test.ts
+++ b/__tests__/operation.test.ts
@@ -5,6 +5,7 @@ import callbackSchema from './__datasets__/callbacks.json';
 import multipleSecurities from './__datasets__/multiple-securities.json';
 import referenceSpec from './__datasets__/local-link.json';
 import deprecatedSchema from './__datasets__/schema-deprecated.json';
+import parametersCommon from './__datasets__/parameters-common.json';
 
 describe('#constructor', () => {
   const oas = Oas.init(petstore);
@@ -30,6 +31,10 @@ describe('#getSummary()', () => {
   it('should return nothing if not present', () => {
     expect(Oas.init(referenceSpec).operation('/2.0/users/{username}', 'get').getSummary()).toBeUndefined();
   });
+
+  it('should allow a common summary to override the operation-level summary', () => {
+    expect(Oas.init(parametersCommon).operation('/anything/{id}', 'get').getSummary()).toBe('[common] Summary');
+  });
 });
 
 describe('#getDescription()', () => {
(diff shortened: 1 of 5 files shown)


__tests__/operation.test.ts
```diff
@@ -30,6 +31,10 @@ describe('#getSummary()', () => {
   it('should return nothing if not present', () => {
     expect(Oas.init(referenceSpec).operation('/2.0/users/{username}', 'get').getSummary()).toBeUndefined();
   });
+
+  it('should allow a common summary to override the operation-level summary', () => {
+    expect(Oas.init(parametersCommon).operation('/anything/{id}', 'get').getSummary()).toBe('[common] Summary');
```
kanadgupta, line 36: This honestly feels a lil counterintuitive to me... I would think that since the operation-level summary is more specific, it should override the common summary 🤔 but if that's the expected behavior within the OAS, fine by me 🤷🏽

__tests__/operation.test.ts
```diff
@@ -966,6 +975,31 @@ describe('#getCallbacks()', () => {
     const operation = Oas.init(petstore).operation('/pet', 'put');
     expect(operation.getCallbacks()).toBe(false);
   });
+
+  it("should return an empty object for the operation if only only callbacks present aren't supported HTTP methods", () => {
```
kanadgupta, line 979:
```suggestion
  it("should return an empty object for the operation if only callbacks present aren't supported HTTP methods", () => {
```

__tests__/operation.test.ts
```diff
@@ -966,6 +975,31 @@ describe('#getCallbacks()', () => {
...
+      paths: {
+        '/anything': {
+          post: {
+            callbacks: {
+              batchSuccess: {
+                '{$url}': {
+                  summary: 'Batch call webhook',
+                },
+              },
+            },
```
kanadgupta, line 995: I'm a bit confused by this test, where's the unsupported HTTP method here?

kanadgupta (approved):
A couple of small comments but otherwise LGTM!

__tests__/operation.test.ts
erunion: Agreed. It makes sense for `parameters` since if you have a path param you don't really want to dupe that on every path item, but for summary or description? ehhh

__tests__/operation.test.ts
erunion: There is none. Instead of having something like `post: <path item>` there's `summary` instead. Since `summary` isn't a valid HTTP method we're checking that running `getCallbacks()` on this operation returns an empty array.

__tests__/operation.test.ts
erunion: how embarassing

__tests__/operation.test.ts
kanadgupta: Ohhhh got it. I see `summary` all the time so nothing looked out of place at first glance. Could you add a lil' comment explaining that?

__tests__/operation.test.ts
kanadgupta: you're fine, it's dec 23rd lol

julshotal (approved):

PR #575, opened 2022-01-05
fix: callbacks not having path-level data to retrieve common summaries

erunion:
## 🧰 Changes

This fixes a regression I introduced in #563 where if you execute either `getSummary()` or `getDescription()` on a `Callback` instance they would fail because the callback doesn't have access to the root `path` data for `this.api.paths`.

To fix this I've overridden the two methods in the `Callback` class to utilize a new `parent` data that the `Callback` constructor now requires so all callbacks have access to their immediate parent [path item object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#pathItemObject) where common summary and description information is stored.

This was a doozy to brain out.

## 🧬 QA & Testing

I've merged the `getSummary()` and `getDescription()` tests into a single block because they were both loading the same operations for their tests -- should hopefully make those tests a little cleaner.

### DIFF
--- a/__tests__/__datasets__/callbacks.json
+++ b/__tests__/__datasets__/callbacks.json
@@ -24,6 +24,8 @@
           "myCallback": {
             "{$request.query.queryUrl}": {
               "post": {
+                "summary": "Callback summary",
+                "description": "Callback description",
                 "requestBody": {
                   "description": "Callback payload",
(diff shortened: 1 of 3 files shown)


julshotal (approved):
lgtm!

PR #577, opened 2022-01-07
chore: make our jsdoc eslint rules a bit less obnoxious

erunion:
## 🧰 Changes

The ESLint rules that we put in place for JSDoc blocks has been a bit obnoxious with TypeScript so I'm slightly tweaking it to: 

* [x] No longer require docblocks be present on functions in TS code. But if docblocks are present they'll be validated.
* [x] Removing the requirement for a `@returns` tag. All of our existing `@returns` tags have essentially summarized what's already been said in the docblock summary so they've been a bit unnecessary and just add clutter.
* [x] Removing the requirement for TS docblocks to have parameters typed as this duplicates the types that we've got with the actual code and just worsens the documentation.


.eslintrc
```diff
@@ -31,12 +30,20 @@
   },
   "overrides": [
     {
-      "files": ["__tests__/__fixtures__/*.js", "src/cli/*.js", "src/cli/commands/*.js", "src/cli/lib/*.js"],
+      "files": ["*.js"],
```
Dashron, line 33: will this lead to node_modules being linted? Is there a way to do a verbose run and output all the files we're checking to make sure nothing obvious wrong is now included?

src/index.ts
```diff
@@ -338,13 +325,7 @@ export default class Oas {
...
         }
 
-        // I wanted to do this here but due to us not
-        // babelifying node_modules and not committing ./.tooling
-        // to git, I'm just gunna do this for now so I can
-        // get on with my life!
-        //
-        // const variable = variables?.[value]
-        const variable = (variables[value] || {}) as RMOAS.ServerVariableObject;
+        const variable = variables?.[value];
```
Dashron, line 328: 🎉  one less `as`!

Dashron (approved):

.eslintrc
erunion: ESLint ignores `node_modules/` by default, but running it with `--debug` it doesn't look in there still.

PR #579, opened 2022-01-11
fix: refactoring allOf schemas to be pre-merged when we generate JSON Schema

erunion:
## 🧰 Changes

Last week we received an API definition that contained an `allOf` that contained another allOf with two things:

* An object with a `data` property that evaluated to a `{}`.
* An object with a `data` property that evaluated to a complete object.

Because we have handling in place to not repair `allOf` children if they are missing a `type` declaration, this case was falling through that and the empty `data` object was getting a `type: string` property added to it because we thought that it was malformed. Further up the stack when we render this schema out we later attempt to merge this newly "fixed" `allOf` and because `type: string` and `type: object` are incompatible we either wouldn't render out anything, or in some cases would fully crash.

Less than ideal!

Since we don't have any great way of determining if a sub-object property is contained within an `allOf` and isn't malformed and will be okay when it's merged with another schema we decided to completely forgo this repair work that we had to repair schemas that were missing a `type` (and weren't either implicitly an object or an array), and instead merge **all** `allOf` schemas when we generate JSON Schema.

Going forward, with this work any tooling that ingests JSON Schema generated with `getParametersAsJsonSchema()` will no longer need to worry about `allOf` cases.

Fix in support of RM-3146.

## 🧬 QA & Testing

There's a lot going on here, but all tests should no longer have any `allOf` expectations.

As for the case that uncovered this issue lives, it in `__tests__/__datasets__/polymorphism-quirks.json` and is tested within `__tests__/operation/get-parameters-as-json-schema.test.js`.


__tests__/__datasets__/schema-deprecated.json
```diff
@@ -222,11 +222,9 @@
                 "deprecated": true,
                 "allOf": [
                   {
-                    "title": "option 1",
```
erunion, line 225: Removed these because they were always going to be merged into a single `title` property.

src/lib/openapi-to-json-schema.ts
```diff
@@ -403,12 +437,17 @@ export default function toJSONSchema(
...
+            Array.isArray(schema.properties[prop]) ||
+            (typeof schema.properties[prop] === 'object' && schema.properties[prop] !== null)
+          ) {
+            schema.properties[prop] = toJSONSchema(schema.properties[prop] as RMOAS.SchemaObject, {
+              currentLocation: `${currentLocation}/${encodePointer(prop)}`,
+              globalDefaults,
+              prevSchemas,
+              refLogger,
+            });
+          }
```
erunion, line 450: Made this change because I discovered that if there was an `x-` extension, and it was a primitive, in an object we'd run it through the generator again and completely the value.

screen_shot_2022-01-10_at_3 44 13_pm

Though thinking about it, maybe this should instead just exclude anything that starts with `x-`?

src/lib/openapi-to-json-schema.ts
```diff
@@ -247,20 +252,56 @@ export default function toJSONSchema(
...
+            } as unknown,
+          }) as RMOAS.SchemaObject;
+        } catch (e) {
+          // If we can't merge the `allOf` for whatever reason (like if one item is a `string` and the other is a
+          // `object`) then we should completely remove it from the schema and continue with whatever we've got. Why?
+          // If we don't, any tooling that's ingesting this will need to account for the incompatible `allOf` and it may
+          // be subject to more breakages than just not having it present would be.
+          const { ...schemaWithoutAllOf } = schema;
+          schema = schemaWithoutAllOf as RMOAS.SchemaObject;
+          delete schema.allOf;
```
erunion, line 284: I'm still kind of iffy on this but this is how our schema form currently handles this case, but I still need to follow up this PR with changes to our schema form to handle if `type` isn't present and treat them as a string.

src/lib/openapi-to-json-schema.ts
```diff
@@ -282,8 +323,9 @@ export default function toJSONSchema(
       if ('mapping' in schema.discriminator && typeof schema.discriminator.mapping === 'object') {
         // Discriminator mappings aren't written as traditional `$ref` pointers so in order to log them to the supplied
         // `refLogger`.
-        Object.keys(schema.discriminator.mapping).forEach(k => {
-          refLogger(schema.discriminator.mapping[k]);
+        const mapping = schema.discriminator.mapping;
+        Object.keys(mapping).forEach(k => {
+          refLogger(mapping[k]);
```
erunion, line 328: Had to make this change because TS didn't like how we were accessing `schema.discriminator` here and didn't think that `discriminator` was present on the `schema` type (despite the same code working two lines above).

__tests__/lib/openapi-to-json-schema.test.ts
```diff
@@ -227,7 +214,12 @@ describe('polymorphism / inheritance', () => {
                   },
                 },
                 {
-                  type: 'integer',
```
erunion, line 230: Changed this because `integer` and `object` schemas can't be merged.

__tests__/lib/openapi-to-json-schema.test.ts
```diff
@@ -455,33 +490,35 @@ describe('`format` support', () => {
...
       ['number', 'double', 0 - Number.MAX_VALUE, Number.MAX_VALUE],
-    ])('`%s`', (type, format, min, max) => {
-      it('should add a `minimum` and `maximum` if not present', () => {
-        expect(toJSONSchema({ type: type as JSONSchema7TypeName, format })).toStrictEqual({
-          type,
-          format,
-          minimum: min,
-          maximum: max,
+    ])('`type: %s`', (type, format, min, max) => {
+      describe(`\`format: ${format}\``, () => {
```
erunion, line 494: Changed this because we were duping all of these integer + format tests under the same `describe` block and they weren't properly exposing themselves in the test output. Thankfully they were still being run, it just wasn't clear what was being run for what.

Dashron (approved):
just one small comment, otherwise stuff looks good. will test more on the readme merge.

src/lib/openapi-to-json-schema.ts
```diff
@@ -247,20 +252,56 @@ export default function toJSONSchema(
   // If we don't have a set type, but are dealing with an `anyOf`, `oneOf`, or `allOf` representation let's run through
   // them and make sure they're good.
   if (RMOAS.isSchema(schema, isPolymorphicAllOfChild)) {
-    ['allOf', 'anyOf', 'oneOf'].forEach((polyType: 'allOf' | 'anyOf' | 'oneOf') => {
+    // If this is an `allOf` schema we should make an attempt to merge so as to ease the burden on the tooling that
+    // ingests these schemas.
+    if ('allOf' in schema && Array.isArray(schema.allOf)) {
+      if ('allOf' in schema) {
```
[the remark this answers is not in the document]
erunion: Hm yeah that is weird and unnecessary. Will fix.

PR #580, opened 2022-01-12
feat: addition of a getRequestBody accessor

erunion:
## 🧰 Changes

Adds a new `getRequestBody` accessor on the `Operation` object that pulls back a request body [Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#mediaTypeObject) for any specific media type that's present on the operation.

This isn't going to be used anywhere just yet but when we refactor/rewrite the `getSchema()` library here it can be used there.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/__tests__/operation.test.ts
+++ b/__tests__/operation.test.ts
@@ -933,6 +933,56 @@ describe('#hasRequestBody()', () => {
   });
 });
 
+describe('#getRequestBody()', () => {
+  it('should return false on an operation without a requestBody', async () => {
+    const oas = Oas.init(petstore);
+    await oas.dereference();
+
+    const operation = oas.operation('/pet/findByStatus', 'get');
+    expect(operation.getRequestBody('application/json')).toBe(false);
+  });
+
+  it('should return false on an operation without the specified requestBody media type', async () => {
+    const oas = Oas.init(petstore);
+    await oas.dereference();
+
+    const operation = oas.operation('/pet', 'put');
+    expect(operation.getRequestBody('text/xml')).toBe(false);
+  });
+
+  it('should return the specified requestBody media type', async () => {
+    const oas = Oas.init(petstore);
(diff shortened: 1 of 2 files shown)


__tests__/operation.test.ts
```diff
@@ -933,34 +933,83 @@ describe('#hasRequestBody()', () => {
...
+
 describe('#getResponseByStatusCode()', () => {
   it('should return false if the status code doesnt exist', () => {
     const operation = Oas.init(petstore).operation('/pet/findByStatus', 'get');
     expect(operation.getResponseByStatusCode(202)).toBe(false);
   });
 
-  it('should return the response', () => {
-    const operation = Oas.init(petstore).operation('/pet/findByStatus', 'get');
+  it('should return the response', async () => {
```
erunion, line 992: I had to rework this test a bit because the dereferencing library we use does parameter reassigning so if one test dereferences the petstore spec then tests that use that spec after it will have a dereferenced spec.

src/operation.ts
```diff
@@ -419,13 +419,33 @@ export default class Operation {
...
+   * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#mediaTypeObject}
+   * @param mediaType Specific request body media type to retrieve if present.
+   */
+  getRequestBody(mediaType: string) {
+    if (!this.hasRequestBody()) {
+      return false;
+    }
+
+    const requestBody = this.schema.requestBody;
+    if (RMOAS.isRef(requestBody) || !(mediaType in requestBody.content)) {
```
Dashron, line 442: Why return false if it's a ref? It's probably worth leaving those details as a comment here. I also don't see any tests for when the req body is a ref.

Dashron (approved):

PR #582, opened 2022-01-20
refactor: use default for samples over generic type

julshotal:
## 🧰 Changes

When there's no example present, right now we just autofill the example sample with the type. This will use the default if it exists before falling back to a generic type-based sample 

I used how we were taking the default for booleans over a generic value and turned it into a reusable function to either use the default if it 1) exists and 2) is the right type, otherwise it falls back onto a provided generic value 

The type check probably isn't super necessary, but on the off chance someone tries to use a wacky default hopefully it'll prevent something from breaking
## 🧬 QA & Testing
see tests

### DIFF
--- a/__tests__/samples/index.test.ts
+++ b/__tests__/samples/index.test.ts
@@ -54,13 +54,14 @@ describe('sampleFromSchema', () => {
         readOnlyDog: {
           readOnly: true,
           type: 'string',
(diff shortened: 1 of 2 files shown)


erunion (approved):
lgtm! merge whenever and lmk so I can tag a release

PR #585, opened 2022-01-24
fix: funky typing with how we extract examples during JSON Schema generation

erunion:
## 🧰 Changes

This doesn't fix any bugs but rather a type issue discovered in #581 where when we store examples in the `openapi-to-json-schema` `prevSchemas` object we were storing them as the OpenAPI Schema Object, which is an object of objects, when it was actually expecting an array of objects.

It's difficult to properly explain but check out https://github.com/readmeio/oas/pull/581#discussion_r785599642 for the TS error that uncovered this.

### DIFF
--- a/__tests__/lib/openapi-to-json-schema.test.ts
+++ b/__tests__/lib/openapi-to-json-schema.test.ts
@@ -703,12 +703,12 @@ describe('`example` / `examples` support', () => {
       }
 
       return {
-        distinctName: {
-          value,
-        },
         // Since this doesn't have a `value` that we can use we should ignore it.
-        distinctName2: {
+        distinctExternal: {
           externalValue: 'https://example.com/example.example',
+        },
+        distinctExample: {
+          value,
         },
       };
     }
--- a/__tests__/lib/openapi-to-json-schema.test.ts
+++ b/__tests__/lib/openapi-to-json-schema.test.ts
@@ -706,6 +706,10 @@ describe('`example` / `examples` support', () => {
         distinctName: {
           value,
         },
+        // Since this doesn't have a `value` that we can use we should ignore it.
+        distinctName2: {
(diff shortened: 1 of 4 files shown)


__tests__/samples/index.test.ts
```diff
@@ -310,12 +310,8 @@ describe('sampleFromSchema', () => {
...
         };
 
-        // 0-20 chops off milliseconds
-        // necessary because test latency can cause failures
-        // it would be better to mock Date globally and expect a string - KS 11/18
-        const expected = new Date().toISOString().substring(0, 20);
-
-        expect(sampleFromSchema(definition)).toContain(expected);
+        // 2022-01-24T21:26:50.058Z
+        expect(sampleFromSchema(definition)).toMatch(/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/);
```
erunion, line 314: This test has been flaky today in CI so instead of trying to mess with mocking `new Date().toISOString()` this is easier and less prone to more flakiness from mocking the Date global.

__tests__/lib/openapi-to-json-schema.test.ts
```diff
@@ -703,7 +703,11 @@ describe('`example` / `examples` support', () => {
       }
 
       return {
-        distinctName: {
+        // Since this doesn't have a `value` that we can use we should ignore it.
+        distinctExternal: {
+          externalValue: 'https://example.com/example.example',
+        },
```
erunion, line 709: I didn't change any assertions or create any new explicit tests here because by putting this above `distinctExample` since it doesn't have a `value` all of the tests within this `describe` block are now guaranteed to fallback to using `distinctExample`.

Dashron (approved):

PR #599, opened 2022-02-05
test: colocating the testing tsconfig with the tests

erunion:
VSCode doesn't support `testconfig.*.json` configs but colocating our testing tsconfig with tests seems to make it work.


Dashron (approved):
yeah this is how I like to maintain separate tsconfigs.

PR #601, opened 2022-02-15
feat: adding a dereference option to preserve ref pointers as titles

erunion:
## 🧰 Changes

In working on codegen in [api](https://npm.im/api) I want to generate Typescript types with the original `$ref` names but the library I'm using [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) doesn't offer any flexibility other than using the `title` present in the JSON Schema; and since we're preserving `$ref` pointers during the dereference process to an `x-readme-ref-name` extension I can't use this data.

This adds a new option to the `dereference` method to allow me to override any present `title` in schemas with the original `$ref` name. It may result in some data loss but in the end I think that with this codegen the types will be cleaner with the original ref names than whatever `title` was originally present within the schema.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/__tests__/__snapshots__/index.test.ts.snap
+++ b/__tests__/__snapshots__/index.test.ts.snap
@@ -1,6 +1,2079 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`#dereference() should add metadata to components pre-dereferencing to preserve their lineage 1`] = `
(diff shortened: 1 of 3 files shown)


Dashron (approved):
talked with jon about some test additions (ensure title is undefined before overwriting in existing test, and add a test for when a title exists and is being overwritten), otherwise looks good.

erunion:
Not going to add more tests to this right now because the dereference process, and `json-schema-ref-parser`, updates the spec you dereference by reference, so even though I haven't dereferenced a petstore in one test yet because it was dereferenced in another that petstore import cache has been updated.

Annoying but I'm not in the mood to completely overhaul the test suite right now.

PR #602, opened 2022-02-17
feat: optionally generate friendlier operationIds

erunion:
## 🧰 Changes

We're currently generating URL slug-like operation IDs when one isn't present, and we use this work in `api` and for operations that don't have an operation ID it makes the UX of calling one of these operations really gnarly:

```js
api['get-pets']().then(res => console.log('json response:', res))
```

With this work we'll be able to generate friendlier ones:

```js
api.getPets().then(res => console.log('json response:', res))
```

## 🧬 QA & Testing

See tests.

### DIFF
--- a/src/operation.ts
+++ b/src/operation.ts
@@ -323,12 +323,13 @@ export default class Operation {
   }
 
   /**
-   * Get an `operationId` for this operation. If one is not present (it's not required by the spec!) a hash of the path
-   * and method will be returned instead.
+   * Get an `operationId` for this operation. If one is not present (it's not required by the spec!)
+   * a hash of the path and method will be returned instead.
    *
-   * @param shouldCamelCase Generate a JS method-friendly operation ID when an `operationId` isn't present.
+   * @param opts
+   * @param opts.camelCase Generate a JS method-friendly operation ID when one isn't present.
    */
-  getOperationId(shouldCamelCase = false): string {
+  getOperationId(opts?: { camelCase: boolean }): string {
     if ('operationId' in this.schema) {
       return this.schema.operationId;
     }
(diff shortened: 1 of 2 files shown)


julshotal (approved):

src/operation.ts
```diff
@@ -326,19 +326,34 @@ export default class Operation {
...
-    const url = this.path
+    const method = this.method.toLowerCase();
+    let operationId = this.path
       .replace(/[^a-zA-Z0-9]/g, '-') // Remove weird characters
       .replace(/^-|-$/g, '') // Don't start or end with -
       .replace(/--+/g, '-') // Remove double --'s
       .toLowerCase();
 
-    return `${this.method.toLowerCase()}_${url}`;
+    if (shouldCamelCase) {
```
Dashron, line 343: if I'm reading this right, we're keeping this backwards compatible? Are you going to enable camel case for anything but the api lib?

I ask because we probably don't want to break explorer links.

src/operation.ts
```diff
@@ -326,19 +326,34 @@ export default class Operation {
...
+      operationId = operationId.replace(/[^a-zA-Z0-9]+(.)/g, (_, chr) => chr.toUpperCase());
+
+      // If the generated operationId already starts with the method (eg. `getPets`) we don't want
+      // to double it up into `getGetPets`.
+      if (operationId.startsWith(method)) {
```
Dashron, line 348: nice little optimization

Dashron (approved):

src/operation.ts
erunion: CamelCase work is only for the API library.

src/operation.ts
```diff
@@ -326,19 +326,34 @@ export default class Operation {
    * Get an `operationId` for this operation. If one is not present (it's not required by the spec!) a hash of the path
    * and method will be returned instead.
    *
+   * @param shouldCamelCase Generate a JS method-friendly operation ID when an `operationId` isn't present.
    */
-  getOperationId(): string {
+  getOperationId(shouldCamelCase = false): string {
```
domharrington, line 331: Any alternative API design here to avoid the [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)?

Maybe `getOperationId({ camelCase: true })` or something?

PR #617, opened 2022-03-05
feat: adding new options to `getParametersAsJsonSchema`

erunion:
## 🧰 Changes

This adds a couple new options to `getParametersAsJsonSchema` (and reworks the method signature into accepting an options object).

Because I am changing the method signature of `getParametersAsJsonSchema` this work will incur a breaking change release.
 
### mergeIntoBodyAndMetadata
So for some background, in [api](https://npm.im/api) method signatures consist of two arguments: `body` and `metadata`. `body` is comprised of request body payloads, and `metadata` is comprised of everything else (query params, headers, path, and cookies).

Within the SDK code generation project I'm working on in RM-3414 I need to insert auto-generated Typescript parameter interfaces into methods:

```ts
interface FindPetsByStatusPathParam {
  /**
   * Pet id to delete
   */
  petId?: number;
  [k: string]: unknown;
}

interface FindPetsByStatusQueryParam {
  api_key?: string;
  [k: string]: unknown;
};
```

The problem however is that for the auto-generated method signatures, I don't want to have type unions everywhere:

```ts
findPetsByStatus(body: findPetsByStatusBody, metadata: FindPetsByStatusPathParam & FindPetsByStatusQueryParam) {
  return this.core.fetch('/pet/findByStatus', 'get', body, metadata);
}
```

With this new option I can generate a single JSON Schema entry and using an `allOf` automatically have [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) feed me a pre-unioned type:

```ts
type FindPetsByStatusMetadata = {
  /**
   * Pet id to delete
   */
  petId?: number;
  [k: string]: unknown;
} & {
  api_key?: string;
  [k: string]: unknown;
};
```

Using that, codegen methods get a lot cleaner:

```ts
findPetsByStatus(body: findPetsByStatusBody, metadata: FindPetsByStatusMetadata) {
  return this.core.fetch('/pet/findByStatus', 'get', body, metadata);
}
```

### retainDeprecatedProperties

This new method is to augment our existing handling of splitting `deprecated` properties out of their original schemas and into a `deprecatedProps` object. When this option is enabled this work will not happen and properties and schemas marked as `deprecated` will stay where they are.

I plan on utilizing this option again in [api](https://npm.im/api) codegen so that if a property is deprecated it gets a `@deprecated` JSDoc tag.

## 🧬 QA & Testing

There's a lot going on in the diff because I overhauled how we're instantiating our example `Oas` instances with a `beforeAll()` block to do dereferencing so you might want to view the diff with whitespace changes excluded.

Beyond that, check the couple of new tests I added for these options.

### DIFF
--- a/__tests__/operation/get-parameters-as-json-schema.test.ts
+++ b/__tests__/operation/get-parameters-as-json-schema.test.ts
@@ -1,14 +1,42 @@
-const Oas = require('../../src').default;
+import type { OperationObject, RequestBodyObject, SchemaObject } from '../../src/rmoas.types';
+import Oas from '../../src';
 
-const createOas = require('../__fixtures__/create-oas').default;
-const circular = require('../__datasets__/circular.json');
-const discriminators = require('../__datasets__/discriminators.json');
-const parametersCommon = require('../__datasets__/parameters-common.json');
-const petstore = require('@readme/oas-examples/3.0/json/petstore.json');
-const petstore_31 = require('@readme/oas-examples/3.1/json/petstore.json');
-const petstoreServerVars = require('../__datasets__/petstore-server-vars.json');
-const deprecated = require('../__datasets__/schema-deprecated.json');
-const polymorphismQuirks = require('../__datasets__/polymorphism-quirks.json');
+import createOas from '../__fixtures__/create-oas';
+
+let circular: Oas;
+let discriminators: Oas;
+let parametersCommon: Oas;
+let petstore: Oas;
+let petstore_31: Oas;
+let petstoreServerVars: Oas;
+let deprecated: Oas;
+let polymorphismQuirks: Oas;
+
+beforeAll(async () => {
+  circular = await import('../__datasets__/circular.json').then(Oas.init);
+  await circular.dereference();
+
+  discriminators = await import('../__datasets__/discriminators.json').then(Oas.init);
+  await discriminators.dereference();
+
+  parametersCommon = await import('../__datasets__/parameters-common.json').then(Oas.init);
+  await parametersCommon.dereference();
+
+  petstore = await import('@readme/oas-examples/3.0/json/petstore.json').then(Oas.init);
+  await petstore.dereference();
+
+  petstore_31 = await import('@readme/oas-examples/3.1/json/petstore.json').then(Oas.init);
+  await petstore_31.dereference();
+
+  petstoreServerVars = await import('../__datasets__/petstore-server-vars.json').then(Oas.init);
+  await petstoreServerVars.dereference();
+
+  deprecated = await import('../__datasets__/schema-deprecated.json').then(Oas.init);
+  await deprecated.dereference();
+
+  polymorphismQuirks = await import('../__datasets__/polymorphism-quirks.json').then(Oas.init);
+  await polymorphismQuirks.dereference();
+});
 
 test('it should return with null if there are no parameters', () => {
   expect(createOas({ parameters: [] }).operation('/', 'get').getParametersAsJsonSchema()).toBeNull();
@@ -16,7 +44,7 @@ test('it should return with null if there are no parameters', () => {
 });
 
 describe('type sorting', () => {
-  const operation = {
+  const operation: OperationObject = {
     parameters: [
       { in: 'path', name: 'path parameter', schema: { type: 'string' } },
       { in: 'query', name: 'query parameter', schema: { type: 'string' } },
@@ -30,7 +58,7 @@ describe('type sorting', () => {
   };
 
   it('should return with a json schema for each parameter type (formData instead of body)', () => {
-    operation.requestBody.content = {
+    (operation.requestBody as RequestBodyObject).content = {
       'application/x-www-form-urlencoded': {
         schema: {
           type: 'object',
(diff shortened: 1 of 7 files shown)


__tests__/operation/get-parameters-as-json-schema.test.ts
```diff
@@ -430,35 +434,6 @@ describe('required', () => {
   it.todo('should make things required correctly for request bodies');
 });
 
-describe('`globalDefaults` option', () => {
```
erunion, line 433: Moved this to the bottom of the file and put it in a new `describe` block for all of the options that are now available.

__tests__/operation/get-parameters-as-json-schema.test.ts
```diff
@@ -750,3 +718,91 @@ describe('deprecated', () => {
...
+      const jwtDefaults = {
+        petId: 1,
+      };
+
+      const jsonSchema = operation.getParametersAsJsonSchema({ globalDefaults: jwtDefaults });
+      expect((jsonSchema[0].schema.properties.petId as SchemaObject).default).toStrictEqual(jwtDefaults.petId);
+    });
+  });
+
+  describe('mergeIntoBodyAndMetadata', () => {
```
erunion, line 748: New tests start here.

src/operation.ts
```diff
@@ -439,15 +433,22 @@ export default class Operation {
    * Convert the operation into an array of JSON Schema schemas for each available type of parameter available on the
    * operation.
    *
-   * @param globalDefaults Contains an object of user defined schema defaults.
-   */
-  getParametersAsJsonSchema(globalDefaults?: Record<string, unknown>) {
-    if (this.parameterJsonSchema) {
```
erunion, line 445: Since this method can return a couple different versions of JSON Schema I'm removing this intermediary caching layer.

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -189,118 +203,146 @@ export default function getParametersAsJsonSchema(operation: Operation, api: OAS
   function transformParameters(): SchemaWrapper[] {
     const operationParams = operation.getParameters();
 
-    return Object.keys(types).map(type => {
```
erunion, line 192: This is all whitespace because I changed it to being set to a `transformed` constant. View this diff with whitespace changes hidden.

src/operation.ts
```diff
@@ -439,15 +433,22 @@ export default class Operation {
...
-      return this.parameterJsonSchema;
-    }
-
-    this.parameterJsonSchema = getParametersAsJsonSchema(this, this.api, globalDefaults);
-    return this.parameterJsonSchema;
+   * @param opts
+   * @param opts.globalDefaults Contains an object of user defined schema defaults.
+   * @param opts.mergeIntoBodyAndMetadata If you want the output to be two objects: body (contains
```
kanadgupta, line 438: you're obviously working in this so maybe this is making more sense in my brain but... is "split" a better descriptor what's happening here than "merge"?

src/operation.ts
```diff
@@ -439,15 +433,22 @@ export default class Operation {
...
+   *    `body` and `formData` JSON Schema) and metadata (contains `path`, `query`, `cookie`, and
+   *    `header`).
+   * @param opts.retainDeprecatedProperties If you wish to **not** split out deprecated properties
```
kanadgupta, line 441: (insert same preface as above) what about "merge" instead of "retain"?

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -189,118 +203,146 @@ export default function getParametersAsJsonSchema(operation: Operation, api: OAS
...
+          allOf: transformed.map(r => r.schema),
+        } as SchemaObject,
+        deprecatedProps: deprecatedProps.length
+          ? {
+              type: 'metadata',
+              schema: {
+                allOf: deprecatedProps,
+              } as SchemaObject,
+            }
+          : null,
```
kanadgupta, line 343: having absolutely zero knowledge of the implications of how this is all consumed, what's the reasoning behind having the fallback set to `null` vs `undefined`?

kanadgupta (approved):
I don't really have much insight into how we use this... but with that said, this new pattern seems sound to me!

src/operation.ts
erunion: It's definitely more "merge" than "split" because this is merging `query`, `path`, `cookie`, and `header` JSON Schema parameter objects into a single `metadata` JSON Schema object.

src/operation.ts
erunion: This is similar. By default in the explorer we split out deprecated properties into a `deprecatedProps` object so we can show deprecated parameters separately from regular parameters and in `api` I don't want to do this because I'd prefer deprecated parameters to (eventually) have a `/** @deprecated */` docblock over them so when you try to reference it in an SDK your IDE lets you know.

src/operation/get-parameters-as-json-schema.ts
erunion: I forget why we did this, but I agree that it should be `undefined`. Will look into it.

erunion: Actually I'm going to leave this as `null` because we end up deleting `deprecatedProps` out of the object if it's falsy here: https://github.com/readmeio/oas/blob/main/src%2Foperation%2Fget-parameters-as-json-schema.ts#L328-L329

Changing these returns to be `undefined` is going to require the `getDeprecated()` method to have some `return undefined` calls which will look weird (and I don't want to suppress the `consistent-return` ESLint rule).

PR #620, opened 2022-03-15
fix: quirks with path matching on malformed path params and params with hypens

erunion:
## 🧰 Changes

This fixes a couple problems that were found today with how our `findOperation` path matching works:

* [ ] If a path parameter was malformed and either had a trailing or preceeding curly bracket, our regex work would throw a fatal exception because it was generating invalid regex (like `:pathParam})`. We're now able to handle these.
* [ ] In my efforts to fix that issue I also discovered that if a path had a path parameter with a hyphen in it, `{path-param}`, we wouldn't be able to match on it **at all** because [path-to-regexp](https://npm.im/path-to-regexp) would only look at the first part of the `:path-param` regex we'd give it. I've added some string replacement work in our path matching code to adhoc transform `:path-param` into `:pathparam` before it's sent over.

## 🧬 QA & Testing

Check out the couple tests I added.

### DIFF
--- a/__tests__/__datasets__/path-matching-quirks.json
+++ b/__tests__/__datasets__/path-matching-quirks.json
@@ -1,5 +1,5 @@
 {
-  "openapi": "3.0.0",
+  "openapi": "3.1.0",
   "info": {
     "title": "Path Matching Quirks",
     "description": "Example API definition to cover some quirks with path matching where a query param in a path might break `Oas.findOperation()`",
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -98,16 +98,35 @@ function transformUrlIntoRegex(url: string) {
...
-  //
-  // For example if the URL is `/post/:param1::param2` we'll be escaping it to `/post/:param1\::param2`.
   return (
     path
-      .replace(/{(.*?)}/g, ':$1')
+      // This regex transforms `{pathParam}` into `:pathParam` so we can regex against it. We're
+      // also handling quirks here like if there's an optional proceeding or trailing curly bracket
+      // (`{{pathParam}` or `{pathParam}}`) as any unescaped curlys, which would be present in
+      // `:pathParam}`, will throw a regex exception.
+      .replace(/({?){(.*?)}(}?)/g, function (str, ...args) {
```
erunion, line 107: This regex was changed from `{(.*?)}` to handle optional proceeding `{` and trailing `}`.

__tests__/__datasets__/path-matching-quirks.json
```diff
@@ -1,5 +1,5 @@
 {
-  "openapi": "3.0.0",
+  "openapi": "3.1.0",
```
erunion, line 2: Changed this to 3.1 because I didn't feel like adding responses to these two new paths.

src/index.ts
Dashron: I'm surprised you don't have to escape the curly braces. aren't they reserved characters in regexes?

Dashron (approved):

src/index.ts
erunion: You're absolutely right and I'm not really sure how this works because it works in [Patterns](https://krillapps.com/patterns/) too:

Screen Shot 2022-03-15 at 10 35 53 AM

Dashron: maybe there's a default flag enabled that makes this work

PR #627, opened 2022-03-23
fix: issue where a securityScheme may be corrupted with internal library data

erunion:
## 🧰 Changes

This resolves an issue I discovered while working on https://github.com/readmeio/api/pull/422 where if you called `operation.prepareSecurity()` and then ran OAS validation against the spec you ingested into `oas` validation would fail because `operation.prepareSecurity()` would append our internal security scheme name tracking information to a `_key` property.

Tests never caught this because they were asserting directly against the corrupted spec data. 🙃


julshotal (approved):
Went over this on a call with Jon - everything looks good 👍
Only thing is adding to the test a spec validation just to double-check it validates correctly

julshotal (approved):

PR #628, opened 2022-03-29
fix: improved error handling on `Oas.findOperation()` calls

erunion:
## 🧰 Changes

This adds error handling from fully malformed paths that we can't repair to the `Oas.findOperation()` method.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,8 +147,17 @@ function generatePathMatches(paths: RMOAS.PathsObject, pathName: string, origin:
   return Object.keys(paths)
     .map(path => {
       const cleanedPath = normalizePath(path);
-      const matchStatement = match(cleanedPath, { decode: decodeURIComponent });
-      const matchResult = matchStatement(prunedPathName);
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -147,8 +147,17 @@ function generatePathMatches(paths: RMOAS.PathsObject, pathName: string, origin:
...
-      const matchResult = matchStatement(prunedPathName);
+
+      let matchResult: MatchResult;
+      try {
+        const matchStatement = match(cleanedPath, { decode: decodeURIComponent });
+        matchResult = matchStatement(prunedPathName) as MatchResult;
+      } catch (err) {
+        // If path matching fails for whatever reason (maybe they have a malformed path parameter)
+        // then we shouldn't also fail.
+        return;
```
gratcliff, line 158: Is it ok to return nothing here when we're relying on the return output? Won't this generate an index with `undefined` in it?

gratcliff (approved):

PR #634, opened 2022-04-07
fix: issue where response schemas in the original spec may get overridden

erunion:
## 🧰 Changes

This resolves a data corruption bug I discovered in https://github.com/readmeio/api where if you compile the JSON Schema for an operation response with `getResponseAsJsonSchema`, and somewhere within that response is an `example` property, we'd update the original API definition object that was supplied to `Oas` with the `examples` syntax that `getResponseAsJsonSchema` reshapes `example` to -- rendering the original API definition in a non-spec validating state.

## 🧬 QA & Testing

I added a test that runs `getResponseAsJsonSchema` against a response in our own spec and then queries the original spec inside the original `Oas` instance to ensure that the schema property where `example` was present is only set to `examples` in the generated JSON Schema.

:upsidedown-cowboy-quadboy:


julshotal (approved):
ty for adding the validation test!

PR #636, opened 2022-04-19
fix: don't attempt to clone undefined objects, resulting in a crash

erunion:
| 🚥 Fix RM-4226 |
| :-- |

## 🧰 Changes

This fixes a crash I introduced in work I did in #634 where if a response does not have a schema we were still attempting to clone it with `JSON.parse(JSON.stringify())`, which fails:

```
> JSON.parse(JSON.stringify(undefined))
VM226:1 Uncaught SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
```

### DIFF
--- a/__tests__/operation/get-response-as-json-schema.test.ts
+++ b/__tests__/operation/get-response-as-json-schema.test.ts
@@ -184,64 +184,80 @@ describe('$schema version', () => {
   });
 });
 
-describe('$ref quirks', () => {
-  it("should retain $ref pointers in the schema even if they're circular", () => {
-    const operation = circular.operation('/', 'put');
-    expect(operation.getResponseAsJsonSchema('200')).toMatchSnapshot();
-  });
+describe('quirks', () => {
+  it('should not crash out when pulling a response that has no schema', () => {
+    const operation = responses.operation('/response-with-example-and-no-schema', 'get');
 
-  it('should default the root schema to a `string` if there is a circular `$ref` at the root', () => {
-    const operation = circular.operation('/', 'put');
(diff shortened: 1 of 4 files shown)


__tests__/operation/get-response-as-json-schema.test.ts
```diff
@@ -184,64 +184,80 @@ describe('$schema version', () => {
...
+            components: circular.api.components,
+          },
+        },
+      ]);
     });
 
-    const definition = readme.getDefinition();
-    const authUnauthorizedResponse = definition.components.responses.authUnauthorized as ResponseObject;
+    it('should not override object references', async () => {
+      const readme = await import('@readme/oas-examples/3.0/json/readme.json').then(res => res.default).then(Oas.init);
```
dok, line 225: Whoa on the `await import`

erunion: :typescript-sparkle:

src/lib/clone-object.ts
```diff
@@ -1,3 +1,7 @@
 export default function cloneObject<T>(obj: T): T {
+  if (obj === undefined) {
```
dok, line 2:
```suggestion
  if (typeof obj === 'undefined') {
```

dok: An alternative option! Not sure if there are pros/cons

dok (approved):

src/lib/clone-object.ts
erunion: nah this works too and is a little cleaner

Dashron (approved):
so many tests, so few changes.

PR #637, opened 2022-04-24
feat: updating response JSON Schema generation to auto-add enums to descriptions

erunion:
| 🚥 Fix RM-988 |
| :-- |

## 🧰 Changes

This updates our `Operation.getResponseAsJsonSchema()` accessor to auto-add a list of enums into generated descriptions if enums are present.

This is how it'll look in the real world:

Screen Shot 2022-04-24 at 12 00 30 PM

## 🧬 QA & Testing

See the test and example dataset I added.

### DIFF
--- a/__tests__/__datasets__/response-enums.json
+++ b/__tests__/__datasets__/response-enums.json
@@ -1,0 +1,82 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Responses w/ enums",
+    "description": "This is a demo API definition for our support to supplement response schema descriptions with enums are present.",
+    "version": "1.0.0"
+  },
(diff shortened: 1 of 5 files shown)


kellyjosephprice (approved):
LGTM

Dashron:
under what conditions do we plan on adding these to the body, always?

I worry a little bit about anyone who already has this data in their description and now will be surfacing duplicate information. That said, I wouldn't have had this concern if we were adding them anywhere but the description so maybe it's a non issue for now and better to have it somewhere than to not have it at all.

Dashron (approved):

erunion:
@Dashron We're going to always be adding these to response descriptions.

PR #677, opened 2022-08-17
feat: porting the library portion of `oas-reducer` into this library

erunion:
| 🚥 Fix RM-5114 |
| :-- |

## 🧰 Changes

* [x] Migrates the library portion of [oas-reducer](https://npm.im/oas-reducer) into being a core part of this library.
    
> ℹ️ To avoid bloating the `oas` bundle size when compiled with Webpack I'm opting to provide this reducer functionality as a sub library as If I included it as `.reducer()` method on the main `Oas` export that would also bundle up the reducers heavy [jsonpath](https:///npm.im/jsonpath) requirement.

## 🧬 QA & Testing

Tests here have been pulled directly over from `oas-reducer`.


kanadgupta (changes requested):
a few small questions/comments but otherwise looking good!

src/lib/reducer.ts
```diff
@@ -0,0 +1,184 @@
...
+    $refs.add(currRef);
+    accumulateUsedRefs(schema, $refs, currRef);
+  });
+}
+
+/**
+ * With an array of tags or object of paths+method combinations, reduce an OpenAPI definition to a
+ * new definition that just contains those tags or path + methods.
+ *
+ * @example <caption>Reduce by an array of tags only.</caption>
```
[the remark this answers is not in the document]
erunion: I stumbled upon it while working on `api` and really like it: https://github.com/readmeio/api/blob/main/packages/api/test/__fixtures__/sdk/readme/index.ts#L25-L49

src/lib/reducer.ts
```diff
@@ -0,0 +1,186 @@
...
+
+import jsonPath from 'jsonpath';
+import jsonPointer from 'jsonpointer';
+import { version as getAPIDefinitionVersion } from 'oas-normalize/dist/lib/utils';
+
+export type ReducerOptions = {
+  /** An array of tags in the OpenAPI definition to reduce by. */
+  paths?: Record<string, string[] | '*'>;
+  /** A key-value object of path + method combinations to reduce by. */
+  tags?: string[];
```
kanadgupta, line 11: I think these descriptions should be switched:
```suggestion
  /** A key-value object of path + method combinations to reduce by. */
  paths?: Record<string, string[] | '*'>;
  /** An array of tags in the OpenAPI definition to reduce by. */
  tags?: string[];
```

src/lib/reducer.ts
```diff
@@ -0,0 +1,186 @@
...
+ * @example <caption>Reduce by a specific path and methods.</caption>
+ * { APIDEFINITION, { paths: { '/pet': ['get', 'post'] } } }
+ *
+ * @example <caption>Reduce by a specific path and all methods it has.</caption>
+ * { APIDEFINITION, { paths: { '/pet': '*' } } }
+ *
+ * @param definition A valid OpenAPI 3.x definition
+ * @param opts Option configuration to reduce by. See the README for details.
+ * @param opts.tags An array of tags in the OpenAPI definition to reduce by.
+ * @param opts.paths A key-value object of path + method combinations to reduce by.
```
kanadgupta, line 60: Probably safe to remove this now?
```suggestion
```

src/lib/reducer.ts
```diff
@@ -0,0 +1,186 @@
...
+    accumulateUsedRefs(schema, $refs, currRef);
+  });
+}
+
+/**
+ * With an array of tags or object of paths+method combinations, reduce an OpenAPI definition to a
+ * new definition that just contains those tags or path + methods.
+ *
+ * @example <caption>Reduce by an array of tags only.</caption>
+ * { APIDEFINITION, { tags: ['pet] } }
```
kanadgupta, line 49: Aren't these technically two different arguments?
```suggestion
 * ( APIDEFINITION, { tags: ['pet] } )
```

kanadgupta (changes requested):

src/lib/reducer.ts
erunion: :ope:

src/lib/reducer.ts
erunion: Yeah that's what I was trying to document originally by having the example be only `{ tags: ['pet] }`.

kanadgupta: gotcha—i haven't seen a ton of JSDocs like these in the wild so not sure how these examples should look!

kanadgupta (approved):

src/lib/reducer.ts
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * @example <caption>Reduce by a specific path and methods.</caption>
+ * { APIDEFINITION, { paths: { '/pet': ['get', 'post'] } } }
+ *
+ * @example <caption>Reduce by a specific path and all methods it has.</caption>
+ * { APIDEFINITION, { paths: { '/pet': '*' } } }
+ *
+ * @param definition A valid OpenAPI 3.x definition
+ * @param opts Option configuration to reduce by. See the README for details.
```
kanadgupta, line 58: Sorry one more small docs-related comment—is this "See the README for details." comment still applicable?

erunion: Oh no, I'll remove that.

PR #691, opened 2022-09-13
docs: add commands section to README

jaredsmithse:
| 🚥 Fix RM-XXX |
| :-- |

## 🧰 Changes

I believe it would be good to show what other commands the tool has besides the `init` command. I think that section is still important, and I've seen other tools show `help` output in their READMEs. This is a shot at doing the same for this tool 🙇 

## 🧬 QA & Testing

Visual verification of newly rendered Markdown should be sufficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,18 @@ oas init
 
 It will walk you through how to document your API with a OpenAPI 3.0 Spec.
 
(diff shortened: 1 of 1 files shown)


erunion (approved):
thanks!

PR #723, opened 2022-12-07
feat(analyzer): improvements to reporting and circular $ref discovery

erunion:
## 🧰 Changes

* [x] Updates every analyzer[^1] query to return a proper `$ref` schema pointer (one that's prefixed with a `#`).
* [x] Updates [@readme/openapi-parser](https://npm.im/@readme/openapi-parser) and [@readme/json-schema-ref-parser](https://npm.im/@readme/json-schema-ref-parser) to incorporate a new API I designed in it for surfacing circular `$ref` pointers to us when they're discovered during dereferencing. https://github.com/readmeio/json-schema-ref-parser/pull/58
* [x] Created a new `Oas.getCircularReferences()` method that will use the afformetioned API to return a list of every circular `$ref` that exists within the API definition.

Finally, I've updated the `circularRefs` OpenAPI analyzer query to use these new APIs so it returns **exactly** what circular `$ref`s exist within an API definition, and not all `$ref`s that still remain in the API definition after dereferencing:

screen_shot_2022-12-07_at_1 32 37_pm

[^1]: What's the analyzer? It's the new JSONPath query system I've been building for the new [`openapi:uses` command in rdme](https://github.com/readmeio/rdme/pull/698).


src/index.ts
```diff
@@ -766,17 +768,40 @@ export default class Oas {
...
    */
   async dereference(
     opts: {
+      /**
+       * A callback method can be supplied to be called when dereferencing is complete. Used for
+       * debugging that the mulit-promise handling within this method works.
+       *
+       * @private
+       */
+      cb?: () => void;
```
erunion, line 799: I don't like having this option here just for testing purposes but because I changed way we use `@readme/json-schema-ref-parser` to instantiate it with a class constructor it's really hard with Jest to spy on `new $RefParser.dereference()` now to ensure that it's only ever called once.

Adding this callback in here, and then calling it in the `.then()` below replicates this behavior.

src/index.ts
```diff
@@ -824,13 +851,30 @@ export default class Oas {
...
       })
       .then((dereferenced: RMOAS.OASDocument) => {
+        let circularRefs: string[] = [];
+        if (parser.$refs.circular) {
+          circularRefs = parser.$refs.circularRefs.map(pointer => {
+            // The circular $refs that are coming out of `json-schema-ref-parser` are prefixed
+            // with the schema path (file path, url, whatever) that the schema exists in. Because
+            // we don't care about this information for this reporting mechanism, and only the
+            // $ref pointer, we're removing it.
+            return `#${pointer.split('#')[1]}`;
```
erunion, line 861: Example pointer coming back from `json-schema-ref-parser` looks like `/path/to/file.yaml#/components/schemas/name`. We only need the latter part so that's what this is doing here.

kanadgupta (approved):
a few small nits but otherwise LGTM!

PR #731, opened 2023-02-23
feat: add case insensitivity for reducer

### LINKED ISSUE #678: OAS reducer library should be able to reduce irregardless of the supplied casing
In `src/lib/reducer` if either the casing in the options for `paths` or `tags` or `paths` methods array doesn't match what's in the OpenAPI definition it won't be matched and will be removed.

Casing shouldn't matter.

mattcole19:
| 🚥 Fixes  #678  |
| :---------------- |

## 🧰 Changes

Modified reducer to keep matching paths, tags, and/or methods regardless of casing.

## 🧬 QA & Testing

Changed casing in existing tests to ensure changes work.

### DIFF
--- a/__tests__/lib/reducer.test.ts
+++ b/__tests__/lib/reducer.test.ts
@@ -20,7 +20,7 @@ describe('reducer', () => {
 
   describe('tag reduction', () => {
     it('should reduce by tags', () => {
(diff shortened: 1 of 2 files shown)


erunion (approved):
this is awesome! thanks so much.

PR #732, opened 2023-02-28
fix: only adding components to generated JSON Schema if we need to

erunion:
## 🧰 Changes

This fixes a pretty large performance issue in `Operation.getParametersAsJSONSchema()` where if an API Definition has a very large schema in a component we will bog down pretty hard converting it to JSON Schema even if we don't need to.

For example this excerpted component schema:

screen_shot_2023-02-27_at_10 37 34_pm

Because these `$ref` pointers, and the parental schema itself, are normally able to be safely dereferenced, we're still adding the into the generated JSON Schema for that parameter or request body that we're generating JSON Schema for on the off chance that they might be used within our API Explorer for schemas that may be circular and we're able to safely render them there.

The problem though is that we're doing this work for everything and everybody. Even if your entire API definition has no circular `$ref` pointers and can 100% be safely dereferenced we still process the entire `#/components` block, converting every component into JSON Schema, and then add that to the JSON Schema we *actually* want.

The operation that above screenshot? That comes from [this page](https://apis-cedro.readme.io/reference/get_services-negotiation-brokerserviceaccountinformation-nomeusuario-conta-codmercado) where because we're re-processing all components, components that have already been processed because they were dereferenced, a single call to `Operation.getParametersAsJSONSchema()` takes ~2000ms.

Remember that this is happening for every user, every API definition, everything.

columbo.

## 🧬 QA & Testing

To fix this I went ahead and fixed the glitch. Because we have some internal logging work within this generator to keep track of when we encounter a `$ref` pointer that wasn't able to be dereferenced we know if we'll need to ever reference the `#/components` object at all. The results are pretty goddamn mindblowing.

That operation that took 2000ms to generate JSON Schema? Now? 3ms. T H R E E.

Try it[^1] if you don't believe me, I honestly didn't either.

```ts
import spec from './cedro.json';
import Oas from './src';

declare global {
  interface Console {
    logx: any;
  }
}

console.logx = (obj: any) => {
  console.log(require('util').inspect(obj, false, null, true));
};

const oas = new Oas(spec as any);

(async () => {
  await oas.dereference();

  const START = process.hrtime();

  const op = oas.operation(
    '/services/negotiation/brokerServiceAccountInformation/{nomeUsuario}/{conta}/{codMercado}',
    'get'
  );

  console.logx(op.getParametersAsJSONSchema());

  const END = process.hrtime(START);
  console.log(`${(END[0] * 1000000000 + END[1]) / 1000000}ms`);
})();
```

There's still a lot of improvement here for when we *do* need to inject these components into the schema (mostly only cases where circular `$ref` pointers are present) but to do that we'll need to process these stored up `$ref` pointers that we found, find them in `#/components` if they exist and then add them. Just to curb the bleeding on this particular user, and this particularly disastrous performance issue I'm opting to not do that now. Large schemas with circular refs will still have poorly unoptimized JSON Schema generation.

[^1]: Use this spec: https://apis-cedro.readme.io/openapi/63eace87ab83b2001e63306c

### DIFF
--- a/src/operation/get-parameters-as-json-schema.ts
+++ b/src/operation/get-parameters-as-json-schema.ts
@@ -179,19 +179,31 @@ export default function getParametersAsJSONSchema(
       return false;
     }
 
-    const components: Partial<ComponentsObject> = {};
+    const components: Partial<ComponentsObject> = {
+      /**
+       * Initializing an empty `components[componentType] = {}` object within the `forEach` below
+       * is incredibly slow because each of these component types has a wide variety of shapes. So
+       * in order to not have TS compilation times that takes literally **seconds** because of a
+       * single line we're instead opting to prefill this object with some empty placeholders that
+       * we'll later remove if they didn't get used.
+       *
+       * Obviously not ideal but I'd rather have a couple lines of boilerplate nonsense than having
+       * to wait a noticeably frustrating amoutn of time for TS Intellisense to reload itself after
+       * you save a line in any file.
+       */
+      examples: {},
+      schemas: {},
+      responses: {},
+      parameters: {},
+      requestBodies: {},
+      headers: {},
+      securitySchemes: {},
+      links: {},
+      callbacks: {},
+    };
 
     Object.keys(api.components).forEach((componentType: keyof ComponentsObject) => {
       if (typeof api.components[componentType] === 'object' && !Array.isArray(api.components[componentType])) {
-        /**
-         * Typescript is INCREDIBLY SLOW parsing this one line. I think it's because of the large
-         * variety of types that that object could represent but I can't yet think of a way to get
-         * around that.
-         *
-         * @todo
-         */
-        components[componentType] = {};
-
         Object.keys(api.components[componentType]).forEach(schemaName => {
           const componentSchema = cloneObject(api.components[componentType][schemaName]);
           components[componentType][schemaName] = toJSONSchema(componentSchema as SchemaObject, {
@@ -200,6 +212,23 @@ export default function getParametersAsJSONSchema(
             transformer: opts.transformer,
           });
         });
+      }
+    });
+
+    // If none of our above component type placeholders got used let's clean them up.
+    [
+      'examples',
+      'schemas',
+      'responses',
+      'parameters',
+      'requestBodies',
+      'headers',
+      'securitySchemes',
+      'links',
+      'callbacks',
+    ].forEach((componentType: keyof ComponentsObject) => {
+      if (!Object.keys(components[componentType]).length) {
+        delete components[componentType];
       }
     });
 
@@ -378,12 +407,18 @@ export default function getParametersAsJSONSchema(
     return null;
   }
 
-  const components = transformComponents();
-
   const typeKeys = Object.keys(types);
-  return [transformRequestBody()]
-    .concat(...transformParameters())
-    .filter(Boolean)
+  const jsonSchema = [transformRequestBody()].concat(...transformParameters()).filter(Boolean);
+
+  // We should only include `components`, or even bother transforming components into JSON Schema,
+  // if we either have circular refs or if we have discriminator mapping refs somehwere and want to
+  // include them.
+  const shouldIncludeComponents =
+    hasCircularRefs || (hasDiscriminatorMappingRefs && opts.includeDiscriminatorMappingRefs);
+
+  const components = shouldIncludeComponents ? transformComponents() : false;
+
+  return jsonSchema
     .map(group => {
       /**
        * Since this library assumes that the schema has already been dereferenced, adding every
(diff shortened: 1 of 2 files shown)


src/operation/get-parameters-as-json-schema.ts
```diff
@@ -179,19 +179,31 @@ export default function getParametersAsJSONSchema(
...
     Object.keys(api.components).forEach((componentType: keyof ComponentsObject) => {
       if (typeof api.components[componentType] === 'object' && !Array.isArray(api.components[componentType])) {
-        /**
-         * Typescript is INCREDIBLY SLOW parsing this one line. I think it's because of the large
-         * variety of types that that object could represent but I can't yet think of a way to get
-         * around that.
-         *
-         * @todo
-         */
-        components[componentType] = {};
```
erunion, line 193: This has been plauging this repo, and any implementation of `oas` since we migrated this library to TS. I frankly got sick of having to wait 2+ seconds for TS Intellisense to reload it self, and not being able to get anywhere with trying to improve the TS perf on this that I decided to blow it up into the above refactor.

I don't love it but I _do_ love not having to literally wait forever for Intellisense to reload itself after saving any change to any TS file in this repo.

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -392,13 +427,9 @@ export default function getParametersAsJSONSchema(
        *
        * @todo
        */
-      if (components) {
-        // We should only include components if we've got circular refs or we have discriminator
-        // mapping refs (we want to include them).
-        if (hasCircularRefs || (hasDiscriminatorMappingRefs && opts.includeDiscriminatorMappingRefs)) {
-          // Fixing typing and confused version mismatches
-          (group.schema.components as ComponentsObject) = components;
-        }
```
erunion, line 401: I was so close to fixing this performance issue when I added line 398. If only I had gone one step further by not running `transformComponents()`. Ah well.

src/operation/get-parameters-as-json-schema.ts
domharrington: Out of interest how did you debug that this line was causing such a slowdown?

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -378,12 +407,18 @@ export default function getParametersAsJSONSchema(
...
-  const components = transformComponents();
-
   const typeKeys = Object.keys(types);
-  return [transformRequestBody()]
-    .concat(...transformParameters())
-    .filter(Boolean)
+  const jsonSchema = [transformRequestBody()].concat(...transformParameters()).filter(Boolean);
+
+  // We should only include `components`, or even bother transforming components into JSON Schema,
+  // if we either have circular refs or if we have discriminator mapping refs somehwere and want to
```
domharrington, line 414: Spelling nit:
```suggestion
  // if we either have circular refs or if we have discriminator mapping refs somewhere and want to
```

domharrington (approved):

src/operation/get-parameters-as-json-schema.ts
erunion: I don't remember exactly how @Dashron and I narrowed this down originally but I believed it involved some sort of tsc disgnosis run and looking through the call tree.

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -179,19 +179,31 @@ export default function getParametersAsJSONSchema(
...
+    const components: Partial<ComponentsObject> = {
+      /**
+       * Initializing an empty `components[componentType] = {}` object within the `forEach` below
+       * is incredibly slow because each of these component types has a wide variety of shapes. So
+       * in order to not have TS compilation times that takes literally **seconds** because of a
+       * single line we're instead opting to prefill this object with some empty placeholders that
+       * we'll later remove if they didn't get used.
+       *
+       * Obviously not ideal but I'd rather have a couple lines of boilerplate nonsense than having
+       * to wait a noticeably frustrating amoutn of time for TS Intellisense to reload itself after
```
kanadgupta, line 191:
```suggestion
       * to wait a noticeably frustrating amount of time for TS Intellisense to reload itself after
```

kanadgupta (approved):
:hell-yeah:

src/operation/get-parameters-as-json-schema.ts
Dashron: I did this: https://github.com/microsoft/TypeScript/wiki/Performance#performance-tracing it gives you a performance file that can be viewed in chromes debugger, and made it very clear that that one line was the slowest. commenting it out proved the issue.

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -179,19 +179,31 @@ export default function getParametersAsJSONSchema(
       return false;
     }
 
-    const components: Partial<ComponentsObject> = {};
+    const components: Partial<ComponentsObject> = {
```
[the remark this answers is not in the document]
Dashron: so don't use partial with large objects. interesting. I wonder if we can add a linter or some other test for that.

PR #739, opened 2023-03-07
fix: bug where request example selection wouldn't always use the first

erunion:
| 🚥 Fixes RM-6662 |
| :---------------- |

## 🧰 Changes

This is a weird one! Right now if you have a request body that utilizes `examples` and have multiple examples within that our `example` determination for `Operation.getParametersToJSONSchema()` will now will pick an example from each down the list as it processes property.

For example:

Screen Shot 2023-03-07 at 2 29 16 PM

Reason this is happening is because the `schema.examples` array we're working with in `lib/openapi-to-json-schema` is a reference and when we run `.shift()` on it we effectively remove that example from being used for the next property we're processing JSON Schema for.

🫠


kanadgupta (approved):
lmao

PR #753, opened 2023-04-18
fix(reducer): make sure `$ref`s are actual pointers

kanadgupta:
## 🧰 Changes

@andrew-rdme was seeing the following error when attempting to reduce a file:

```
$ref.substring is not a function
```

which is coming from this line:

https://github.com/readmeio/oas/blob/6162c8445a2dd2727944105ff76d4b2f39360be5/src/lib/reducer.ts#L33

This is happening because of an edge case where a spec file might have an object definition with a property called `$ref`. In other words, say the object were to look something like this:

```json
{
  "hello": "world",
  "$ref": "example"
}
```

then its schema definition would look something like:

```json
{
  "type": "object",
  "properties": {
    "hello": {
      "type": "string",
      "description": "A property called `hello`"
    },
    "$ref": {
      "type": "string",
      "description": "A property called $ref to see what happens"
    }
  }
}
```

The reducer was looking blindly for `$ref` keys and assuming their values are always strings. This tweaks the logic ever-so-slightly to ensure that the value of the `$ref` is indeed a string before we attempt to load its `$ref`.

## 🧬 QA & Testing

Added a fixture and a test to confirm that the fix works as expected.

### DIFF
--- a/__tests__/__datasets__/petstore-ref-quirks.json
+++ b/__tests__/__datasets__/petstore-ref-quirks.json
@@ -1,0 +1,997 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "description": "This is a sample server Petstore server.  You can find out more about Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).  For this sample, you can use the api key `special-key` to test the authorization filters.",
+    "version": "1.0.0",
+    "title": "Swagger Petstore",
+    "termsOfService": "http://swagger.io/terms/",
+    "contact": {
+      "email": "apiteam@swagger.io"
+    },
+    "license": {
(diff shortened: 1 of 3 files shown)


__tests__/__datasets__/petstore-ref-quirks.json
```diff
@@ -0,0 +1,997 @@
...
+          },
+          "status": {
+            "type": "string",
+            "description": "Order Status",
+            "enum": ["placed", "approved", "delivered"]
+          },
+          "$ref": {
+            "type": "string",
+            "description": "A property called $ref to see what happens"
+          },
```
kanadgupta, line 856: This file is mostly just `petstore`, this is the only chunk that I added.

erunion (approved):
Ill tag a release tomorrow

PR #754, opened 2023-04-18
ci: run tests on node 20

kanadgupta:


.github/workflows/ci.yml
```diff
@@ -25,9 +26,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Install npm@7
-        run: npm install -g npm@7
-
```
kanadgupta, line 30: Figured we have good enough coverage with our Node 14 tests to not do this!

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
           - 16
           - 18
           - 19
```
kanadgupta, line 18: Kept Node 19 in here since it doesn't EOL until June: https://nodejs.dev/en/about/releases/

erunion (approved):

PR #756, opened 2023-04-28
fix: use the last description in a merged allOf schema

erunion:
## 🧰 Changes

When [json-schema-merge-allof](https://npm.im/json-schema-merge-allof) merges an `allOf` and two or more schemas within that share the same property, but not the same description, the first description seen is always going to be the one used in the resulting merged schema.

For example this schema:

```js
{
  allOf: [
    {
      type: 'object',
      properties: {
        uri: {
          type: 'string',
          description: 'first uri description',
        },
        robots: {
          type: 'string',
          description: 'first robots description',
        },
        meta: {
          type: 'string',
          description: 'only meta description',
        },
      },
    },
    {
      type: 'object',
      properties: {
        uri: {
          description: 'second uri description',
          readOnly: true,
        },
        robots: {
          description: 'second robots description',
          readOnly: true,
          deprecated: true,
        },
      },
    },
  ],
}
```

When it's merged, the description for `uri` will always be "first uri description". Because when folks use `allOf` it's sometimes to merge a common schema into a more specific one, not using the `description` from the applied schemas doesn't make sense.

With this work, the merged schema's `uri` property will have a `description` of "second uri description".

### DIFF
--- a/src/lib/openapi-to-json-schema.ts
+++ b/src/lib/openapi-to-json-schema.ts
@@ -310,6 +310,14 @@ export default function toJSONSchema(
         schema = mergeJSONSchemaAllOf(schema as RMOAS.JSONSchema, {
           ignoreAdditionalProperties: true,
           resolvers: {
+            // `merge-json-schema-allof` by default takes the first `description` when you're
+            // merging an `allOf` but because generally when you're merging two schemas together
+            // with an `allOf` you want data in the subsequent schemas to be applied to the first
+            // and `description` should be a part of that.
+            description: (obj: string[]) => {
+              return obj.slice(-1)[0];
+            },
+
             // `merge-json-schema-allof` doesn't support merging enum arrays but since that's a
             // safe and simple operation as enums always contain primitives we can handle it
             // ourselves with a custom resolver.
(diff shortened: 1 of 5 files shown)


src/lib/openapi-to-json-schema.ts
```diff
@@ -310,6 +310,14 @@ export default function toJSONSchema(
         schema = mergeJSONSchemaAllOf(schema as RMOAS.JSONSchema, {
           ignoreAdditionalProperties: true,
           resolvers: {
+            // `merge-json-schema-allof` by default takes the first `description` when you're
+            // merging an `allOf` but because generally when you're merging two schemas together
+            // with an `allOf` you want data in the subsequent schemas to be applied to the first
+            // and `description` should be a part of that.
+            description: (obj: string[]) => {
+              return obj.slice(-1)[0];
+            },
```
erunion, line 319: Here's the line in `json-schema-merge-allof` where it's always using the first description instead of the last. https://github.com/mokkabonna/json-schema-merge-allof/blob/master/src/index.js#L229

erunion: It also does this same behavior for `$id`, `$ref`, `$schema`, `default`, and `title`. I think the first[^1] behavior is fine for `$id`, `$ref`, and `$schema`, but we might want to also do this same last pick for `default` and `title`.

[^1]: https://www.youtube.com/watch?v=ciG-Xs7mBwU

Dashron (approved):
I think this makes sense!

PR #760, opened 2023-05-03
feat: improved support for mixed types and `nullable`

erunion:
| 🚥 Fixes RM-4713, RM-4328, CX-68, RM-1814 |
| :---------------- |

## 🧰 Changes

* [x] Adds support for OpenAPI 3.1's new `type: [...]` mixed type declaration. When we encounter one of these mixed types during JSON Schema generation, to ease our form system, we're automatically converting them into a `oneOf` schema.
  * If `null` is one of these options then it's removed from the `type` array and `nullable: true` is added to the other available options.
* [x] Adds support for `nullable`.

### DIFF
--- a/__tests__/lib/openapi-to-json-schema.test.ts
+++ b/__tests__/lib/openapi-to-json-schema.test.ts
@@ -12,23 +12,88 @@ beforeAll(async () => {
   await petstore.dereference();
 });
 
-test('should preserve our `x-readme-ref-name` extension', () => {
-  expect(
-    toJSONSchema({
+describe('`type` support', () => {
+  describe('mixed types', () => {
+    it('should transform a mixed type into a oneOf', () => {
+      expect(
+        toJSONSchema({
+          type: ['string', 'number'],
+          description: 'tktk',
+        })
+      ).toStrictEqual({
+        oneOf: [
+          { type: 'string', description: 'tktk' },
+          { type: 'number', description: 'tktk' },
+        ],
+      });
+    });
+
+    it('should remove duplicate entries', () => {
+      expect(
+        toJSONSchema({
+          type: ['string', 'string'],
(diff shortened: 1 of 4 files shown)


src/lib/openapi-to-json-schema.ts
```diff
@@ -367,6 +362,35 @@ export default function toJSONSchema(
...
+      // If we have a `null` type but there's only two types present then we can remove `null` as
+      // an option and flag the whole schema as `nullable`.
+      if (schema.type.includes('null')) {
+        schema.type = schema.type.filter(t => t !== 'null');
+        schema.nullable = true;
+      }
+
+      if (schema.type.length === 1) {
+        schema.type = schema.type.shift();
+      } else {
```
kanadgupta, line 379: wait so this block just makes the `schema.type` array empty, right? why are we doing this?

(not disagreeing, just not following)

__tests__/lib/openapi-to-json-schema.test.ts
```diff
@@ -908,14 +973,14 @@ describe('`example` / `examples` support', () => {
 
     it('should function through the normal workflow of retrieving a json schema and feeding it an initial example', () => {
       const operation = petstore.operation('/pet', 'post');
-      const schema: SchemaObject = operation.getParametersAsJSONSchema()[0].schema;
+      const schema = operation.getParametersAsJSONSchema()?.[0].schema as SchemaObject;
```
kanadgupta, line 976: isn't it better to avoid `as` here?

erunion: It is but the types on all this JSON Schema and OpenAPI stuff is a complete disaster.

kanadgupta (approved):

PR #764, opened 2023-05-23
fix: ensure summaries + descriptions are strings

kanadgupta:
| 🚥 Fixes ISSUE_ID |
| :---------------- |

## 🧰 Changes

- [x] Fixes an error that users were seeing in [this issue](https://github.com/readmeio/rdme/issues/814): `this.schema.description.trim is not a function`
- [x] Reverts our `.trim()` logic since that's technically unexpected behavior.

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/src/operation.ts
+++ b/src/operation.ts
@@ -78,8 +78,8 @@ export default class Operation {
   }
 
   getSummary(): string {
-    if (this.schema?.summary) {
-      return this.schema.summary.trim();
+    if (this.schema?.summary && typeof this.schema.summary === 'string') {
+      return this.schema.summary;
     } else if (this.api.paths[this.path].summary) {
       return this.api.paths[this.path].summary;
     }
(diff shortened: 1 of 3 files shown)


src/operation.ts
```diff
@@ -78,8 +78,8 @@ export default class Operation {
   }
 
   getSummary(): string {
-    if (this.schema?.summary) {
-      return this.schema.summary.trim();
+    if (this.schema?.summary && typeof this.schema.summary === 'string') {
+      return this.schema.summary;
     } else if (this.api.paths[this.path].summary) {
```
erunion, line 83: Want to add this `typeof` logic to these other blocks? These ones handle callback objects.

erunion (changes requested):

erunion (approved):

erunion (approved):

PR #778, opened 2023-07-02
feat: surface what the default value is for an enum 

darrenyong:
| 🚥 Fixes RM-7104 |
| :---------------- |

## 🧰 Changes

Currently, we don't surface what the default value is for an enum anywhere so users won't know if there is a default available or not aside from the default form state. This change adds the default value to the schema description so users can tell what the description is.

### DIFF
--- a/src/lib/openapi-to-json-schema.ts
+++ b/src/lib/openapi-to-json-schema.ts
@@ -786,6 +786,13 @@ export default function toJSONSchema(
 
   // Only add a default value if we actually have one.
   if ('default' in schema && typeof schema.default !== 'undefined') {
+    // If it's an enum, add the default to the description if it's not in the response schema.
+    if ('enum' in schema && !addEnumsToDescriptions) {
+      schema.description = schema.description
+        ? `${schema.description} default: ${schema.default}`
+        : `default: ${schema.default}`;
+    }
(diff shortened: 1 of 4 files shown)


src/lib/openapi-to-json-schema.ts
```diff
@@ -786,6 +786,13 @@ export default function toJSONSchema(
 
   // Only add a default value if we actually have one.
   if ('default' in schema && typeof schema.default !== 'undefined') {
+    // If it's an enum, add the default to the description if it's not in the response schema.
+    if ('enum' in schema && !addEnumsToDescriptions) {
+      schema.description = schema.description
+        ? `${schema.description} default: ${schema.default}`
+        : `default: ${schema.default}`;
```
erunion, line 793:
```suggestion
        ? `${schema.description}\n\nDefault: ${schema.default}`
        : `Default: ${schema.default}`;
```

erunion: Also not at all opposed to you trimming any trailing `\n`'s off descriptions we set in the JSON Schema because with this it'll end up adding 3 `\n`'s for those cases which'll look like crap.

erunion (approved):

PR #780, opened 2023-07-11
feat(jsonschema): dereference `default` into schemas when compiling them

erunion:
## 🧰 Changes

We've got a problem in our form system where currently if any schema has a `default`, regardless or not if it's `required` that data will show up in the form and in the code snippet.

Fixing this is something else though because while our form system does not have support for knowing if something is `required` or not, it's easy to fix. What isn't easy to fix is that our form system has support for setting `default` above the object you're defining. For example:

```js
{
  type: 'object',
  default: { level0: { level1: [1, 2, 3], level2: [5, 6, 7] } },
  properties: {
    level0: {
      type: 'object',
      properties: {
        level1: {
          type: 'array',
          items: { type: 'number' },
        },
        level2: {
          type: 'array',
          items: { type: 'number' },
        },
      },
      required: ['level1'],
    }
  }
}
```

Being able to filter out all of these child defaults from being shown is extremely difficult because when it's processing `level2` it knows it's not required but it doesn't know that it already has had a default set -- one that it shouldn't.

To resolve this we've devised a solution to, akin to how we handle this same problem with `example` and `examples`, to dereference these values down into their children schemas. With this fix, the above schema will now be compiled to the following:

```js
{
  type: 'object',
  properties: {
    level0: {
      type: 'object',
      properties: {
        level1: {
          type: 'array',
          default: [1, 2, 3],
          items: { type: 'number' },
        },
        level2: {
          type: 'array',
          default: [5, 6, 7],
          items: { type: 'number' },
        },
      },
      required: ['level1'],
    }
  }
}
```

There is a pretty huge caveat with this work though in that it does **not** supporting these parental defaults against polymorphic schemas:

```js
{
  type: 'object',
  properties: {
    level1: {
      type: 'object',
      default: {
        level2: {
          leaf1: 'a',
        },
      },
      properties: {
        level2: {
          oneOf: [
            {
              type: 'object',
              properties: {
                leaf1: {
                  type: 'string',
                },
              },
            },
          ],
        },
      },
    },
  },
}
```

In this case, the `default` will be completely dereferenced out of existence because we can't easily apply it to the items within the `oneOf`. Granted supporting these kinds of default objects on polymorphic schemas is an extreme edge case so I'm comfortable not supporting that right now.


src/lib/openapi-to-json-schema.ts
```diff
@@ -59,9 +57,14 @@ export interface toJSONSchemaOptions {
   isPolymorphicAllOfChild?: boolean;
 
   /**
-   * Array of parent schemas to utilize when attempting to path together examples.
+   * Array of parent `default` schemas to utilzie when attempting to path together schema defaults.
```
darrenyong, line 60: > utilzie
small typo!

PR #781, opened 2023-07-19
feat(jsonschema): adding support for invalid `required: true` properties

erunion:
| 🚥 Resolves RM-7587 |
| :---------------- |

## 🧰 Changes

JSON Schema does not have any support for `required: <boolean>` but because some of our users do this, and it does not throw OpenAPI validation because of some extremely loose typings around `schema` in the official JSON Schema definitions that the OAI offers[^1], we're opting to support these users and upgrade their invalid `required` definitions into ones that our tooling can interpret.

[^1]: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json#L1114-L1121

### DIFF
--- a/__tests__/lib/openapi-to-json-schema/required.test.ts
+++ b/__tests__/lib/openapi-to-json-schema/required.test.ts
@@ -1,0 +1,128 @@
+import type { SchemaObject } from '../../../src/rmoas.types';
+
+import toJSONSchema from '../../../src/lib/openapi-to-json-schema';
+import generateJSONSchemaFixture from '../../__fixtures__/json-schema';
+
+describe('`required` support in `openapi-to-json-schema`', () => {
+  it('should support required', () => {
+    const schema: SchemaObject = {
+      type: 'object',
+      default: {
+        array_parent: [1, 2, 3],
+      },
+      properties: {
+        array: {
+          type: 'array',
+          default: ['foo', 'bar'],
+          items: {
+            type: 'string',
+          },
+        },
+        array_optional: {
+          type: 'array',
+          default: ['foo', 'bar'],
+          items: {
+            type: 'string',
+          },
+        },
+        array_parent: {
+          type: 'array',
+          items: {
+            type: 'number',
+          },
+        },
+      },
+      required: ['array'],
+    };
+
+    const compiled = toJSONSchema(schema);
+    expect(compiled.required).toStrictEqual(['array']);
+  });
+
+  describe('`required` boolean quirks', () => {
+    it('should support a nested object `required` boolean', () => {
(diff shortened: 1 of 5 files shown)


src/lib/openapi-to-json-schema.ts
```diff
@@ -431,6 +431,15 @@ export default function toJSONSchema(
...
             schema[polyType][idx] = toJSONSchema(item as RMOAS.SchemaObject, polyOptions);
           }
+
+          // Ensure that we don't have any invalid `required` booleans lying around.
+          if (
+            isObject(schema[polyType][idx]) &&
+            'required' in (schema[polyType][idx] as SchemaObject) &&
+            typeof (schema[polyType][idx] as SchemaObject).required === 'boolean'
+          ) {
+            delete (schema[polyType][idx] as SchemaObject).required;
```
erunion, line 441: This'll filter out cases of `required: true` being inside of a schema that looks like this:

```js
oneOf: [
  { type: 'string', required: true}
]
```

__tests__/lib/openapi-to-json-schema/required.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            type: 'array',
+            items: {
+              type: 'string',
+            },
+          },
+        },
+      });
+    });
+
+    it('should add a nested `required` boolea into an already existing parent `required` array', () => {
```
darrenyong, line 93: > boolea

typo!

src/lib/openapi-to-json-schema.ts
```diff
@@ -674,6 +683,13 @@ export default function toJSONSchema(
             refLogger,
             transformer,
           });
+
+          // If we have a non-array `required` entry in our `items` schema then it's invalid and we
+          // should remove it. We only support non-array boolean `required` properties inside object
+          // properties.
+          if (isObject(schema.items) && 'required' in schema.items && !Array.isArray(schema.items.required)) {
```
darrenyong, line 690: if I'm understanding this part correctly, I don't think `required` on `items` is something our validator accepts: [bin](https://bin.readme.com/s/64b8283621693d002128f405)

erunion: Ah cool. I'll leave this in here just in case we have some weird cases deep in the DB for it.

__tests__/lib/openapi-to-json-schema/required.test.ts
erunion, line 93:
```suggestion
    it('should add a nested `required` boolean into an already existing parent `required` array', () => {
```

darrenyong (approved):

PR #782, opened 2023-07-24
fix: crash where we didn't support non-standard component types

erunion:
| 🚥 Resolves RM-7605 |
| :---------------- |

## 🧰 Changes

This resolves a problem in our JSON Schema generation where if the API definition in question uses non-standard `components` types, like `#/components/x-definition`, when we attempt to JSON Schema-ify that `x-definition` component block we'd crash because `x-definition` wasn't being pre-populated into our list of known components.


__tests__/__datasets__/percolate-engineering.json
kanadgupta, line 1: wanna rename this file to something like `non-standard-components.json`?

src/operation/get-parameters-as-json-schema.ts
```diff
@@ -220,26 +220,9 @@ export default function getParametersAsJSONSchema(
...
-      responses: {},
-      parameters: {},
-      requestBodies: {},
-      headers: {},
-      securitySchemes: {},
-      links: {},
-      callbacks: {},
+      ...Object.keys(api.components)
+        .map(componentType => ({ [componentType]: {} }))
+        .reduce((prev, next) => Object.assign(prev, next), {}),
```
kanadgupta, line 225: the pros seem to outweigh the cons here but is this gonna muck up intellisense again?

kanadgupta (approved):
a few small comments otherwise LGTM!

src/operation/get-parameters-as-json-schema.ts
erunion: Seems to have had no affect on intellisense speeds. I think the original thing that fixed Intelisense was doing `Partial<ComponentsObject>` and since these things like `x-definitions` doesn't exist in the OAS types it's fine.

kanadgupta (approved):

PR #791, opened 2023-08-18
fix(reducer): retain root-defined security definitions

erunion:
| 🚥 Resolves https://github.com/readmeio/rdme/issues/860 |
| :------------------- |

## 🧰 Changes

If you use the reducer with an API definition that has a root-level `security` configuration we wouldn't retain the `securityScheme` components for those resulting in an invalid OpenAPI definition.


kanadgupta (approved):
👑 you dropped this etc. etc.

PR #793, opened 2023-08-26
feat: added support for cjs and esm

### LINKED ISSUE #774: Support for ESM
### Problem

Right now the `package.json` only contains **main** and **types**. That works most of the time fine but when using native ESM it will complain because inside the `dist/index.js` there is a `require` which is not supported by ESM.

### Solution

- Add "type": "module"
- Add native ESM bundle
- Add exports with support for CJS(as today) and ESM
```
 "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.js", //esm
      "module": "./dist/index.js", //esm
      "default": "./dist/index.cjs" //cjs
    }
  },
```

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm

prisis:
| 🚥 Resolves #774 | 
| :------------------- |

## 🧰 Changes

- Added support for cjs and esm with tsup
- Upgraded the node version from 14 to 16 because of the used eslint version, node 14 cant be supported (eslint-typescript use ??= syntax).

## 🧬 QA & Testing

Check the output for cjs and esm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1213,6 +1213,26 @@
       "integrity": "sha512-+Fj43pSMwJs4KRrH/938Uf+uAELIgVBmQzg/q1YG10djyfA3TnrU8N8XzqCh/okZdszqBQTZf96idMfE5lnwTA==",
       "dev": true
     },
+    "node_modules/@testing-library/dom": {
+      "version": "9.3.1",
+      "resolved": "https://registry.npmjs.org/@testing-library/dom/-/dom-9.3.1.tgz",
+      "integrity": "sha512-0DGPd9AR3+iDTjGoMpxIkAsUihHZ3Ai6CneU6bRRrffXMgzCdlNk43jTrD2/5LT6CBb3MWTP8v510JzYtahD2w==",
+      "dev": true,
+      "peer": true,
+      "dependencies": {
+        "@babel/code-frame": "^7.10.4",
+        "@babel/runtime": "^7.12.5",
+        "@types/aria-query": "^5.0.1",
+        "aria-query": "5.1.3",
+        "chalk": "^4.1.0",
+        "dom-accessibility-api": "^0.5.9",
+        "lz-string": "^1.5.0",
+        "pretty-format": "^27.0.2"
+      },
+      "engines": {
+        "node": ">=14"
+      }
+    },
     "node_modules/@tsconfig/node10": {
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/@tsconfig/node10/-/node10-1.0.9.tgz",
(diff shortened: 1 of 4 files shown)


erunion (approved):
oh wow! thanks so much for this. i'm going to wait for @kanadgupta to take a look at it before merging it in.

tsconfig.json
```diff
@@ -4,10 +4,12 @@
...
     "declaration": true,
     "esModuleInterop": true,
-    "lib": ["dom", "es2020"],
+    "lib": ["DOM", "ES2020"],
     "noImplicitAny": true,
     "outDir": "./dist",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "target": "ES2020",
+    "moduleResolution": "node"
```
kanadgupta, line 12: nit but mind alphabetizing this (i.e., moving the `moduleResolution` field after `lib`)?

package.json
```diff
@@ -31,7 +42,7 @@
     "url": "https://github.com/readmeio/oas/issues"
   },
   "scripts": {
-    "build": "tsc",
+    "build": "cross-env NODE_ENV=production tsup",
```
kanadgupta, line 45: per my comment above:

```suggestion
    "build": "tsup",
```

tsup.config.ts
```diff
@@ -0,0 +1,25 @@
+import type { Options } from 'tsup';
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { defineConfig } from 'tsup';
+
+import tsconfig from './tsconfig.json';
+
+export default defineConfig((options: Options) => ({
+  ...options,
+  entry: ['src/index.ts'],
```
kanadgupta, line 10: not your fault here since we don't really document but this well, but we rely on a few transpiled files from this library over in `rdme` ([example](https://github.com/readmeio/rdme/blob/51a80867c45de15e2b41af0c4bd5bbc61b932804/src/cmds/openapi/convert.ts#L2)). i tried checking out this branch locally and `npm link`-ing but wasn't able to get this working (even with the changes below) so this will require some deeper investigation 😕 is there a way to get `tsup` to transpile without bundling?

```suggestion
  entry: ['src/index.ts', 'src/rmoas.types.ts', 'src/analyzer/index.ts', 'src/lib/reducer.ts'],
```

kanadgupta (changes requested):
Hey thanks for this @prisis! I did some preliminary investigation and noticed a few issues below. Happy to jump in and help this get across the finish line when I'm back in office next week.

package.json
```diff
@@ -4,10 +4,21 @@
   "description": "Comprehensive tooling for working with OpenAPI definitions",
   "license": "MIT",
   "author": "ReadMe <support@readme.io> (https://readme.com)",
+  "sideEffects": false,
```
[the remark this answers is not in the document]
prisis: This is used for https://webpack.js.org/guides/tree-shaking/

package.json
```diff
@@ -4,10 +4,21 @@
...
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.js",
+      "import": "./dist/index.mjs"
+    },
+    "./package.json": "./package.json"
```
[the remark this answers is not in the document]
prisis: A lot of packages are you this export for example preact or nextjs

This can be used to support more then one version of a package


import pkg from "oas/package.json"

pkg.version....

tsup.config.ts
prisis: Will add the operation.ts to the list too, saw it someware

package-lock.json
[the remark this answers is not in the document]
prisis: Sorry, had a old version of npm (using most of the time pnpm :D)

erunion (approved):
amazing, thank you.

erunion:
@stijnvanhulle Sure, want to submit a PR?

PR #807, opened 2023-10-04
feat(oas-extensions): moving `oas-extensions` into this repo

erunion:
## 🧰 Changes

This moves our [oas-extensions](https://github.com/readmeio/oas-extensions) library into the new monorepo structure that I started in https://github.com/readmeio/oas/pull/806.


.github/workflows/ci.yml
```diff
@@ -31,5 +31,4 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: npm ci
-      - run: npm run build
```
erunion, line 34: Moved this into a `prelint` step instead.

packages/oas-extensions/src/index.ts
```diff
@@ -0,0 +1,134 @@
+import type Oas from 'oas';
+import type Operation from 'oas/operation';
```
erunion, line 2: Over in `oas-extensions` this was originally this:

```ts
// eslint-disable-next-line import/no-named-default
import type { default as Oas, Operation } from 'oas';
```

But because `Operation` is no longer being exported there[^1] I had to change this.

[^1]: https://github.com/readmeio/oas/pull/799/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L901

packages/oas-extensions/tsconfig.json
erunion, line 1: We can, and probably should, use the same TSConfig for every package in the monorepo but I'm going to leave that as a followup refactor after all these packages are moved over.

PR #810, opened 2023-10-05
refactor(monorepo): moving the library over to a monorepo structure

erunion:
## 🧰 Changes

We've got a lot of OpenAPI tooling in various separate repositories and as most, if not all, of them depend on `oas` for core functionality it's been a real beast to try to keep them all up to date as we adopt ESM in `oas`. It would be a lot easier if they were all under the same roof so this is the first step towards that.

I'll follow this PR up with more to move other libraries over into this new monorepo structure.


package.json
```diff
@@ -1,121 +1,56 @@
...
-    "remove-undefined-objects": "^4.0.2"
+  "engines": {
+    "node": ">=18"
   },
+  "workspaces": [
+    "./packages/*"
+  ],
   "devDependencies": {
-    "@commitlint/cli": "^17.7.1",
+    "@commitlint/cli": "^17.6.6",
```
kanadgupta, line 29: why did this downgrade?

package.json
```diff
@@ -1,121 +1,56 @@
...
     "@commitlint/config-conventional": "^17.7.0",
-    "@readme/eslint-config": "^13.0.1",
-    "@readme/oas-examples": "^5.12.0",
-    "@readme/openapi-parser": "^2.5.0",
-    "@types/json-schema-merge-allof": "^0.6.1",
-    "@types/memoizee": "^0.4.6",
-    "@types/node": "^20.5.9",
-    "@vitest/coverage-v8": "^0.34.4",
+    "@readme/eslint-config": "^12.2.1",
```
kanadgupta, line 31: why's this v12?

package.json
kanadgupta, line 1: does this file need a `name` attribute?

(fun fact: you can run `npm pkg fix` on `npm@9` or higher as a little package.json linter)

package.json
erunion: not sure but i fixed this in https://github.com/readmeio/oas/pull/811.

package.json
erunion: `npm pkg fix` wants to add `name` and `version` but because it's more of a meta package for Lerna it's not really needed.

kanadgupta (approved):

PR #811, opened 2023-10-05
refactor: migrate oas libraries of ours into this new monorepo

erunion:
## 🧰 Changes

This migrates the following libraries into the new monorepo structure that I've started with https://github.com/readmeio/oas/pull/810:

* [x] [oas-to-har](https://github.com/readmeio/oas-to-har)
* [x] [oas-normalize](https://github.com/readmeio/oas-normalize)
* [x] [oas-extensions](https://github.com/readmeio/oas-extensions)

Others that may be moved later:

* [oas-examples](https://github.com/readmeio/oas-examples): Maybe but I think it's fine as-is. 
* [oas-to-snippet](https://github.com/readmeio/oas-to-snippet): This needs some work to be ESM ready so it'll come over later.
* [openapi-parser](https://github.com/readmeio/openapi-parser): Our fork of [APIDevTools/swagger-parser](https://github.com/APIDevTools/swagger-parser) has diverged from upstream pretty well so I might fully detach ourselves and move that in here so all of our OpenAPI tooling is under one roof. TBD though.


packages/oas-extensions/README.md
```diff
@@ -0,0 +1,15 @@
+# @readme/oas-extensions
+
+A list of extensions that ReadMe has added to extend the OpenAPI Specification.
+
+https://docs.readme.com/docs/openapi-extensions
+
+[![Build](https://github.com/readmeio/oas/workflows/CI/badge.svg)](https://github.com/readmeio/oas/tree/main/packages/oas-extensions)
+
+[![](https://raw.githubusercontent.com/readmeio/.github/main/oss-header.png)](https://readme.io)
```
kanadgupta, line 9: ohhh thank you for this

packages/oas/.npmignore
kanadgupta, line 1: since we're specifying files via that `files` array in the `package.json`, we might not need a lot of this but that can happen in a separate PR.

packages/oas-to-har/tsup.config.ts
kanadgupta, line 1: would love to make a shared `tsup` config at the top level (but again, can happen in a separate PR)

.github/workflows/ci.yml
```diff
@@ -31,4 +31,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: npm ci
+
+      # With the way these libraries use each other we need to run our build step for linting in a
```
kanadgupta, line 35: i think it'd be safer to configure stuff like this via lerna, that way we can run a top-level `npm run build` command. i think something like this? https://lerna.js.org/docs/concepts/task-pipeline-configuration#define-task-dependencies-aka-task-pipelines

kanadgupta (approved):
i have a lot of ideas for how we can consolidate a lot of the config here, but for the sake of this PR, LGTM!

.github/workflows/ci.yml
erunion: It only seems like we're able to say "run <task> before <task>" but only on a global level. I don't see any way with Lerna/NX to set up something that says "run `oas-normalize` build before `oas` build`.

kanadgupta: i might play around with turborepo at some point: https://turbo.build/repo/docs/core-concepts/monorepos/task-dependencies

packages/oas-extensions/README.md
erunion: yeah i was sick of seeing that cloudfront.net link

kanadgupta (approved):

packages/oas-extensions/README.md
kanadgupta: I have [an open design request to update this](https://linear.app/readme-io/issue/RM-7811/refreshed-banner-image-for-usage-across-oss-repos) and i was literally thinking of doing this so we can just update the graphic in one fell swoop

PR #813, opened 2023-10-05
refactor: consolidating our ts and tsup configs

erunion:
## 🧰 Changes

Consolidates our many TS and `tsup` configs across these new packages.

I've also added a root `attw` script that we can run to run [attw](https://arethetypeswrong.github.io/) across every package.

### DIFF
--- a/packages/oas-extensions/tsconfig.json
+++ b/packages/oas-extensions/tsconfig.json
@@ -1,14 +1,8 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "allowJs": true,
     "baseUrl": "./src/",
-    "declaration": true,
-    "esModuleInterop": true,
-    "module": "ESNext",
-    "moduleResolution": "bundler",
-    "noImplicitAny": true,
-    "outDir": "./dist",
-    "target": "ES2020"
+    "outDir": "./dist"
   },
   "include": ["./src/**/*"]
 }
--- a/packages/oas-extensions/tsconfig.json
+++ b/packages/oas-extensions/tsconfig.json
@@ -1,8 +1,4 @@
 {
   "extends": "../../tsconfig.base.json",
-  "compilerOptions": {
-    "baseUrl": "./src/",
-    "outDir": "./dist"
-  },
   "include": ["./src/**/*"]
 }
--- a/packages/oas-extensions/tsconfig.json
+++ b/packages/oas-extensions/tsconfig.json
@@ -1,4 +1,5 @@
 {
   "extends": "../../tsconfig.base.json",
-  "include": ["./src/**/*"]
+  "include": ["./src/**/*"],
+  "exclude": ["dist"]
 }
--- a/packages/oas-normalize/tsconfig.json
+++ b/packages/oas-normalize/tsconfig.json
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     // DOM is needed because `oas-normalize` uses `fetch`.
-    "lib": ["DOM"]
+    "lib": ["DOM"],
   },
-  "include": ["./src/**/*"]
+  "include": ["./src/**/*"],
+  "exclude": ["dist"]
 }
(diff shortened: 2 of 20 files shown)


kanadgupta:
`attw` isn't running in our linting right? I don't see it in our dev-deps or anything!

erunion:
Yeah `attw` isn't running anywhere, just setting it up for us to run manually on our machines right now. Want to build it into the linting step at some point but the Node10 resolution stuff throws a hard exit on the process so it'll be tough to build a status check around.

tsconfig.base.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "declaration": true,
+    "esModuleInterop": true,
+    "lib": ["ES2023"],
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
+    "noImplicitAny": true,
+    "target": "ES2020"
```
kanadgupta, line 10: someday 🥺

```suggestion
    "strict": true,
    "target": "ES2020"
```

kanadgupta:
> but the Node10 resolution stuff throws a hard exit on the process so it'll be tough to build a status check around

yeah I've noticed that and have been thinking of creating a feature request in the ATTW repo

packages/oas/tsconfig.json
```diff
@@ -1,15 +1,9 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "allowJs": true,
     "baseUrl": "./src/",
-    "declaration": true,
-    "esModuleInterop": true,
     "lib": ["ES2023"],
```
kanadgupta, line 5: looks like we already have this in the base config:

```suggestion
```

packages/oas-extensions/tsconfig.json
```diff
@@ -1,14 +1,8 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "allowJs": true,
     "baseUrl": "./src/",
```
kanadgupta, line 4: it looks like all of these have the same `baseUrl`, should we move that to the base config?

packages/oas-normalize/tsconfig.json
```diff
@@ -1,17 +1,12 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "allowJs": true,
     "baseUrl": "./src",
-    "declaration": true,
-    "esModuleInterop": true,
-    "module": "ESNext",
-    "moduleResolution": "bundler",
+    "lib": ["DOM"],
```
kanadgupta, line 5: could we stick with the base `lib` value? why does this library need DOM APIs?

```suggestion
```

packages/oas-to-har/tsconfig.json
```diff
@@ -1,17 +1,10 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "allowJs": true,
     "baseUrl": "./src",
-    "declaration": true,
     "downlevelIteration": true,
-    "esModuleInterop": true,
     "lib": ["DOM", "ES2020"],
```
kanadgupta, line 6: could we stick with the base `lib` value? why does this library need DOM APIs?

```suggestion
```

tsconfig.base.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
```
kanadgupta, line 3: also your call on this but I've been using this as a handy reference lately: https://www.totaltypescript.com/tsconfig-cheat-sheet#building-for-a-library-in-a-monorepo

and it recommends adding the following options in monorepos:

```suggestion
    "allowJs": true,
    "composite": true,
    "sourceMap": true,
    "declarationMap": true
```

packages/oas-normalize/tsconfig.json
erunion: `oas-normalize` uses `fetch` which only exists in `DOM`.

packages/oas-extensions/tsconfig.json
erunion: Moving these to the base config makes TS think that our base URL for the source is `./src` not `packages/*/src`.

kanadgupta: huh, I got it working here: https://github.com/readmeio/api/pull/733/files#diff-0b280a445be167f54cd916c0718c952b8e0d4ca9621903d05eec25efd4c6ed6d

PR #816, opened 2023-10-07
fix(oas): slightly refactoring how our utilities are exported

erunion:
## 🧰 Changes

Doing `export default { .. }` and then loading a specific utility as `import { ... } from 'oas/utils'` gets weird with `tsup`, and it thinks that our utilities don't exist there, so I'm refactoring `oas/utils` to no longer have a default export.


kanadgupta (approved):
probably a breaking change yeah?

PR #819, opened 2023-10-09
fix(oas-extensions): moving `oas` from being a peer to a hard dep

erunion:
## 🧰 Changes

Something I didn't know is that with Lerna if you use the `file:../<dirname>` syntax for loading dependencies in a monorepo, but do that for a peer dependency, Lerna and NPM won't change that `file:` reference when publishing a release. This was causing issues within `oas-extensions` where because `oas` was listed as a peer dependency when I loaded in the latest release of `oas-extensions` it was trying to load `oas` from the same `file:` path in another repository of mine. 

Yikes!

### DIFF
--- a/packages/oas-extensions/package.json
+++ b/packages/oas-extensions/package.json
@@ -39,7 +39,7 @@
     "prepack": "npm run build",
     "test": "vitest run --coverage"
   },
-  "peerDependencies": {
+  "dependencies": {
     "oas": "file:../oas"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/oas-extensions/package.json
```diff
@@ -39,7 +39,7 @@
     "prepack": "npm run build",
     "test": "vitest run --coverage"
   },
-  "peerDependencies": {
```
erunion, line 42: This was originally a peer dep in https://github.com/readmeio/oas-extensions to save me the hassle of bumping it whenever I published a major release of `oas` but because they're all in the same repository now it makes sense to just have this listed as a regular dependency.

PR #820, opened 2023-10-10
fix(oas): use of `typesVersions` for full support of `exports` in TypeScript

stijnvanhulle:
## 🧰 Changes

When using TypeScript with version `23.x.x` it will complain it can not find the subpackages(see `exports` in package.json). So to make it work `typesVersions` needs to be added and then the typecheck will be fine.

https://arethetypeswrong.github.io/?p=oas%4023.0.0

https://nodejs.org/api/packages.html#subpath-imports

> "default" - the generic fallback that always matches. Can be a CommonJS or ES module file. This condition should always come last.

### DIFF
--- a/packages/oas/package.json
+++ b/packages/oas/package.json
@@ -8,38 +8,42 @@
   "type": "module",
   "exports": {
     ".": {
-      "require": "./dist/index.cjs",
-      "import": "./dist/index.js"
+      "import": "./dist/index.js",
+      "module": "./dist/index.js",
+      "default": "./dist/index.cjs"
     },
     "./analyzer": {
-      "require": "./dist/analyzer/index.cjs",
-      "import": "./dist/analyzer/index.js"
+      "import": "./dist/analyzer/index.js",
+      "module": "./dist/analyzer/index.js",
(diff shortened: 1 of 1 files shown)


packages/oas/package.json
```diff
@@ -75,7 +104,6 @@
     "lint:js": "eslint . --ext .js,.ts --ignore-path ../../.gitignore",
     "lint:types": "tsc --noEmit",
     "prebuild": "rm -rf dist/",
-    "prepack": "npm run build",
```
erunion, line 78: Can you revert this, the `oas-normalize` and `gitHead` changes?

erunion (changes requested):

erunion (approved):
fascinating. thanks so much for this. will let @kanadgupta take a look at this before merging it in.

erunion:
Yeah `node` module resolution definitely doesn't support `exports` because `node` maps to Node 10. To be honest I've not really been prioritizing `node` in this ESM + CJS work in our repositories because of that reason. I'm cool to support it if we can get it working with minimal invasiveness to `package.json` though.

PR #826, opened 2023-10-28
feat(oas): refactoring `oas-extensions` completely into `oas` core

erunion:
## 🧰 Changes

Some folks want to be able to control the ordering of parameter groups on their API docs so I am creating a new `x-readme.parameter-ordering` extension that will allow them to do that. However in the creation of that work I ran into an issue where what I was doing in `oas` was going to require pulling `oas-extensions` in and as `oas-extensions` depends on `oas` that would create a nasty circular dependency. Instead of trying to finagle around that I've decided to completely blow away `oas-extensions` and pull its work directly into `oas` where I think it makes more sense to live.

* [x] Complete deprecation of the `@readme/oas-extensions` package.
  * Have also removed this refactored this dependency out of `oas-to-har` and `oas-to-snippet`.
* [x] Updating the API for `Oas#getExtension` and `Oas#hasExtension` to optionally support supplying them an `Operation` instance to see if either that operation or the main API definition has a given extension.
* [x] Creation of `Oas#validateExtension` and `Oas#validateExtensions` for validating one or all of our own custom OpenAPI extensions.
* [x] Creation of a new `x-readme.parameter-ordering` extension.
* [x] Removal of two extensions `x-readme.samples-enabled` and `x-readme.send-defaults`.
  * These two extensions have been deprecated since we rebuilt the API Explorer in 2021 and now that our old API Explorer is no more these deprecated extensions no longer need to exist.
* [x] Addition of lots of documentation around our custom extensions.
* [x] Integration of the new `parameter-ordering` extension into our parameter JSON Schema generation to sort that output according to how the incoming OpenAPI definition or Operation prefers.

## 🎒 `parameter-ordering` extension

Things to know about this new `parameter-ordering` extension:

1. It must contain all of our available parameter types: `path`, `query`, `body`, `cookie`, `form`, and `header`.
2. It's supported at the operation or global root API definition level. Like with the rest of our extension handling we'll prioritize the operation-level definition if present.

```js
{
  "x-readme": {
    "parameter-ordering": ['path', 'query', 'header', 'cookie', 'body', 'form']
  }
}
```


packages/oas-to-har/src/index.ts
```diff
@@ -280,7 +280,7 @@ export default function oasToHar(
   };
 
   if (opts.proxyUrl) {
-    if (getExtension(PROXY_ENABLED, oas, operation)) {
+    if (oas.getExtension(PROXY_ENABLED, operation)) {
```
erunion, line 283: This API existing directly on `Oas` feels a **lot** cleaner.

PR #827, opened 2023-10-31
refactor(oas): removing default exports from a number of areas

erunion:
## 🧰 Changes

We're running into a host of problems integrating our new CJS+ESM dual bundles that this repo offers now in our main app and have been hitting the following error in TS:

> The type of this expression cannot be named without a 'resolution-mode' assertion, which is an unstable feature. Use nightly TypeScript to silence this error. Try updating with 'npm install -D typescript@next'.ts(2841)

The error doesn't make any sense but I was able to discover that it's being thrown because we're in a CJS environment there and parts of `oas` don't have proper CJS exports:

before

This is happening because some of these files have a `default` and named exports in the same file and those aren't compatible with the way CJS libraries get generated from TS. As with most ESM things we've been running into lately this evolved into me losing an afternoon to yakshaving and creating a new ESLint plugin, [eslint-plugin-readme](https://npm.im/eslint-plugin-readme), that ships a new rule [no-dual-exports](https://github.com/readmeio/standards/blob/main/packages/eslint-plugin/docs/no-dual-exports.md) that will flag these dual export cases to us in our `@readme/eslint-config/esm` config.

So what all is this PR? A whoooooole lot.

* [x] The `Operation` class in `oas/operation` is now a named export.
* [x] Renamed `oas/rmoas.types` to `oas/types`.
* [x] The TS types that were in `oas/analyzer` now live in `oas/analyzer/types`.
* [x] The reducer, `oas/lib/reducer`, now lives at `oas/reducer`.
* [x] `getParametersAsJSONSchema` in `operation/get-parameters-as-json-schema` is now a named export.

## 🧬 QA & Testing

After all these refactors `attw` is now happy with my work so I should be able to resume upgrading our `oas` usage in our main app:

after


packages/oas/test/operation/__snapshots__/index.test.ts.snap
kanadgupta, line 1: can't seem to deduce out why there are several new snapshot files but none being deleted 🤔

kanadgupta (approved):

PR #833, opened 2023-11-01
refactor: update getTags() function to handle webhooks

darrenyong:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes
I've updated the `.getTags()` function to be able to handle webhooks as well in preparation for our webhooks support. This makes it so that any webhooks present in the file will have their tags correctly generated and put in the correct category. Prior to this, it was just placing the endpoint pages in random places. 

This particular test will fail until the [updated](https://github.com/readmeio/oas-examples/pull/90) `webhooks.yaml` file on the `oas-examples` repo is pulled in.

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -758,6 +758,20 @@ export default class Oas {
       });
     });
 
+    Object.entries(this.getWebhooks()).forEach(([path, webhooks]) => {
+      Object.values(webhooks).forEach(webhook => {
+        const tags = webhook.getTags();
+        if (setIfMissing && !tags.length) {
+          allTags.add(path);
+          return;
+        }
+
+        tags.forEach(tag => {
+          allTags.add(tag.name);
+        });
+      });
+    });
+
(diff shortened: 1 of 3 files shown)


packages/oas/src/index.ts
```diff
@@ -758,6 +758,20 @@ export default class Oas {
       });
     });
 
+    Object.entries(this.getWebhooks()).forEach(([path, webhooks]) => {
```
erunion, line 761: Can you remove the "this method doesn't support webhooks` comment in the docblock of this method?

packages/oas/test/index.test.ts
erunion, line 1: I bumped `@readme/oas-examples` with your webhook example tag change to `5.12.1`.

erunion (changes requested):

packages/oas/src/index.ts
darrenyong: done!

PR #834, opened 2023-11-02
feat(oas-to-snippet): plugin architecture

erunion:
## 🧰 Changes

This pulls in the new plugin architecture I added to `@readme/httpsnippet` in https://github.com/readmeio/httpsnippet/pull/206 so we can now do a couple great things:

* [x] Decouple [httpsnippet-client-api](https://npm.im/httpsnippet-client-api) completely from this library. `httpsnippet-client-api` depending on `oas` and being used here has historically made bumping major `oas` releases a lot more problematic than it should be.
  * I'm going to treat this as a breaking change since `api` snippets will no longer be natively supported by this library without the `httpsnippet-client-api` plugin for it.
* [x] Allow for the dynamic installation of any new HTTPSnippet-compatible plugin, thereby extending our code snippet generation to custom externally developed language clients.
  * For example a certain unnamed finance company could create a new package on NPM `httpsnippet-plugin-companyname` and we could generate snippets with a new `[node, companyname]` language 🧙 like magic 🪄 

I've also refactored our `/supportedLanguages` export to now be a `getSupportedLanguages` function we'll call to retrieve the languages we support. Had to make this change in order to load custom plugins into the supported languages configuration and make them be available as a supported language. 🐔 🥚 

I don't have any docs written on the plugin system but will write those at some point before this work goes live in our API Explorer.


packages/oas-to-snippet/README.md
```diff
@@ -24,41 +24,62 @@ import petstore from './petstore.json';
 const apiDefinition = new Oas(petstore);
 const operation = apiDefinition.operation('/pets', 'get');
 
-// This is a keyed object containing formData for your operation. Available keys are: path,
```
erunion, line 27: Trimmed these Markdown code snippet lines down to 100 character lengths because 120 looks like shit on small screens on NPM and Github.

packages/oas-to-snippet/src/supportedLanguages.ts
```diff
@@ -1,261 +0,0 @@
...
-  csharp: {
-    highlight: 'text/x-csharp',
-    httpsnippet: {
-      lang: 'csharp',
-      default: 'restsharp',
-      targets: {
-        httpclient: { name: 'HttpClient' },
-        restsharp: {
-          name: 'RestSharp',
-          install: 'dotnet add package RestSharp',
```
erunion, line 43: This `supportedLanguages` config now lives at the `DEFAULT_LANGAUGES` config in `languages.ts` and is largely the same except for two changes:

* `api` is no longer listed here under `node` because that's been offloaded to our `httpsnippet-client-api` plugin.
* `install` has moved directly into `@readme/httpsnippet` client configurations so that information is as close to the snippet generator as possible.

packages/oas-to-snippet/src/languages.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  return languages;
+}
+
+export function getLanguageConfig(languages: SupportedLanguages, lang: Language) {
+  let config: LanguageConfig;
+  let language: TargetId;
+  let target: ClientId;
+
+  switch (lang) {
```
darrenyong, line 303: this looks so much cleaner than that big if statement!

packages/oas-to-snippet/src/index.ts
```diff
@@ -45,20 +46,39 @@ export default async function oasToSnippet(
        */
       variableName?: string;
     };
+
+    /**
+     * `httpsnippet` plugins to extend snippet generation to.
+     *
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    plugins?: ClientPlugin<Record<string, any>>[];
```
kanadgupta, line 55:
```suggestion
    plugins?: ClientPlugin<Record<string, unknown>>[];
```

packages/oas-to-snippet/src/languages.ts
```diff
@@ -0,0 +1,367 @@
...
+      target = 'curl';
+      break;
+
+    // Our `api` snippet plugin, `httpsnippet-client-api`, that we use in ReadMe used to be called
+    // `node-simple`. For backwards compatibility we're retaining this old identifier here.
+    case 'node-simple':
+      config = languages.node;
+      language = 'node';
+      target = 'api';
+      break;
```
kanadgupta, line 316: noooo 😭 wait since this will be a breaking change anyways, can we remove this?

packages/oas-to-snippet/README.md
```diff
@@ -24,41 +24,62 @@ import petstore from './petstore.json';
...
-// highlighting (with @readme/syntax-highlighter, for example).
-const { code, highlightMode } = oasToSnippet(apiDefinition, operation, formData, auth, language, { openapi: { registryIdentifier } });
+```js
+import oasToSnippet from '@readme/oas-to-snippet';
+import httpsnippetClientAPIPlugin from 'httpsnippet-client-api';
+
+const snippet = await oasToSnippet(
+  petstore,
+  petstore.operation('/user/login', 'get'),
+  formData,
```
kanadgupta, line 66: wanna add some example usage/comments here that indicate how these variables are constructed?

packages/oas-to-snippet/README.md
```diff
@@ -24,41 +24,62 @@ import petstore from './petstore.json';
...
+    },
+    plugins: [httpsnippetClientAPIPlugin],
+  },
+);
 ```
 
 ## Supported Languages
 
-Since this library uses [HTTP Snippet](https://github.com/Kong/httpsnippet), we support most of its languages, and their associated targets, which are the following:
+Since this library uses [HTTPSnippet](https://npm.im/@readme/httpsnippet) we support most of its languages and their associated targets which are the following:
```
kanadgupta, line 82: :hell-yeah:

packages/oas-to-snippet/src/index.ts
```diff
@@ -68,21 +88,18 @@ export default async function oasToSnippet(
...
-    targetOpts.identifier = opts?.openapi?.variableName;
-  }
+    // Our `httpsnippet-client-api` plugin uses these options so we need to pass them along.
+    if (plugin.target === 'node' && plugin.client.info.key === 'api') {
+      targetOpts.api = {
+        definition: oas ? oas.getDefinition() : null,
+        identifier: opts?.openapi?.variableName,
+        registryURI: opts?.openapi?.registryIdentifier,
+      };
+    }
```
kanadgupta, line 101: maybe i'm misunderstanding how all of this fits together, but i feel like this logic (and its corresponding opts) should be in the plugin code, not here

packages/oas-to-snippet/src/languages.ts
```diff
@@ -0,0 +1,367 @@
...
+  lang: Language,
+  /**
+   * For `api` snippet generation we need supply a ReadMe API Registry identifier that'll be used
+   * to retrieve the OpenAPI definition to generate their SDK through the `npx api install` command.
+   * These users won't be installing the `api` package itself to run the snippets we generate for
+   * them.
+   *
+   * @example @developers/v2.0#17273l2glm9fq4l5
+   */
+  registryIdentifier?: string,
```
kanadgupta, line 357: i imagine other plugins might want custom installation steps similar to `api` does, could we make this (and its corresponding JSDocs) more generic (i.e., call it `packageName`)?

kanadgupta (commented):
a few initial questions, i might set up a proper sandbox environment to test this out with @darrenyong later today!

packages/oas-to-snippet/README.md
kanadgupta: fwiw i don't think this is a huge deal since both npm and github support side-scrolling on these code samples and even 100 characters is still too wide for certain screens like mobile etc.

packages/oas-to-snippet/src/languages.ts
erunion: there might still be specs in the wild with `node-simple` in the `x-samples-languages` extension.

packages/oas-to-snippet/README.md
erunion: These exist above in the readme.

packages/oas-to-snippet/src/index.ts
erunion: it is, but these options need to get passed along from here into the plugin and the plugin wants these options as an `api` object. im not really sure on the best way to handle a plugin that has its own options like this other than this thing i've got here.

packages/oas-to-snippet/src/languages.ts
erunion: this work is a little confusing but the `registryIdentifier` here is the `npx api install <registry identifier>` part of things and `variableName` is the eventual post-installation package name.

packages/oas-to-snippet/src/languages.ts
erunion: actually looks like we have code in the main app to remap `node-simple` to `[node, api]`. going to remove this then.

packages/oas-to-snippet/src/languages.ts
erunion: could maybe have the config for the plugin supplied alongside the plugin but that's going to require a **lot** of work to httpsnippet as it doesn't support options like this

```js
const snippet = await oasToSnippet(
  petstore,
  petstore.operation('/user/login', 'get'),
  formData,
  auth,
  // `[node, api]` is not a standard language in `oas-to-snippet` but is
  // dynamically made available via this loaded plugin.
  ['node', 'api'],
  {
    plugins: [httpsnippetClientAPIPlugin({ api: { registryIdentifier: '@petstore/v2.0#17273l2glm9fq4l5', }})],
  },
);
```

packages/oas-to-snippet/src/index.ts
erunion: Have to use `any` here.

Screen Shot 2023-11-03 at 10 03 00 AM

packages/oas-to-snippet/src/languages.ts
erunion: also removing our backcompat for `curl` being rewritten to `shell` as we have that same logic upstream in our explorer

packages/oas-to-snippet/README.md
kanadgupta: woof my bad 😵‍💫

packages/oas-to-snippet/README.md
```diff
@@ -24,41 +24,62 @@ import petstore from './petstore.json';
...
 const language = 'node';
 
-// This `registryIdentifier` option is only necessary when using the `['node', 'api']` language
-// combination. It controls how these snippets are used according to the code generation tooling
-// that we offer with https://api.readme.dev/.
-const registryIdentifier: '@petstore/v2.0#17273l2glm9fq4l5';
+// This will return an object containing `code` and `highlightMode`. `code` is
+// the generated code snippet, while `highlightMode` is the language mode you
+// can use to render it for syntax highlighting (with `codemirror` for example).
+const { code, highlightMode } = oasToSnippet(apiDefinition, operation, formData, auth, language });
```
kanadgupta, line 52:
```suggestion
const { code, highlightMode } = await oasToSnippet(apiDefinition, operation, formData, auth, language);
```

kanadgupta (approved):

PR #836, opened 2023-11-06
feat: add getSummary and getDescription methods

darrenyong:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes
Add `getSummary()` and `getDescription()` methods to Webhook class.

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -724,7 +724,7 @@ export default class Oas {
     Object.keys(api.webhooks ? api.webhooks : []).forEach(id => {
       webhooks[id] = {} as Record<RMOAS.HttpMethods, Webhook>;
       Object.keys(api.webhooks[id]).forEach((method: RMOAS.HttpMethods) => {
-        webhooks[id][method] = this.operation(id, method, { isWebhook: true });
+        webhooks[id][method] = this.operation(id, method, { isWebhook: true }) as Webhook;
       });
     });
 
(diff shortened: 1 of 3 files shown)


packages/oas/src/index.ts
```diff
@@ -724,7 +724,7 @@ export default class Oas {
     Object.keys(api.webhooks ? api.webhooks : []).forEach(id => {
       webhooks[id] = {} as Record<RMOAS.HttpMethods, Webhook>;
       Object.keys(api.webhooks[id]).forEach((method: RMOAS.HttpMethods) => {
-        webhooks[id][method] = this.operation(id, method, { isWebhook: true });
+        webhooks[id][method] = this.operation(id, method, { isWebhook: true }) as Webhook;
```
darrenyong, line 727: TypeScript was upset that we were assigning an `Operation` class to a `Webhook` class so added this since it should always be a `Webhook` and never an `Operation` in this function.

packages/oas/src/operation/index.ts
```diff
@@ -857,4 +857,32 @@ export class Callback extends Operation {
   }
 }
 
-export class Webhook extends Operation {}
+export class Webhook extends Operation {
+  /**
+   * OpenAPI API Definition that this webhook originated from.
+   */
+  declare api: RMOAS.OAS31Document;
```
darrenyong, line 864: TypeScript threw an error here saying that the `api` property was going to overwrite the base `api` property in `Operation` so I had to add a `declare` here to mark it intentional.

erunion: TIL

erunion (approved):

PR #842, opened 2024-01-09
feat: add support for ordering via `tags` array

darrenyong:
| 🚥 Resolves CX-514 |
| :------------------- |

## 🧰 Changes
These changes adds support for ordering pages by the `tags` array as defined in OAS spec.

Per [OAS Spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#fixed-fields):
`The order of the tags can be used to reflect on their order by the parsing tools.`

The tags on `this.api.tags` comes in in the correct order, however, we create a new array `Array.from(allTags)` with tags added as we parse the file from top down. With these changes, I sort this array against the correctly ordered `oasTags`. Since the sorting will put tags that exist only at the endpoint level to the front, I also separate them and tack them on to the end of the sorted tags.

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -743,6 +743,11 @@ export default class Oas {
   getTags(setIfMissing = false) {
     const allTags = new Set<string>();
 
+    const oasTags =
+      this.api.tags?.map(tag => {
+        return tag.name;
+      }) || [];
+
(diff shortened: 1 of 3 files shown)


erunion (approved):
nice

PR #847, opened 2024-03-11
chore(deps): lodash security vuln

kanadgupta:
| 🚥 Resolves(?) https://github.com/readmeio/oas/security/dependabot/14 |
| :------------------- |

## 🧰 Changes

Swaps out our `lodash.get` and `lodash.set` packages for the more modern ES equivalents to hopefully take care of https://github.com/readmeio/oas/security/dependabot/14

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/packages/oas-to-har/package.json
+++ b/packages/oas-to-har/package.json
@@ -50,8 +50,7 @@
   },
   "dependencies": {
     "@readme/data-urls": "^3.0.0",
-    "lodash.get": "^4.4.2",
-    "lodash.set": "^4.3.2",
+    "lodash-es": "^4.17.21",
     "oas": "file:../oas",
     "qs": "^6.11.2",
     "remove-undefined-objects": "^5.0.0"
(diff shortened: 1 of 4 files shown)


packages/oas-to-har/package.json
```diff
@@ -50,17 +50,15 @@
   },
   "dependencies": {
     "@readme/data-urls": "^3.0.0",
-    "lodash.get": "^4.4.2",
-    "lodash.set": "^4.3.2",
+    "lodash-es": "^4.17.21",
```
erunion, line 53: didn't we have issues with this package here in the past where it **only** supports ESM?

kanadgupta: (sigh) I can't say I remember for sure... I think https://github.com/readmeio/oas-to-har/commit/a3421796c3a90fee3d8720c26c4f03df79cdaacb was actually the issue?

kanadgupta: worst-case scenario we use `tsup` for `oas-to-har` so we could probably bundle these directly if it ends up being a problem

erunion (approved):

PR #851, opened 2024-04-02
fix(oas): numerous tweaks to our camelCase `operationId` generation

erunion:
## 🧰 Changes

Talking with @julshotal this afternoon I stumbled upon some oversight with our Git project planning where I wasn't sure what we'd use for endpoint pages we're storing in Git, either alternating between using `title` (when present) or the `camelCase` option on `Operation.getOperationId()`. Long story short working it out with her I discovered a bug in this method where we weren't camelcasing underscores in auto-generated IDs (like `postAc_eq_hazard180` -> `postAcEqHazard180`). Frustratingly enough I had a test that was testing for this very specific thing, the test just wasn't asserting the right expectation.

Additionally while working on a fix for this I'm also proposing another small tweak to this same method to no longer generate operation IDs with consecutive words. This issue is admittedly pretty rare but when it does happen we end up with some bunk-ass IDs like `GET /pet/{pet_id}` being transformed into `getPetPetId`. The new work I've put in here will now filter out consecutive words and generate `getPetId` in this case. Still not the _best_ operation ID for this endpoint because it's not returning a pet ID but it at least doesn't look like total ass.

### DIFF
--- a/packages/oas/test/operation/index.test.ts
+++ b/packages/oas/test/operation/index.test.ts
@@ -1072,7 +1072,7 @@ describe('#getOperationId()', () => {
       });
 
       const operation = spec.operation('/ac_eq_hazard/18.0', 'post');
-      expect(operation.getOperationId({ camelCase: true })).toBe('postAc_eq_hazard180');
+      expect(operation.getOperationId({ camelCase: true })).toBe('postAcEqHazard180');
     });
 
     it('should not double up on a method prefix if the path starts with the method', () => {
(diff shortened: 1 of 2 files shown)


packages/oas/test/operation/index.test.ts
```diff
@@ -1072,7 +1072,7 @@ describe('#getOperationId()', () => {
       });
 
       const operation = spec.operation('/ac_eq_hazard/18.0', 'post');
-      expect(operation.getOperationId({ camelCase: true })).toBe('postAc_eq_hazard180');
+      expect(operation.getOperationId({ camelCase: true })).toBe('postAcEqHazard180');
```
erunion, line 1075: In our new Git world `postAcEqHazard180.md` will end up being the filename for this endpoint. It's still not great but is at least a bit cleaner than `postAc_eq_hazard180.md`

julshotal (approved):

kanadgupta (approved):

packages/oas/test/operation/index.test.ts
```diff
@@ -1110,7 +1110,25 @@ describe('#getOperationId()', () => {
       });
 
       const operation = spec.operation('/candidate/{candidate_id}/', 'get');
-      expect(operation.getOperationId({ camelCase: true })).toBe('getCandidateCandidate_id');
+      expect(operation.getOperationId({ camelCase: true })).toBe('getCandidateId');
+    });
+
+    it('should not create an operationId that includes the same word in a consequtive sequence', () => {
```
domharrington, line 1116:
```suggestion
    it('should not create an operationId that includes the same word in a consecutive sequence', () => {
```

domharrington (approved):

erunion:
Going to hold off on merging this in until after our OpenAPI tooling dependencies are upgraded in our main app.

kanadgupta (approved):

PR #860, opened 2024-04-10
feat(oas-to-snippet): surface installation instructions with snippets

erunion:
## 🧰 Changes

I don't know why I thought this was a good idea but having to manually invoke `getClientInstallationInstructions()` to pull back installation instructions for a client you're generating a snippet for makes absolutely no sense. This modifies the returned object from `oas-to-snippet` to now include an `install` property that includes this data.

### DIFF
--- a/packages/oas-to-snippet/src/index.ts
+++ b/packages/oas-to-snippet/src/index.ts
@@ -8,7 +8,7 @@ import type { Operation } from 'oas/operation';
 import { HTTPSnippet, addClientPlugin } from '@readme/httpsnippet';
 import generateHar from '@readme/oas-to-har';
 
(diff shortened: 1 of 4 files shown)


packages/oas-to-snippet/src/index.ts
```diff
@@ -53,7 +53,32 @@ export default function oasToSnippet(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     plugins?: ClientPlugin<any>[];
   } = {},
-): { code: string | false; highlightMode: string | false } {
+): {
```
kanadgupta, line 56: snuck in some JSDocs here, let me know what you think

kanadgupta (approved):
small refactoring suggestion below otherwise LGTM! feel free to tweak those JSDocs to your liking

PR #861, opened 2024-04-11
fix(oas-to-snippet): use correct val for registry identifier

kanadgupta:
oops lol (my fault too, i should have looked more closely at the snapshots)


erunion (approved):
how embarassing

PR #865, opened 2024-04-24
ci: dynamic args for `actions/setup-node`

kanadgupta:
Node 22 just dropped — adding some funkier syntax so we can dynamically support all the Node releases we want to. Added some comments with links to resources on the subject!


erunion (approved):
😎

PR #866, opened 2024-04-26
feat(oas): example groups accessor

kanadgupta:
## 🧰 Changes

We've had new UI for **example groups** (f.k.a. [request/response pairs](https://docs.readme.com/main/changelog/pairing-requests-and-responses)) for a bit now, but the way we currently determine these groups was strewn about a few React components in our codebase and the entire setup has already become a bit of a mess to follow (and probably isn't performant at all). In anticipation of expanding its support for native OAS and all of the different parameters that users can define examples for, I decided that bulk of the processing logic should live in the `oas` library, where we have caching and stricter types and tests.

The bulk of this PR is related to a new accessor on the `Operation` model called `getExampleGroups()` — this returns a cleanly formatted object that returns all of the necessary information for these example groups. For a given example group, here's all the information contained in this object:
- Custom code sample(s) — this is a list because we might have several code samples in several different languages for the current example
- The OAS request example data, as formatted for usage in `oas-to-har`
- Information about the corresponding OAS response example
- The name of the example group that's going to be surfaced in the tabbed UI

The logic for generating this object supports a few new things:
- Supports examples defined in the `parameters` object (e.g., cookie/header/path/query parameters)
- Properly handles `form-urlencoded` data the way our explorer receives it
- Properly handles the precedence of custom code samples vs. native OAS example definitions (the former take precedence)
- Ensures that native OAS examples definitions are only present if the request keys match the response key

There were a few tiny cleanups and changes I made as part of this:
- [x] Took the `DataForHAR` type that was originally defined in `oas-to-har`, moved it to `oas`, and re-exported it in `oas-to-har`. `oas-to-har` already depends on `oas` and this was the best way I could think of to utilize that type in both libraries.
- [x] Tightened up the typing on our `ParameterObject` and `RequestBodyExamples` types
- [x] Better JSDocs on our custom code samples extension
- [x] Bumped our `oas-examples` to the latest (still need to get https://github.com/readmeio/oas-examples/pull/106 reviewed)

## 🧬 QA & Testing

I added quite a few tests that should give us good coverage on this front. If you want an idea of what this new `ExamplePairs` object looks like, peep at the new snapshots.


packages/oas/src/operation/lib/get-example-groups.ts
```diff
@@ -0,0 +1,207 @@
...
+ * For standard OAS request parameter (e.g., body/header/query/path/etc.) examples,
+ * they are only present in the return object if they have a corresponding response example
+ * (i.e., a response example with the same key in the `examples` map).
+ */
+export function getExampleGroups(operation: Operation): ExampleGroups {
+  let namelessCodeSamples = 0;
+  const groups: ExampleGroups = {};
+
+  // add custom code samples
+  const codeSamples = getExtension('code-samples', operation.api, operation) as Extensions['code-samples'][];
```
erunion, line 128: Something for a future improvement but it'd be rad if `getExtension` could return the type based on the extension you specified so you could get rid of `as Extensions['code-samples'][]` here.

Might be able to accomplish it with type overrides? Would be neat.

packages/oas/src/operation/lib/get-example-groups.ts
```diff
@@ -0,0 +1,207 @@
...
+        name,
+        customCodeSamples: [sample],
+      };
+    }
+  });
+
+  // if we added any custom code samples, add the corresponding response examples and return
+  if (Object.keys(groups).length) {
+    addMatchingResponseExamples(groups, operation);
+    return groups;
```
erunion, line 157: how come we return here?

packages/oas/src/types.ts
```diff
@@ -112,7 +130,9 @@ export type OperationObject = (OpenAPIV3_1.OperationObject | OpenAPIV3.Operation
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameterObject}
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameterObject}
  */
-export type ParameterObject = OpenAPIV3_1.ParameterObject | OpenAPIV3.ParameterObject;
+export type ParameterObject = {
+  in: 'cookie' | 'header' | 'path' | 'query';
```
erunion, line 134: Do the param objects we're using here not have `in` already/

kanadgupta: sort of, but it's typed as a string 😔

CleanShot 2024-04-29 at 11 23 45@2x

packages/oas/src/operation/lib/get-example-groups.ts
kanadgupta: if we have any custom code samples, we don't also want to add native OAS request examples. the custom code samples take precedence

packages/oas/src/types.ts
erunion: oh BROTHER

PR #868, opened 2024-04-30
ci: vitest workspaces

kanadgupta:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

Messing around with [Vitest Workspaces](https://vitest.dev/guide/workspace.html) and it seems pretty great. Previously, we had to `cd` into the correct subpackage in order to run `npx vitest [path to file]` but now we can just do it from the root 😌 the downside here is that in `oas-normalize` there's one test that will now fail unless it's being run from the project root. Feels like a worthy trade-off, but let me know if you disagree.

As part of this, I also upgraded our Vitest deps to the latest.

## 🧬 QA & Testing

If tests (which should be faster now?) pass, we should be in good shape!

### DIFF
--- a/packages/oas-normalize/test/__fixtures__/bundle/definition.json
+++ b/packages/oas-normalize/test/__fixtures__/bundle/definition.json
@@ -29,12 +29,12 @@
         "content": {
           "application/json": {
             "schema": {
-              "$ref": "./test/__fixtures__/bundle/pet.json"
+              "$ref": "./packages/oas-normalize/test/__fixtures__/bundle/pet.json"
             }
           },
           "application/xml": {
             "schema": {
-              "$ref": "./test/__fixtures__/bundle/pet.json"
+              "$ref": "./packages/oas-normalize/test/__fixtures__/bundle/pet.json"
             }
           }
         },
--- a/vitest.workspace.ts
+++ b/vitest.workspace.ts
@@ -1,0 +1,1 @@
+export default ['packages/*'];
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,8 @@
         "alex": "^11.0.1",
         "eslint": "^8.57.0",
         "lerna": "^8.1.2",
-        "prettier": "^3.2.5"
+        "prettier": "^3.2.5",
+        "vitest": "^1.5.3"
       },
       "engines": {
         "node": ">=18"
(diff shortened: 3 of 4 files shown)


packages/oas-normalize/test/__fixtures__/bundle/definition.json
```diff
@@ -29,12 +29,12 @@
         "content": {
           "application/json": {
             "schema": {
-              "$ref": "./test/__fixtures__/bundle/pet.json"
+              "$ref": "./packages/oas-normalize/test/__fixtures__/bundle/pet.json"
```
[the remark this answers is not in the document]
kanadgupta: i mentioned this in the PR description — not sure which is better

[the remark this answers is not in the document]
erunion: can probably change this file from `definition.json` to `definition.js` and have it export the def + change this ref to use `require.resolve('oas-normalize/test/__fixtures__/bundle/pet.json')`. wonder if that'd work. it wouldn't be a relative import anymore but i don't think that matters too much since relative file refs are tested in `json-schema-ref-parse` and `openapi-parser` anyways.

erunion (approved):
cool with this as-is tho

kanadgupta:
@erunion mind taking another look? tried out a new approach, see this chunk of the updated PR description:

> I forked a test in `oas-normalize` into its own config + test file and made it so it uses `chdir` as part of the test. That way, we can run tests from the both the root and from the subpackages and it'll ✨ just work ✨

vitest.workspace.ts
erunion, line 1: should `oas-normalize` have its own workspace config so if you run `vitest` from inside that package it'll run the chdir test too?

kanadgupta: oooh i hadn't thought of that - let me play around!

kanadgupta: added in https://github.com/readmeio/oas/pull/868/commits/eebc72b4f336584b1968b0dfd89fbd1a11b360d4 — works like a charm ✨

kanadgupta: (i dislike how we have four config files now, but i can't seem to think of a simpler approach here that addresses everything)

erunion (approved):

PR #875, opened 2024-05-10
ci: simplify vitest setup

kanadgupta:
## 🧰 Changes

I noticed that Vitest is going to switching its default `pool` setting to `forks` in the upcoming major release (see https://github.com/vitest-dev/vitest/pull/5047), and realized that we could very easily just switch to that setting in `oas-normalize`, which vastly simplifies (and even speeds up, per my testing) our test setup.

## 🧬 QA & Testing

Are all 1175 tests still running and passing?


erunion (approved):
:sickos:

PR #879, opened 2024-06-11
feat(oas): creation of a new `friendlyCase` option for `.getOperationId()`

erunion:
| 🚥 Resolves https://github.com/readmeio/oas/pull/869 |
| :------------------- |

## 🧰 Changes

This resurrects the work I originally did in https://github.com/readmeio/oas/pull/851, but unfortunately had to revert, to create a new `friendlyCase` option on `Operation.getOperationId()` for generating friendlier camelCase operation IDs that we can use in our upcoming Git storage system for API references. 

## 🧬 QA & Testing

I reorganized a bunch of the tests for `camelCase` to retain them all along with this new option.

### DIFF
--- a/packages/oas/test/operation/index.test.ts
+++ b/packages/oas/test/operation/index.test.ts
@@ -6,6 +6,7 @@ import { beforeAll, describe, it, expect } from 'vitest';
 
 import Oas from '../../src/index.js';
 import { Operation, Callback } from '../../src/operation/index.js';
+import { createOasForPaths } from '../__fixtures__/create-oas.js';
 
 let petstore: Oas;
 let callbackSchema: Oas;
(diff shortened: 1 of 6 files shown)


packages/oas/test/operation/index.test.ts
```diff
@@ -1051,127 +1052,184 @@ describe('#getOperationId()', () => {
...
+        expect(operation.getOperationId({ [option]: true })).toBe(expected);
       });
+    });
 
-      const operation = spec.operation('/anything', 'get');
-      expect(operation.getOperationId({ camelCase: true })).toBe(expected);
+    describe('given an operationId that begins with a number', () => {
+      it.each([
+        ['camelCase', '_400oD_Browse_by_Date_Feed'],
+        ['friendlyCase', '_400oDBrowseByDateFeed'],
```
darrenyong, line 1204: just curious, is there a reason we're preceding these operationId's with an underscore?

erunion: These are eventually used as method names in `api` and JS methods can't begin with a number.

darrenyong (approved):

packages/oas/test/operation/index.test.ts
darrenyong: ah, got it!

PR #888, opened 2024-07-10
fix(oas/json schema): add default resolver for unknown keywords

kanadgupta:
| 🚥 Resolves RM-10218 |
| :------------------- |

## 🧰 Changes

For folks that use custom keywords in their polymorphic objects, we now specify a resolution strategy (i.e., simply select the first value) as opposed to totally ignoring the object.

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -322,19 +322,11 @@ export function toJSONSchema(data: RMOAS.SchemaObject | boolean, opts: toJSONSch
               return arr;
             },
 
-            // JSON Schema ony supports examples with the `examples` property, since we're
-            // ingesting OpenAPI definitions we need to add a custom resolver for its `example`
-            // property.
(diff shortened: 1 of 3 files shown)


packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -322,19 +322,11 @@ export function toJSONSchema(data: RMOAS.SchemaObject | boolean, opts: toJSONSch
               return arr;
             },
 
-            // JSON Schema ony supports examples with the `examples` property, since we're
-            // ingesting OpenAPI definitions we need to add a custom resolver for its `example`
-            // property.
-            example: (obj: unknown[]) => obj[0],
```
kanadgupta, line 328: we no longer need to specify these keywords since they all the did the same thing as our `defaultResolver` 🎊

erunion (approved):

darrenyong (approved):
well that was easy!

PR #889, opened 2024-07-22
fix(oas): don't add defaults to enum param descriptions

kanadgupta:
| 🚥 Resolves CX-947 |
| :------------------- |

## 🧰 Changes

With our new "Defaults to" copy that we're surfacing for request parameters, we now have duplicated info for parameters that are both enums and contain a default value ([bin site preview is here](https://bin.readme.com/s/669ea01aef9101006ed22937)):

CleanShot 2024-07-22 at 13 16 33@2x

This PR removes that description-updating logic from our `openapi-to-json-schema` handling.

## 🧬 QA & Testing

Is this the right spot to make this change? This logic is only for request parameters right?

### DIFF
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -791,23 +791,16 @@ export function toJSONSchema(data: RMOAS.SchemaObject | boolean, opts: toJSONSch
       // Defaults for `object` and types have been dereferenced into their children schemas already
       // above so we don't need to preserve this default anymore.
       delete schema.default;
+    } else if (
+      ('allowEmptyValue' in schema && schema.allowEmptyValue && schema.default === '') ||
+      schema.default !== ''
+    ) {
+      // If we have `allowEmptyValue` present, and the default is actually an empty string, let it
+      // through as it's allowed.
     } else {
-      // If it's an enum and not the response schema, add the default to the description.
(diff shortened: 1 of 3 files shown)


packages/oas/src/lib/openapi-to-json-schema.ts
kanadgupta, line 1: would recommend hiding whitespace, ESLint really did a number on this

packages/oas/test/lib/openapi-to-json-schema.test.ts
```diff
@@ -908,13 +908,13 @@ describe('`description` support', () => {
           type: 'string',
           enum: ['CP', 'NON_CP'],
           default: 'NON_CP',
-          description: 'Enumed property with a default: `NON_CP`\n\nDefault: `NON_CP`',
+          description: 'Enumed property with a default: `NON_CP`\n',
```
erunion, line 911: where's that `\n` coming from now?

kanadgupta: it's hardcoded here: https://github.com/readmeio/oas/blob/74fd996a30ab62103335d8342949a2086f3db191/packages/oas/test/lib/openapi-to-json-schema.test.ts#L881

i'm guessing it was there to test the whitespace handling?

erunion: huh yeah maybe

erunion (approved):

PR #893, opened 2024-08-30
feat(operation/getSecurityWithTypes): expose required scopes

kanadgupta:
## 🧰 Changes

In our `operation.getSecurityWithTypes()` method, this adds/documents a new property called `_requirements` that contains the required scopes for the operation + security scheme[^1].

I also exported our `SecurityType` type and added a JSDoc for it.

## 🧬 QA & Testing

Should be a minor additive change. Do tests/snapshots pass?

[^1]: [read more about the security requirement object here](https://spec.openapis.org/oas/v3.1.0#security-requirement-object)

### DIFF
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -223,6 +223,7 @@ export class Operation {
           security: {
             ...security,
             _key: key,
+            _requirements: requirement[key],
           },
         };
       });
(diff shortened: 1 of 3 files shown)


packages/oas/src/operation/index.ts
```diff
@@ -252,8 +252,12 @@ export class Operation {
           if (!prev[security.type]) prev[security.type] = [];
 
           // Only add schemes we haven't seen yet.
-          const exists = prev[security.type].findIndex(sec => sec._key === security.security._key);
-          if (exists < 0) {
+          const exists = prev[security.type].some(sec => sec._key === security.security._key);
+          if (!exists) {
```
kanadgupta, line 256: this just moves us from `Array.findIndex` to `Array.some` for readability purposes

erunion (approved):

PR #898, opened 2024-09-04
feat: add Operation.isWebhook()

darrenyong:
| 🚥 Resolves N/A |
| :------------------- |

## 🧰 Changes
I personally find myself checking if the current Operation is a Webhook or not. I looked at the `Operation` class and didn't see anything like this. Thought it'd be useful so we don't have to keep type checking in the main repo. 

I put the check on the Operation class since a Webhook will always be an Operation class but an Operation class might not be a Webhook or Callback class. 

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/README.md
+++ b/packages/oas/README.md
@@ -163,7 +163,7 @@ const operation = petstore.operation('/pet', 'post');
 | `.isJson()` | Determine if this operation requires its payload to be delivered as JSON. |
 | `.isMultipart()` | Determine if this operation requires its data to be sent as a multipart payload. |
 | `.isXml()` | Determine if this operation requires its data to be sent as XML. |
-| `.isWebhook()` | Determine if this operation is an instance of the `Webhook` class . |
+| `.isWebhook()` | Determine if this operation is an instance of the `Webhook` class. |
 | `.getExampleGroups()` | Returns an object with groups of all example definitions (body/header/query/path/response/etc.). The examples are grouped by their key when defined via the `examples` map. |
 | `.getHeaders()` | Retrieve all headers that can either be sent for or returned from this operation. This includes header-based authentication schemes, common header parameters, and request body and response content types. |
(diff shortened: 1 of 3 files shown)


packages/oas/src/operation/index.ts
```diff
@@ -153,6 +153,11 @@ export class Operation {
     return matchesMimeType.xml(this.getContentType());
   }
 
+  // Checks if the current operation is a webhook or not
```
erunion, line 156:
```suggestion
  /**
   * Checks if the current operation is a webhook or not.
   *
   */
```

erunion (changes requested):
can you also add this to the operation class docs in the readme?

packages/oas/src/operation/index.ts
darrenyong: updated comment format and added to readme!

packages/oas/README.md
```diff
@@ -163,6 +163,7 @@ const operation = petstore.operation('/pet', 'post');
 | `.isJson()` | Determine if this operation requires its payload to be delivered as JSON. |
 | `.isMultipart()` | Determine if this operation requires its data to be sent as a multipart payload. |
 | `.isXml()` | Determine if this operation requires its data to be sent as XML. |
+| `.isWebhook()` | Determine if this operation is an instance of the `Webhook` class . |
```
erunion, line 166:
```suggestion
| `.isWebhook()` | Determine if this operation is an instance of the `Webhook` class. |
```

packages/oas/src/operation/index.ts
```diff
@@ -153,6 +153,13 @@ export class Operation {
     return matchesMimeType.xml(this.getContentType());
   }
 
+  /**
+   * Checks if the current operation is a webhook or not
```
erunion, line 157:
```suggestion
   * Checks if the current operation is a webhook or not.
   *
```

erunion (approved):

Issue #901, opened 2024-09-10
Vulnerability path-to-regexp


loferreiranuno:
Theres a vulnerability for path-to-regexp https://github.com/advisories/GHSA-9wv6-86v2-598j .
Can you apply the fix?

erunion:
This is now resolved with `oas@25.0.0`.

PR #902, opened 2024-09-10
chore(deps): upgrading `path-to-regexp`

erunion:
| 🚥 Resolves https://github.com/readmeio/oas/issues/901 |
| :------------------- |

## 🧰 Changes

`path-to-regexp@7` has a vulnerability regarding backtick expressions https://github.com/advisories/GHSA-9wv6-86v2-598j so we need to upgrade it.

* [x] Swaps out `fetch-mock` in `oas-normalize` for `nock`. `fetch-mock` was using a **very** old version of `path-to-regexp`
* [x] Upgrades `path-to-regexp` in `oas`.
  * Unfortunately they've changed the the `MatchResult` interface we use to no longer return `index` so this may need to be treated as a breaking change as we have returned that value on `.findOperation()` and `.findOperationWithoutMethod()`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,222 +59,6 @@
         "node": ">=6.9.0"
       }
     },
-    "node_modules/@babel/compat-data": {
-      "version": "7.24.4",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.24.4.tgz",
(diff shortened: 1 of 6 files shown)


kanadgupta (approved):
lmao i was literally about to DM you about this ⚡

PR #904, opened 2024-09-12
fix: add empty `value` check in getRequestBodyExamples()

darrenyong:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes
There is a weird case in our main repo where we initialize `Webhooks` with an empty `requestBodyExamples` with the following structure:
```
requestBodyExamples: [
  {
    mediaType: 'application/json',
    examples: [{
      value: undefined
    }],
  },
];
```

This causes `Operation.getRequestBodyExamples()` to return with an empty request example even though we should be creating an example for them. This adds an extra check so that we avoid returning the empty example and proceed to the subsequent `getRequestBodyExamples()` call that generates an example for the user.

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -708,7 +708,7 @@ export class Operation {
    *
    */
   getRequestBodyExamples(): RequestBodyExamples {
-    const isRequestExampleValueDefined = typeof this.requestBodyExamples?.[0]?.examples?.[0].value !== 'undefined'
+    const isRequestExampleValueDefined = typeof this.requestBodyExamples?.[0]?.examples?.[0].value !== 'undefined';
 
     if (this.requestBodyExamples && isRequestExampleValueDefined) {
       return this.requestBodyExamples;
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -708,7 +708,9 @@ export class Operation {
    *
    */
   getRequestBodyExamples(): RequestBodyExamples {
-    if (this.requestBodyExamples) {
+    const nonEmptyRequestExample = this.requestBodyExamples?.[0]?.examples?.[0].value
+
+    if (this.requestBodyExamples && nonEmptyRequestExample) {
       return this.requestBodyExamples;
     }
 
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -708,9 +708,9 @@ export class Operation {
    *
    */
   getRequestBodyExamples(): RequestBodyExamples {
-    const nonEmptyRequestExample = this.requestBodyExamples?.[0]?.examples?.[0].value;
+    const isRequestExampleValueDefined = typeof this.requestBodyExamples?.[0]?.examples?.[0].value !== 'undefined'
 
-    if (this.requestBodyExamples && nonEmptyRequestExample) {
+    if (this.requestBodyExamples && isRequestExampleValueDefined) {
       return this.requestBodyExamples;
     }
 
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -708,7 +708,7 @@ export class Operation {
    *
    */
   getRequestBodyExamples(): RequestBodyExamples {
-    const nonEmptyRequestExample = this.requestBodyExamples?.[0]?.examples?.[0].value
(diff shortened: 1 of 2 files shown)


packages/oas/src/operation/index.ts
```diff
@@ -708,7 +708,9 @@ export class Operation {
    *
    */
   getRequestBodyExamples(): RequestBodyExamples {
-    if (this.requestBodyExamples) {
+    const nonEmptyRequestExample = this.requestBodyExamples?.[0]?.examples?.[0].value;
```
darrenyong, line 711: this is ugly :(

packages/oas/test/operation/lib/get-requestbody-examples.test.ts
```diff
@@ -5,20 +5,53 @@ import cleanStringify from '../../__fixtures__/json-stringify-clean.js';
...
+  webhooksOas = await import('@readme/oas-examples/3.1/json/webhooks.json').then(r => r.default).then(Oas.init);
+  await webhooksOas.dereference();
 });
 
 test('should return early if there is no request body', () => {
   const operation = operationExamples.operation('/nothing', 'get');
   expect(operation.getRequestBodyExamples()).toStrictEqual([]);
 });
 
+test('webhooks', () => {
```
erunion, line 26: can you give this a description?

packages/oas/test/operation/lib/get-requestbody-examples.test.ts
```diff
@@ -5,20 +5,53 @@ import cleanStringify from '../../__fixtures__/json-stringify-clean.js';
...
+  const webhookOperation = webhooksOas.operation('newPet', 'post', { isWebhook: true });
+  webhookOperation.requestBodyExamples = [
+    {
+      mediaType: 'application/json',
+      examples: [
+        {
+          value: undefined,
```
erunion, line 33: are you able to recreate this without having to set `.requestBodyExamples` to the operation instance?

erunion (dismissed):

packages/oas/src/operation/index.ts
kanadgupta: lol nauuur it's great! would this also solve the problem at hand? roughly the same from an ugliness standpoint but i think it's safer since it would allow falsy values like `0`, `false`, etc. (might also be good to have a test case for falsy values too!)

```suggestion
    const isRequestExampleValueDefined = typeof this.requestBodyExamples?.[0]?.examples?.[0].value !== 'undefined';
```

packages/oas/test/operation/lib/get-requestbody-examples.test.ts
kanadgupta: i'd be cool with creating a openapi definition fixture in this repo for this! don't think it's noteworthy enough to publish a new version of `oas-examples` or anything like that

packages/oas/test/operation/lib/get-requestbody-examples.test.ts
kanadgupta, line 26: something like this perhaps?

```suggestion
test('should generate a request example for a webhook operation with an undefined example value', () => {
```

kanadgupta (changes requested):
whoa incredible find here 🫨 small suggestion to account for a potential edge case but overall happy with this!

packages/oas/test/operation/lib/get-requestbody-examples.test.ts
```diff
@@ -5,20 +5,54 @@ import cleanStringify from '../../__fixtures__/json-stringify-clean.js';
...
+
+  webhooksOas = await import('@readme/oas-examples/3.1/json/webhooks.json').then(r => r.default).then(Oas.init);
 });
 
 test('should return early if there is no request body', () => {
   const operation = operationExamples.operation('/nothing', 'get');
   expect(operation.getRequestBodyExamples()).toStrictEqual([]);
 });
 
+test('should generate a request example for an operation with an undefined example value', async () => {
```
kanadgupta, line 25: now that we understand the underlying issue a bit better, i think this would be a better title:

```suggestion
test('should re-intialize the request examples after the oas is dereferenced', async () => {
```

kanadgupta (approved):
small test renaming suggestion, otherwise LGTM!

kanadgupta (approved):

PR #909, opened 2024-10-17
feat: make proxy address customizable

llimllib:
## 🧰 Changes

Convert proxyUrl from a boolean to a string, so now it represents the URL of the proxy server instead of an option about whether to use a proxy server or not.

This is a breaking change and `oas-to-har` should get a version major bump when it's released

The reason to make this change is so that I can test a beta version of try.readme.io safely in staging

## 🧬 QA & Testing

I added a test, but this should be a pretty minimal change

### DIFF
--- a/packages/oas-to-har/package.json
+++ b/packages/oas-to-har/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@readme/oas-to-har",
   "description": "Utility to transform an OAS operation into a HAR representation",
-  "version": "23.2.27",
+  "version": "23.3.0",
   "author": "Jon Ursenbach <jon@ursenba.ch>",
   "license": "ISC",
   "sideEffects": false,
--- a/packages/oas-to-har/package.json
+++ b/packages/oas-to-har/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@readme/oas-to-har",
   "description": "Utility to transform an OAS operation into a HAR representation",
-  "version": "23.3.0",
+  "version": "23.2.27",
   "author": "Jon Ursenbach <jon@ursenba.ch>",
   "license": "ISC",
   "sideEffects": false,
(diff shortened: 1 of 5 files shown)


packages/oas-to-har/package.json
kanadgupta, line 1: can you revert this change? we typically separate these into their own commit since this also will trigger a bump in `oas-to-snippet` and we use `lerna` to handle the version bumping

packages/oas-to-har/src/lib/types.ts
```diff
@@ -3,6 +3,9 @@ export type AuthForHAR = Record<string, number | string | { pass?: string; user?
 export { DataForHAR } from 'oas/types';
 
 export interface oasToHarOptions {
+  // The URL to use for the proxy; defaults to https://try.readme.io
+  proxyAddress?: string;
```
kanadgupta, line 7: can we just consolidate this and `proxyUrl` and we can make this a breaking change? the fact that these are separated out into two options is a bit confusing

kanadgupta (changes requested):

packages/oas-to-har/src/index.ts
```diff
@@ -223,9 +223,11 @@ export default function oasToHar(
...
   auth: AuthForHAR = {},
   opts: oasToHarOptions = {
-    // If true, the operation URL will be rewritten and prefixed with https://try.readme.io/ in
-    // order to funnel requests through our CORS-friendly proxy.
-    proxyUrl: false,
+    // the URL of a proxy to use. Requests will be preefixed with its value;
+    // for example if you use "https://try.readme.io", a request to
+    // "https://example.com/some/api" will be sent to
+    // "https://try.readme.io/https://example.com/some/api"
+    proxyUrl: '',
```
kanadgupta, line 230: i think you can remove all of this, including the default value (esp since hovering over `proxyUrl` will surface that description if you use my JSDoc suggestion from above)

```suggestion
```

llimllib:
@kanadgupta do you want to bump the version or shall I?

kanadgupta:
@llimllib i got it! i'm going to bump some dependencies also quick

llimllib:
OK, I'd like to get it into the monorepo asap so tag me on a review

PR #910, opened 2024-10-17
chore: prevent tests from being run in subpackages

kanadgupta:
## 🧰 Changes

This repo is a bit weird in that we need to build the dists before running our test suite, so running tests from the subpackages might result in confusing errors. This PR makes it so we have a bit of a nicer error in case of that.


packages/oas-to-har/test/index.test.ts
kanadgupta, line 1: smol type fix i noticed — we don't type check tests so it's not really a concern

llimllib (approved):

PR #915, opened 2024-11-20
fix: correctly handle `nullable: false`

darrenyong:
| 🚥 Resolves RM-11473 |
| :------------------- |

## 🧰 Changes
Currently, if a schema has `nullable: false`, (idk why since it defaults to false? maybe some tooling explicitly defines it as such) we aren't properly handling it as we're only checking for the existence of a `nullable` property. 

I added a condition to only add the `null` type if `nullable: true`. If it's `false`, we don't want to do anything with it, but we still want to delete it so it doesn't show up in the schema. Added a test too. 

## 🧬 QA & Testing


erunion (approved):
lmk when you want to publish this and i can handle it for ya

PR #918, opened 2024-12-04
feat(oas): add ref name to analyzer

mjcuva:
| 🚥 Resolves RM-11499 |
| :------------------- |

## 🧰 Changes

Add a check to the analyzer if the definition previously had a reference in it. Once we dereference a file we add `x-readme-ref-name` as a way to keep track of the previous name. This surfaces if that extension exists in the definition being analyzed. 

## 🧬 QA & Testing

Tests should pass, and analyzer should have a new `x-readme-ref-name` field being returned.

### DIFF
--- a/packages/oas/src/analyzer/index.ts
+++ b/packages/oas/src/analyzer/index.ts
@@ -31,6 +31,7 @@ export default async function analyzer(definition: OASDocument): Promise<OASAnal
   const staticHeaders = README_QUERIES.staticHeaders(definition);
   const rawBody = README_QUERIES.rawBody(definition);
   const refNames = README_QUERIES.refNames(definition);
+
   const analysis: OASAnalysis = {
     general: {
       mediaTypes: {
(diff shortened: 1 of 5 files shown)


packages/oas/src/analyzer/types.ts
```diff
@@ -33,6 +33,7 @@ export interface OASAnalysis {
      */
     raw_body?: OASAnalysisFeature;
     'x-default': OASAnalysisFeature;
+    'x-readme-ref-name': OASAnalysisFeature;
```
kanadgupta, line 36: mind adding a JS doc for this since this isn't a public extension?

kanadgupta (commented):
i'll defer to @erunion for final sign off since this is squarely his wheelhouse but looks good to me!

packages/oas/src/analyzer/types.ts
mjcuva: Done! Let me know if you would word it differently

kanadgupta: that's great, thank you!

packages/oas/src/analyzer/index.ts
```diff
@@ -30,7 +30,7 @@ export default async function analyzer(definition: OASDocument): Promise<OASAnal
   const explorerDisabled = README_QUERIES.explorerDisabled(definition);
   const staticHeaders = README_QUERIES.staticHeaders(definition);
   const rawBody = README_QUERIES.rawBody(definition);
-
+  const refNames = README_QUERIES.refNames(definition);
```
erunion, line 33:
```suggestion
  const refNames = README_QUERIES.refNames(definition);

```

erunion (approved):

PR #919, opened 2024-12-13
feat(oas): adding support for common parameter detection to the analyzer

erunion:
## 🧰 Changes

This adds support to `oas/analyzer` for surfacing information on if a given API definition utilizes [common parameters](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object).

I've also gone through and updated a bunch of bad deep links to parts of the OAS. Would be nice if the OAI would stop breaking these deep links every couple of years.

### DIFF
--- a/packages/oas/README.md
+++ b/packages/oas/README.md
@@ -119,8 +119,8 @@ Because this library has full TypeScript types and docblocks this README is not 
 <!-- prettier-ignore-start -->
 | Method | Description |
 | :--- | :--- |
-| `.getExtension()` | Retrieve a given [specification extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specificationExtensions) if it exists at the root of the API definition.  |
-| `.hasExtension()` | Determine if a given [specification extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specificationExtensions) exists on the root of the API definition. |
+| `.getExtension()` | Retrieve a given [specification extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specification-extensions) if it exists at the root of the API definition.  |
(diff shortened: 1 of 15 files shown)


packages/oas/src/analyzer/queries/openapi.ts
```diff
@@ -42,11 +42,21 @@ export async function circularRefs(definition: OASDocument) {
...
 }
 
+/**
+ * Determine if a given API definition utilizes common parameters.
+ *
+ * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#path-item-object}
+ * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object}
+ */
+export function commonParameters(definition: OASDocument) {
+  return query(['$..paths[*].parameters'], definition).map(res => refizePointer(res.pointer));
```
kanadgupta, line 52: i believe the answer is "no" but just to double check — are there any other places in the tree where a parameter can be defined (other than at the path- and operation-level)?

packages/oas/README.md
```diff
@@ -119,8 +119,8 @@ Because this library has full TypeScript types and docblocks this README is not
 <!-- prettier-ignore-start -->
 | Method | Description |
 | :--- | :--- |
-| `.getExtension()` | Retrieve a given [specification extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specificationExtensions) if it exists at the root of the API definition.  |
-| `.hasExtension()` | Determine if a given [specification extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specificationExtensions) exists on the root of the API definition. |
+| `.getExtension()` | Retrieve a given [specification extension](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specification-extensions) if it exists at the root of the API definition.  |
```
kanadgupta, line 122: all of this work looks crazy tedious and i'd love to figure out how to auto-generate docs like these at some point, but in the meantime... bless u for this

kanadgupta (approved):

PR #920, opened 2024-12-20
feat(oas-normalize): reworking how `.validate()` works, new `.convert()` method

erunion:
## 🐳  Context

Within our new Git backend we've run into a fun problem where with a refactor I'm working on to stop caching upconverted and dereferenced OpenAPI definitions in a database cache, instead opting to pull them straight from our Git filesystem, I have unfortunately did a number on an internal endpoint of ours' performance by reducing it from 300+ ops/sec to ~30.

The reason this happened is because in order for this endpoint to continue to return upconverted OpenAPI definitions (dereferencing is no longer wanted there) I have to run the file we have in Git, which may be an OpenAPI, Swagger, Postman, or YAML file, through the `.validate()` method in `oas-normalize` in order to do this conversion.

Because I already know that these API definitions are valid I just want to convert, however unfortunately `oas-normalize` doesn't have any available methods for this.

## 🧰 Changes

> [!IMPORTANT]  
> These are breaking changes to `oas-normalize`.

In order to resolve the above conundrum I am heavily reworking how the existing `.validate()` method works to now **only** validate a given API definition. It will continue to convert Postman collections to OpenAPI in order to validate them as OpenAPI definitons (because we don't have a Postman validator), but will now just return a boolean in the event of the definition being valid, or throwing an error if it isn't.

For the conversion side of this that `.validate()` no longer handles I am introducing a new `.convert()` method that will do as it says: convert a given API definition to an OpenAPI JSON object. If it's already OpenAPI, and JSON, it'll return what's already there, but if it's Swagger or Postman it'll do the conversion to OpenAPI there and return the result of that -- the same behavior that `.validate()` previously contained.

Additionally with this, the `.validate()` method had a `convertToLatest` flag that allowed you, if you so wished, to **not** convert a Swagger definition to OpenAPI after validating it. I am removing this flag because as far as I could tell we were always supplying `convertToLatest` (it was the default behavior). Instead in its place, if you just want to validate a Swagger definition now, and not convert it to OpenAPI, don't feed it through `.convert()` after running `.validate()`.

To be honest I always really disliked that `.validate()` would validate _and_ convert; seemed like a weird dark pattern and potential footgun.

## ⏱️ Benchmarks

Because the work that spawned this was discovered through the process of writing a benchmark I'm also creating some new benchmarks here for the `.validate()` and `.convert()` methods. Diffing the current `.validate()` benchmark versus the new `.convert()` one you can see pretty clearly that this new `.convert()` method will greatly improve our performance over the current implementation and usage of `.validate()`.

#### `.validate()` currently in `main`

validate-before

#### `.validate()` now

validate-after

#### `.convert()`

Screenshot 2024-12-19 at 4 11 12 PM


.github/workflows/ci.yml
```diff
@@ -27,8 +27,14 @@ jobs:
...
-      - name: Install deps and run tests
-        run: npm cit
+      - name: Install deps
+        run: npm ci
+
+      - name: Run tests
+        run: npm test
+
+      - name: Run benchmarks
+        run: npm run bench
```
erunion, line 37: It would be rad if there were a github action to dump the contents of these benchmarks into a PR comment.

packages/oas-normalize/README.md
```diff
@@ -27,126 +27,161 @@ npm install oas-normalize
...
+    // The API definition is valid!
   })
   .catch(err => {
-    console.log(err);
+    console.error(err);
   });
 ```
 
-### `#bundle()`
+### `.load()`
```
erunion, line 48: `.load()` has long been marked as private with a `@private` JSDoc but we actually use it in a number of apps of ours so might as well make it _actually_ public and documented.

packages/oas-normalize/src/index.ts
```diff
@@ -81,10 +83,7 @@ export default class OASNormalize {
     }
   }
 
-  /**
-   * @private
-   */
-  static async convertPostmanToOpenAPI(schema: any) {
+  private static async convertPostmanToOpenAPI(schema: any) {
```
erunion, line 86: Making this one _actually_ private because it's only used internally.

.github/workflows/ci.yml
domharrington: Yeahh 100%. Like i said in slack, i think you'd have to store the benchmark results somewhere and compare the current run to a known baseline or something.

domharrington (approved):
Yes!! Good job 👏

packages/oas/src/reducer/index.ts
```diff
@@ -1,7 +1,6 @@
 import type { ComponentsObject, HttpMethods, OASDocument, TagObject } from '../types.js';
 
 import jsonPointer from 'jsonpointer';
-import { getAPIDefinitionType } from 'oas-normalize/lib/utils';
```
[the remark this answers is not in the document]
kanadgupta: ty for this

packages/oas-normalize/test/index.test.ts
```diff
@@ -271,13 +332,22 @@ describe('#validate', () => {
...
     });
 
-    await expect(o.validate()).rejects.toMatchSnapshot();
+    try {
+      await o.validate();
+      assert.fail();
+    } catch (err) {
+      expect(err).toBeInstanceOf(SyntaxError);
+      expect(err.message).toMatchSnapshot();
+      expect(err.details).toMatchSnapshot();
```
kanadgupta, line 347: i think you should be able to do this instead:
```suggestion
      expect(err).toMatchSnapshot();
```

packages/oas-normalize/src/index.ts
```diff
@@ -141,19 +140,48 @@ export default class OASNormalize {
...
-      /**
-       * Automatically convert the supplied API definition to the latest version of OpenAPI.
-       */
-      convertToLatest?: boolean;
       parser?: openapiParser.Options;
-    } = { convertToLatest: false },
-  ): Promise<OpenAPI.Document> {
-    const convertToLatest = opts.convertToLatest;
+    } = {},
+  ): Promise<boolean> {
```
kanadgupta, line 184: this technically only ever returns `true`, correct? since it just throws an error otherwise?
```suggestion
  ): Promise<true> {
```

packages/oas-normalize/README.md
```diff
@@ -27,126 +27,161 @@ npm install oas-normalize
...
+   8 |   "servers": [
+>  9 |     {
+     |     ^ ☹️  url is missing here!
+  10 |       "urll": "http://petstore.swagger.io/v2"
+  11 |     }
+  12 |   ],
 ```
 
-`message` is almost always there, but `path` is less dependable.
+However if you would like to programatically access this information the `SyntaxError` error that is thrown contains a `details` array of [AJV](https://npm.im/ajv) errors:
```
kanadgupta, line 133: TIL!

packages/oas-normalize/README.md
kanadgupta, line 1: thoughts on marking our support for postman collections as experimental? and maybe adding a callout like this?

> [!WARNING]
> Support for Postman collections is experimental. If you've supplied a Postman collection to the library, it will **always** be converted to OpenAPI, using [`@readme/postman-to-openapi`](https://npm.im/@readme/postman-to-openapi) before doing any bundling, validating, etc.

packages/oas-normalize/README.md
```diff
@@ -27,126 +27,161 @@ npm install oas-normalize
...
+```ts
+const definition = await oas.bundle();
+console.log(definition);
 ```
 
-### `#validate({ convertToLatest?: boolean })`
+### `.validate()`
 
-Validate and optionally convert to OpenAPI, a given API definition. This supports Swagger 2.0, OpenAPI 3.x API definitions as well as Postman 2.x collections.
+Validate a given API definition. This supports Swagger 2.0 and OpenAPI 3.x API definitions as well as Postman 2.x collections.
```
kanadgupta, line 102:
```suggestion
Validate a given API definition. This supports Swagger 2.0 and OpenAPI 3.x API definitions, as well as Postman 2.x collections.
```

kanadgupta (changes requested):
only blocking suggestion is the type change but otherwise very supportive of this! let's chat offline (maybe after the holiday) about `rdme` implications b/c i think we need to revisit how we handle API definition uploads in `rdme@10`

packages/oas-normalize/README.md
julshotal: +1 to this being a good callout

julshotal (approved):

packages/oas-normalize/src/index.ts
erunion: TS doesn't let you do that.

Screenshot 2024-12-20 at 9 46 08 AM

packages/oas-normalize/test/index.test.ts
erunion: That unfortunately only snapshots the error message.

Screenshot 2024-12-20 at 9 49 57 AM

packages/oas-normalize/src/index.ts
kanadgupta: oh i think it's because we're returning `Promise.reject()` rather than throwing the errors — just confirmed locally that this passes:

CleanShot 2024-12-20 at 11 52 58@2x

erunion: hm yeah idk why i did promise.reject here. will move them over

packages/oas-normalize/README.md
```diff
@@ -27,126 +27,156 @@ npm install oas-normalize
...
+console.log(definition);
 ```
 
-### `#validate({ convertToLatest?: boolean })`
+### `.validate()`
 
-Validate and optionally convert to OpenAPI, a given API definition. This supports Swagger 2.0, OpenAPI 3.x API definitions as well as Postman 2.x collections.
+Validate a given API definition. This supports Swagger 2.0 and OpenAPI 3.x API definitions, as well as Postman 2.x collections.
 
 Please note that if you've supplied a Postman collection to the library it will **always** be converted to OpenAPI, using [@readme/postman-to-openapi](https://npm.im/@readme/postman-to-openapi), and we will only validate resulting OpenAPI definition.
```
kanadgupta, line 99: now that we have that callout you probably don't need this anymore:

```suggestion
```

kanadgupta (approved):

packages/oas-normalize/README.md
erunion: damn, thought i got all of them

packages/oas-normalize/test/index.test.ts
kanadgupta: ah okay — this technically also handles the `expect(err).toBeInstanceOf(SyntaxError);` assertion also but yeah two steps forward two steps back here so fine with this as is

PR #922, opened 2025-01-04
docs: moved method to correct chapter

niklashaug:
I moved the `hasRequiredRequestBody` to the Request Body chapter


erunion (approved):
thank you!

Issue #925, opened 2025-01-23
[oas/reducer] does not support deeply nested $refs


erunion:
If you have a `$ref` like the following `#/components/examples/event-min/value` when we reduce the operation we will exclude `#/components/examples/event-min` because the `$ref` accumulator within the reducer (the thing that determines what refs are used), only looks at `#/components/examples/event-min`. Reducing an operation with a `$ref` like this results in data loss.

There is a skipped test for this case now in https://github.com/readmeio/oas/blob/eb81371f41c68a8e0f050660f5d893dfa0c87b35/packages/oas/test/reducer/index.test.ts#L210-L221

PR #926, opened 2025-01-23
fix(oas): account for deep `$ref` pointers when reducing an API def

erunion:
| 🚥 Resolves https://github.com/readmeio/oas/issues/925 |
| :------------------- |

## 🧰 Changes

This updates `oas/reducer` to account for deep `$ref` pointers like `#/components/examples/event-min/value`. Normally when we're running through components to remove we only look at `#/components/examples/event-min`, however when a schema like this is deeply referenced we won't pick up that it's used and end up removing `event-min` from the components block, resulting in a corrupted schema that will no longer validate.

### DIFF
--- a/packages/oas/src/reducer/index.ts
+++ b/packages/oas/src/reducer/index.ts
@@ -39,6 +39,12 @@ function accumulateUsedRefs(schema: Record<string, unknown>, $refs: Set<string>,
   }
 
   getUsedRefs($refSchema).forEach(({ value: currRef }) => {
+    // Because it's possible to have a parameter named `$ref`, which our lookup would pick up as a
+    // false positive, we want to exclude that from `$ref` matching as it's not really a reference.
+    if (typeof currRef !== 'string') {
+      return;
+    }
+
     // If we've already processed this $ref don't send us into an infinite loop.
     if ($refs.has(currRef)) {
       return;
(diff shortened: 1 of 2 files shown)


packages/oas/test/reducer/index.test.ts
```diff
@@ -17,16 +15,6 @@ import reduceQuirks from '../__datasets__/reduce-quirks.json';
 import securityRootLevel from '../__datasets__/security-root-level.json';
 import tagQuirks from '../__datasets__/tag-quirks.json';
 
-declare global {
-  interface Console {
-    logx: any;
-  }
-}
-
-console.logx = (obj: any) => {
```
kanadgupta, line 26: lol wut

packages/oas/src/reducer/index.ts
```diff
@@ -178,7 +184,18 @@ export default function reducer(definition: OASDocument, opts: ReducerOptions =
   if ('components' in reduced) {
     Object.keys(reduced.components).forEach((componentType: keyof ComponentsObject) => {
       Object.keys(reduced.components[componentType]).forEach(component => {
-        if (!$refs.has(`#/components/${componentType}/${component}`)) {
+        // If our `$ref` either is a full, or deep match, then we should preserve it.
+        const refIsUsed =
+          $refs.has(`#/components/${componentType}/${component}`) ||
```
kanadgupta, line 189: (non-blocking) i feel like i've seen ref syntax that uses `#components` and not `#/components` — is that valid? should we account for that here?

kanadgupta (approved):
smol comment below, potentially outside the scope of this PR

packages/oas/test/reducer/index.test.ts
erunion: this snuck in in a commit i made this morning where i added that skipped failing test. `console.logx` is a tool i use to dump full objects.

packages/oas/src/reducer/index.ts
```diff
@@ -39,6 +39,12 @@ function accumulateUsedRefs(schema: Record<string, unknown>, $refs: Set<string>,
   }
 
   getUsedRefs($refSchema).forEach(({ value: currRef }) => {
+    // Because it's possible to have a parameter named `$ref`, which our lookup would pick up as a
+    // false positive, we want to exclude that from `$ref` matching as it's not really a reference.
```
julshotal, line 43: of course you can do that 🙃

julshotal (approved):

packages/oas/src/reducer/index.ts
erunion: i have also seen that before but it is not valid. our parser doesn't pick up on it now, but should at some point.

packages/oas/src/reducer/index.ts
erunion: we love it here, folks

packages/oas/src/reducer/index.ts
erunion: i had a note about this in one of my gitto PRs over the past two weeks but can't link those here.

PR #933, opened 2025-02-14
refactor: modern TS touchups

kanadgupta:
## 🧰 Changes

since we're on modern TS, i enabled the following options in our `tsconfig.base.json` and refactored a bunch of things downstream to account for these changes:
- [`isolatedDeclarations`](https://www.typescriptlang.org/tsconfig/#isolatedDeclarations) - the types in this repo are rather complex, so this should help speed up typechecking (both in this repo and in any TS repo that uses these libraries) and type generation
- [`isolatedModules`](https://www.typescriptlang.org/tsconfig/#isolatedModules) - [per matt pocock](https://www.totaltypescript.com/tsconfig-cheat-sheet#base-options): "this option prevents a few TS features which are unsafe when treating modules as isolated files". didn't have to do any add'l work to enable this so 🤷🏽
- [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) — this makes it so we're stricter about `import type` declarations. this should help with treeshaking.

matt pocock recommends both of these options:
https://www.totaltypescript.com/tsconfig-cheat-sheet#base-options
https://www.totaltypescript.com/typescript-5-3#isolated-declarations

## 🧬 QA & Testing

does everything still pass?

### DIFF
--- a/packages/oas/src/utils.ts
+++ b/packages/oas/src/utils.ts
@@ -2,6 +2,8 @@ import findSchemaDefinition from './lib/find-schema-definition.js';
 import matchesMimeType from './lib/matches-mimetype.js';
 import { types as jsonSchemaTypes } from './operation/lib/get-parameters-as-json-schema.js';
 
-const supportedMethods = new Set(['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace']);
+const methods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'] as const;
+
+const supportedMethods: Set<(typeof methods)[number]> = new Set(methods);
 
 export { findSchemaDefinition, jsonSchemaTypes, matchesMimeType, supportedMethods };
--- a/packages/oas/src/utils.ts
+++ b/packages/oas/src/utils.ts
@@ -2,8 +2,6 @@ import findSchemaDefinition from './lib/find-schema-definition.js';
 import matchesMimeType from './lib/matches-mimetype.js';
 import { types as jsonSchemaTypes } from './operation/lib/get-parameters-as-json-schema.js';
 
-const methods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'] as const;
-
-const supportedMethods: Set<(typeof methods)[number]> = new Set(methods);
+const supportedMethods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'] as const;
 
 export { findSchemaDefinition, jsonSchemaTypes, matchesMimeType, supportedMethods };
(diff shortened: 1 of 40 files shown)


packages/oas/src/utils.ts
```diff
@@ -2,6 +2,8 @@ import findSchemaDefinition from './lib/find-schema-definition.js';
 import matchesMimeType from './lib/matches-mimetype.js';
 import { types as jsonSchemaTypes } from './operation/lib/get-parameters-as-json-schema.js';
 
-const supportedMethods = new Set(['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace']);
+const methods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'] as const;
+
+const supportedMethods: Set<(typeof methods)[number]> = new Set(methods);
```
erunion, line 7: could honestly probably do away with this `Set`. the only place where we're using this is in `oas/src/index.ts` where we're doing a `.has()` call on it. could be refactored to an `.includes()` check.

tsconfig.base.json
```diff
@@ -1,17 +1,19 @@
...
+    "isolatedModules": true,
     "lib": ["ES2023"],
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "noImplicitAny": true,
     "outDir": "./dist",
     "skipLibCheck": true,
     "sourceMap": true,
-    "target": "ES2020"
+    "target": "ES2020",
```
erunion, line 16: should we bump this to ES2024?

tsconfig.base.json
```diff
@@ -1,17 +1,18 @@
 {
   "compilerOptions": {
-    "allowJs": true,
```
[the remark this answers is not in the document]
erunion: I'm going to flip on [readme/prefer-typescript](https://github.com/readmeio/standards/blob/main/packages/eslint-plugin/docs/rules/prefer-typescript.md) in this repo later today to enforce that we don't have any `*.js` files at all.

erunion (approved):
king shit

tsconfig.base.json
kanadgupta: sure!

packages/oas/src/utils.ts
kanadgupta: cool, will fix

kanadgupta: https://github.com/readmeio/oas/pull/933/commits/e1843963b5529f2e1c328439bdc19cd75ce9a73d

tsconfig.base.json
kanadgupta: https://github.com/readmeio/oas/pull/933/commits/a5ad9e1e07fbd0e98e3068b8a46500e0d352e376

PR #936, opened 2025-02-18
feat(parser): rewriting the parser (again) to make it fully static

erunion:
## 🐳 Context

In working to replace `@apidevtools/json-schema-ref-parser` in `oas` with our newly refreshed and rewritten OpenAPI parser (`packages/parser`) I uncovered a problem where API definitions that would previously be dereferenced without any issues were now failing validation. It turns out that when I rewrote the parser last week I had tied the `parse()` method so closely to everything and the bespoke validation that happened in `parse` was now being run when it hadn't been before.

Because we **don't** want that to happen, and also because the bespoke validation in `parse` is better suited for `validate()` I've rewritten our OpenAPI parser... again.

## 🧰 Changes

* [x] **Fully** decoupled our parser from `@apidevtools/json-schema-ref-parser` by having our parser package now be a series of exported functions: `parse`, `validate`, `bundle`, and `dereference`.
* [x] Paired down the configurable options that the parser had previously to now just contain a few things you can configure against `json-schema-ref-parser` (eg. you can no longer optionally disable spec or schema validation). Basically any option that we've never set ourselves with this library is no longer an option.
* [x] Either fully removed or moved the bespoke validation that was happening in `parse` to be done as part of `validate`. Many of the things that were being validated against within that are already able to be captured and validated as part of our standard schema validation process with Ajv so there was no need to have two systems validating the same thing.
* [x] Updated `@readme/better-ajv-errors` to remove emojis from error messages that had them. After some internal discussion we came to the agreement that they weren't really adding anything that wasn't already handled by our usage of [@babel/code-frame](https://npm.im/@babel/code-frame).
* [x] Updated the docs.


packages/parser/src/repair.ts
erunion, line 1: this work already existed but was under `src/utils.ts`.

packages/parser/test/specs/circular/circular.test.ts
```diff
@@ -13,67 +12,41 @@ import validatedAPI from './validated.js';
...
-    const parser = new OpenAPIParser();
-    const api = await parser.parse(path.rel('specs/circular/circular.yaml'));
+    const api = await parse(relativePath('specs/circular/circular.yaml'));
 
-    expect(api).to.equal(parser.schema);
     expect(api).to.deep.equal(parsedAPI.api);
-    expect(parser.$refs.paths()).to.deep.equal([path.abs('specs/circular/circular.yaml')]);
   });
 
-  it(
```
erunion, line 24: Removed these because our parser no longer has a resolve method, it just uses the internal `json-schema-ref-parser` one when we invoke methods in that library.

packages/parser/test/specs/invalid/no-paths-or-webhooks.yaml
```diff
@@ -1,4 +1,4 @@
-openapi: '3.1'
+openapi: '3.1.0'
```
erunion, line 1: Changed this because this spec is testing validation for when `paths` and `webhooks` aren't present but `openapi` not being `3.1.0` was throwing spec validation.

packages/parser/test/specs/object-source/object-source.test.ts
```diff
@@ -1,33 +1,18 @@
...
     expect(api).to.deep.equal(dereferencedAPI);
-
-    // The API path should be the current directory, and all other paths should be absolute
-    const expectedPaths = [
-      path.cwd(),
-      path.abs('specs/object-source/definitions/definitions.json'),
-      path.abs('specs/object-source/definitions/name.yaml'),
-      path.abs('specs/object-source/definitions/required-string.yaml'),
-    ];
-    expect(parser.$refs.paths()).to.have.same.members(expectedPaths);
```
erunion, line 27: Our parser no longer exposes the `$refs` property that `json-schema-ref-parser` has so I've removed these assertions. That `$refs` stuff is well tested within `json-schema-ref-parser` already.

packages/parser/test/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "isolatedDeclarations": false,
```
erunion, line 4: Turned this off for tests because it was causing some weird typing issues in a couple fixtures.

packages/parser/README.md
```diff
@@ -53,87 +50,73 @@ npm install @readme/openapi-parser
...
 
 // The `api` object is a normal JSON object so you can access any part of the
 // API definition using object notation.
 console.log(api.definitions.person.properties.firstName); // => { type: "string" }
 ```
 
-#### `.bundle()`
+### `.bundle()`
 
 Bundles all referenced files and URLs into a single API definition that only has _internal_ `$ref` pointers. This lets you split up your definition however you want while you're building it, but later combine all those files together when it's time to package or distribute the API definition to other people. The resulting defintiion size will be small, since it will still contain _internal_ JSON references rather than being fully-dereferenced.
```
mjcuva, line 100:
```suggestion
Bundles all referenced files and URLs into a single API definition that only has _internal_ `$ref` pointers. This lets you split up your definition however you want while you're building it, but later combine all those files together when it's time to package or distribute the API definition to other people. The resulting definition size will be small, since it will still contain _internal_ JSON references rather than being fully-dereferenced.
```

mjcuva (approved):
Just found one small typo in the docs (that's probably been there for ages haha) but otherwise this looks really good! To me it seems a lot easier to read through and reason about, and I like this pattern for using it a lot more.

kanadgupta (approved):
**very** excited for all these changes 🫶🏽

a few smol suggestions but nothing big nor blocking. great work!

packages/parser/src/repair.ts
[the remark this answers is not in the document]
erunion: yea

kanadgupta (approved):

PR #937, opened 2025-02-18
feat(oas): replacing `json-schema-ref-parser` with our new parser library

erunion:
## 🧰 Changes

This fully swaps our our, now deprecated, `json-schema-ref-parser` fork with our OpenAPI parser as it does everything that we were using that for.

### DIFF
--- a/packages/oas-to-snippet/package.json
+++ b/packages/oas-to-snippet/package.json
@@ -61,7 +61,6 @@
     "httpsnippet-client-api": "^7.0.0-beta.4",
     "oas": "file:../oas",
     "tsup": "^8.0.2",
-    "type-fest": "^4.18.3",
     "typescript": "^5.2.2",
     "vitest": "^3.0.5"
   },
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -1072,6 +1072,7 @@ export default class Oas {
         this.dereferencing = {
           processing: false,
           complete: true,
+          // We need to convert our `Set` to an array in order to match the typings.
           circularRefs: [...circularRefs],
         };
 
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -3,7 +3,7 @@ import type * as RMOAS from './types.js';
 import type { OpenAPIV3_1 } from 'openapi-types';
 import type { Match, ParamData } from 'path-to-regexp';
 
-import $RefParser from '@readme/json-schema-ref-parser';
+import { dereference } from '@readme/openapi-parser';
 import { pathToRegexp, match } from 'path-to-regexp';
 
 import {
@@ -286,8 +286,8 @@ export default class Oas {
       // eslint-disable-next-line try-catch-failsafe/json-parse
       oas = JSON.parse(oas) as RMOAS.OASDocument;
     }
-    // @todo throw an exception here instead of allowing an empty oas
-    this.api = oas;
+
+    this.api = oas || ({} as RMOAS.OASDocument);
     this.user = user || {};
 
     this.promises = [];
@@ -1019,7 +1019,7 @@ export default class Oas {
     // Because referencing will eliminate any lineage back to the original `$ref`, information that
     // we might need at some point, we should run through all available component schemas and denote
     // what their name is so that when dereferencing happens below those names will be preserved.
-    if (api && api.components && api.components.schemas && typeof api.components.schemas === 'object') {
+    if (api.components && api.components.schemas && typeof api.components.schemas === 'object') {
       Object.keys(api.components.schemas).forEach(schemaName => {
         // As of OpenAPI 3.1 component schemas can be primitives or arrays. If this happens then we
         // shouldn't try to add `title` or `x-readme-ref-name` properties because we can't. We'll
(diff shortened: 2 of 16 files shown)


packages/oas/test/analyzer/queries/openapi.test.ts
```diff
@@ -89,8 +89,13 @@ describe('analyzer queries (OpenAPI)', () => {
   describe('circularRefs', () => {
     it('should determine if a definition has circular refs when it does', async () => {
       await expect(QUERIES.circularRefs(circular)).resolves.toStrictEqual([
+        '#/components/schemas/BodyPart/properties/parent',
```
erunion, line 92: `json-schema-ref-parser` apparently got some improvements to circular reference detection because these new expectations are definitely circular.

packages/parser/src/repair.ts
```diff
@@ -19,12 +17,16 @@ function fixServers(
   server: OpenAPIV3_1.ReferenceObject | OpenAPIV3.ParameterObject | OpenAPIV3.ServerObject,
   path: string,
 ) {
-  // A erver URL starting with "/" tells that it is not an HTTP(s) URL.
+  // A server URL starting with "/" tells that it is not an HTTP(s) URL.
   if (server && 'url' in server && server.url && server.url.startsWith('/')) {
-    const inUrl = url.parse(path);
```
erunion, line 24: The `url.parse` method that we were importing from this was literally just a wrapper for `new URL()`.

packages/parser/src/lib/reduceAjvErrors.ts
```diff
@@ -1,4 +1,4 @@
-import type { ErrorObject } from 'ajv/dist/2020';
+import type { ErrorObject } from 'ajv/dist/2020.js';
```
erunion, line 1: We had to add these to resolve some weird ESM-y errors that were being thrown

packages/parser/src/index.ts
```diff
@@ -1,7 +1,6 @@
 import type { APIDocument, ParserOptions } from './types.js';
 
-import $RefParser from '@apidevtools/json-schema-ref-parser';
-import _dereference from '@apidevtools/json-schema-ref-parser/dist/lib/dereference';
+import $RefParser, { dereferenceInternal } from '@apidevtools/json-schema-ref-parser';
```
erunion, line 3: Noticed that they were exporting the dereference method we were using from the main export so I refactored this to not have us pull it out of the dist directory.

packages/parser/src/types.ts
```diff
@@ -17,14 +19,14 @@ export interface ParserOptions {
...
      */
-    circular?: $RefParserOptions['dereference']['circular'];
+    circular?: boolean | 'ignore';
 
     /**
      * Callback invoked during circular reference detection.
      *
      * @param path - The path that is circular (ie. the `$ref` string)
      */
-    onCircular?: $RefParserOptions['dereference']['onCircular'];
```
erunion, line 27: I tried to refactor this to use the `ParserOptions` export that `json-schema-ref-parser` offers, and not using the default from the dist, but the `DeepPartial` type that hey have wrapping `ParserOptions` blows away the types on functions and turns them into `{}`. 

Spent some time today trying to resolve this so I could submit a PR but threw in the towel because I couldn't get it working right.

packages/parser/src/util.ts
```diff
@@ -37,7 +38,7 @@ export function normalizeArguments<S extends APIDocument = APIDocument>(
  *
  */
 export function convertOptionsForParser(options: ParserOptions): Partial<$RefParserOptions> {
-  return getNewOptions({
+  return merge(getJsonSchemaRefParserDefaultOptions(), {
```
erunion, line 41: Refactored this to not use the `getNewOptions` method that they have in their dist that essentially just invokes a custom `merge()` method. I tried my best to _not_ do this with lodash but ran into a whole bunch of weird issues. Might take another stab in a followup because I would really love to remove lodash from this.

packages/oas-to-snippet/package.json
```diff
@@ -61,7 +61,6 @@
     "httpsnippet-client-api": "^7.0.0-beta.4",
     "oas": "file:../oas",
     "tsup": "^8.0.2",
-    "type-fest": "^4.18.3",
```
kanadgupta, line 64: wonder if we should add `knip` into this repo

packages/oas/src/index.ts
```diff
@@ -1044,39 +1044,35 @@ export default class Oas {
...
-        }
-
         this.api = dereferenced;
 
         this.promises = promises;
         this.dereferencing = {
           processing: false,
           complete: true,
-          circularRefs,
+          circularRefs: [...circularRefs],
```
kanadgupta, line 1075: mind adding a comment about why we're doing this?

kanadgupta (approved):

packages/oas-to-snippet/package.json
erunion: yeah, will in a followup

packages/oas/src/index.ts
```diff
@@ -1044,39 +1044,36 @@ export default class Oas {
...
-        }
-
         this.api = dereferenced;
 
         this.promises = promises;
         this.dereferencing = {
           processing: false,
           complete: true,
-          circularRefs,
+          // We need to convert our `Set` to an array in order to match the typings.
```
kanadgupta, line 1075: Oh oops didn’t realize this was a set 🫠

PR #939, opened 2025-02-19
feat(parser): replacing `ono` with regular errors

erunion:
## 🧰 Changes

* [x] Replaces our dependency on `@jsdevtools/ono` with a mix of `ReferenceError` and a new custom `ValidationError` class. I'm still not sold on using these three though and will likely revisit them when I start on the new error warning level system.
* [x] Rewriting most every bespoke error that we throw so that they're more consistent, have properties wrapped in backticks, or to properly use list formatting.
* [x] Removes our dependency on `lodash` for building configs for `json-schema-ref-parser`.

### DIFF
--- a/packages/parser/src/validators/spec/swagger.ts
+++ b/packages/parser/src/validators/spec/swagger.ts
@@ -1,5 +1,6 @@
 import type { IJsonSchema, OpenAPIV2 } from 'openapi-types';
 
+import { ValidationError } from '../../errors.js';
 import { swaggerHTTPMethods, pathParameterTemplateRegExp } from '../../lib/index.js';
 
 /**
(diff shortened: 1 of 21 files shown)


packages/parser/src/validators/spec/swagger.ts
```diff
@@ -113,25 +111,27 @@ function validatePathParameters(params: OpenAPIV2.ParameterObject[], pathId: str
...
+          `Validation failed. \`${operationId}\` has a path parameter named \`${param.name}\`, but there is no corresponding \`{${param.name}}\` in the path string.`,
         );
       }
 
       placeholders.splice(match, 1);
     });
 
   if (placeholders.length > 0) {
-    throw ono.syntax(`Validation failed. ${operationId} is missing path parameter(s) for ${placeholders}`);
+    const list = new Intl.ListFormat('en', { style: 'long', type: 'conjunction' }).format(
```
kanadgupta, line 130: `Intl` seems so so handy for stuff like this

kanadgupta (approved):
really excited about this 🥹 `ono` has given me grief for years. good riddance 🚮

barring the test failures (looks like it's mostly just outdated snapshots), LGTM!

packages/parser/src/validators/spec/swagger.ts
erunion: big `Intl` head over here

PR #943, opened 2025-02-24
feat(parser): new system for downgrading some OpenAPI errors to warnings

erunion:
## 🐳 Context

We need a new system in ReadMe to downgrade certain types of OpenAPI specification errors to warnings, for customers that use our legacy API editor, in order to migrate API definitions that aren't in the best shape into our new Git infrastructure.

To facilitate this new product requirement we need to totally overhaul how our OpenAPI parser internally locates and returns errors, resulting in an almost complete rewrite of our parsing engines error handling systems.

> [!IMPORTANT]
> This is a breaking change for most all of our OpenAPI tooling that exist within this repository.

### 🪴 OpenAPI Parser

Our OpenAPI Parser, `@readme/openapi-parser`, will no longer return a dereferenced schema when you call `validate()`, in its place a `ValidationResult` object is now returned.

```ts
{
  valid: boolean;
  errors: ErrorDetails[]; // Only present when `valid` is `false`.
  warnings: WarningDetails[];
  additionalErrors: number;
  specification: "OpenAPI" | "Swagger" | null;
}
```

Paired with this `validate()` can now be supplied with a new ESLint-like ruleset to optionally downgrade specific OpenAPI specification-level problems to be a general warning which considering the API definition they exist within to be valid:

```ts
const result = await validate(petstore, {
  validate: {
    rules: {
      openapi: {
        "path-parameters-not-in-path": "warning",
      },
    },
  },
});
```

When these warnings, previously having been considered to be an error, are detected they are thrown into the `warnings` array of `ValidationResult`.

```json
{
  "valid": true,
  "warnings": [
    {
      "message": "`/paths/users/{username}/get` has a path parameter named `username`, but there is no corresponding `{username}` in the path string."
    }
  ],
  "additionalErrors": 0,
  "specification": "OpenAPI"
}
```

Some of the rules that we now support are:

- [x] `array-without-items`
- [x] `duplicate-non-request-body-parameters`
- [x] `duplicate-operation-id`:
- [x] `non-optional-path-parameters`
- [x] `path-parameters-not-in-parameters`
- [x] `path-parameters-not-in-path`

With this `ValidationResult` we are also now able to return more than one specification-level problem at a time, previously these were always detected and immediately thrown as a `SyntaxError` exception.

Additionally because `validate()` no longer throws any exceptions at all, unless the supplied API definition is neither Swagger or OpenAPI, I'm bundling in a new `compileErrors` utility function that can be used to compile a `ValidationResult` to a human-readable string:

```ts
import { compileErrors } from "@readme/openapi-parser";

console.log(await validate(petstore).then(compileErrors));
```

```
OpenAPI schema validation failed.

ADDITIONAL PROPERTY must NOT have additional properties

  17 |             "name": "status",
  18 |             "in": "query",
> 19 |             "type": "array",
     |             ^^^^^^ type is not expected to be here!
  20 |             "schema": {
  21 |               "items": {
  22 |                 "type": "string","
```

##### Additional Changes

- [x] The `validate.colorizeErrors` option has been moved to `validate.errors.colorize`
- [x] Because validate now just purely validates an API defintiion it no longer executes any schema repair work that the other methods (`dereference`, `bundle`, etc.) for fixing relative OpenAPI server configs.
- [x] Our default HTTP timeout for retrieving API definitions has been reduced down to 5 seconds. The [@apidevtools/json-schema-ref-parser](https://npm.im/@apidevtools/json-schema-ref-parser) default is 60.

### 🏪 oas-normalize

In order to make this new warning system useful from the entry point of `oas-normalize` I have changed `.validate()` to now return the parsers `ValidationResult` instead of `true` if an API definition is valid.

For invalid API definitions a `ValidationError` exception will continue to be thrown, and for translating a `ValidationResult` object into a human-readable string I am also re-exporting `compileErrors` from the parser into `oas-normalize/lib/utils`.

##### Additional Changes

* [x] Because we're returning this `ValidationResult` object directly now everywhere we are no longer exposing the raw AJV errors that the parser used to generate human-friendly messages with [@readme/better-ajv-errors](https://npm.im/@readme/better-ajv-errors)

## ⏭️ Next Steps

This is all an absolute breaking change so I'll be bumping, likely, every package in this repository by a main release, however because our parser is also used in [jest-expect-openapi](https://npm.im/jest-expect-openapi) I plan on pulling that, and [jest-expect-har](https://npm.im/jest-expect-har), into this repository and updating that library to support this new parser mechanism before publishing any packages.

Once those packages are out then begins the work of actually integrating these new changes into our backend. 🙃


packages/parser/README.md
```diff
@@ -92,7 +151,7 @@ const api = await dereference(myAPI);
 
 // The `api` object is a normal JSON object so you can access any part of the
 // API definition using object notation.
-console.log(api.definitions.person.properties.firstName); // => { type: "string" }
+console.log(api.components.schemas.person.properties.firstName); // => { type: "string" }
```
erunion, line 154: Changed this example to be less Swagger-y

packages/parser/src/validators/spec/openapi.ts
erunion, line 1: The diff for this and `spec/swagger.ts` are kind of a lot but the big change aside from moving into being class-based validators is to support the new rule system for calling `this.reportWarning` or `this.reportError`.

packages/parser/test/vitest.matchers.ts
```diff
@@ -0,0 +1,119 @@
...
+  }): Promise<R>;
+}
+
+declare module 'vitest' {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  interface Assertion<T = any> extends CustomMatchers<T> {}
+  interface AsymmetricMatchersContaining extends CustomMatchers {}
+}
+
+export async function toValidate(
```
erunion, line 34: This matcher feels like a **lot** and I'm not in love with what I did here.

packages/parser/src/index.ts
```diff
@@ -129,25 +147,68 @@ export async function validate<S extends APIDocument = APIDocument>(
...
+}
 
-  return parser.schema;
+/**
+ * A utility to transform the `ValidationResult` from a `validate()` call to a human-readable
+ * string.
+ *
+ */
+export function compileErrors(result: ValidationResult): string {
+  const specName = !result.specification ? 'API definition' : result.specification;
```
kanadgupta, line 189:
```suggestion
  const specName = result.specification || 'API definition';
```

packages/parser/src/index.ts
```diff
@@ -129,25 +147,68 @@ export async function validate<S extends APIDocument = APIDocument>(
...
+  const status = !result.valid ? 'failed' : 'succeeded, but with warnings';
+
+  let message = `${specName} schema validation ${status}.\n`;
+  message += '\n';
```
kanadgupta, line 193:
```suggestion
  let message = `${specName} schema validation ${status}.\n\n`;
```

packages/parser/src/types.ts
```diff
@@ -8,6 +8,76 @@ export type APIDocument<T extends object = NonNullable<unknown>> =
...
+      additionalErrors: number;
+      specification: 'OpenAPI' | 'Swagger' | null;
+    }
+  | {
+      valid: true;
+      warnings: WarningDetails[];
+      specification: 'OpenAPI' | 'Swagger' | null;
+    };
+
+export interface ParserRulesOpenAPI {
```
kanadgupta, line 33: smol assertion to ensure that every rule conforms to this structure — feel free to take it or leave it

```suggestion
export interface ParserRulesOpenAPI extends Record<string, 'error' | 'warning'> {
```

packages/parser/src/index.ts
```diff
@@ -129,25 +147,68 @@ export async function validate<S extends APIDocument = APIDocument>(
...
+    if (result.valid === false && result.errors.length) {
+      message += '\n\nWe have also found some additional warnings:\n\n';
+    }
+
+    message += result.warnings.map(warn => warn.message).join('\n\n');
+  }
+
+  if (result.valid === false && result.additionalErrors > 0) {
+    message += `\n\nPlus an additional ${result.additionalErrors} errors. Please resolve the above and re-run validation to see more.`;
+  }
```
kanadgupta, line 211: i'm wondering if `message` should be a `string[]` and on line 213 below, we just return `message.join('\n\n')`? the whitespace stuff is a little hard to follow atm

packages/parser/test/specs/deep-circular/deep-circular.test.ts
```diff
@@ -37,23 +40,7 @@ describe('API with deeply-nested circular $refs', () => {
...
-      api.paths['/family-tree'].get.responses['200'].schema.properties.level1.properties.level2.properties.name.type,
-    );
-    expect(api.paths['/family-tree'].get.responses['200'].schema.properties.name.type).toStrictEqual(
-      api.paths['/family-tree'].get.responses['200'].schema.properties.level1.properties.level2.properties.level3
-        .properties.name.type,
-    );
-    expect(api.paths['/family-tree'].get.responses['200'].schema.properties.name.type).toStrictEqual(
-      api.paths['/family-tree'].get.responses['200'].schema.properties.level1.properties.level2.properties.level3
-        .properties.level4.properties.name.type,
-    );
```
kanadgupta, line 56: this coverage looks useful, could we keep it?

packages/parser/test/specs/large-file-memory-leak/large-file-memory-leak.test.ts
```diff
@@ -9,16 +8,20 @@ describe('Large file memory leak protection', { timeout: 20000 }, () => {
     ['cloudflare spec (stringified to a single line)', 'cloudflare-stringified.json'],
     ['cloudflare spec', 'cloudflare.json'],
   ])('%s', async (name, file) => {
-    try {
-      await validate(relativePath(`specs/large-file-memory-leak/${file}`));
+    const res = await validate(relativePath(`specs/large-file-memory-leak/${file}`));
+    if (res.valid === true) {
```
kanadgupta, line 12: this if-block seems redundant

packages/parser/test/specs/large-file-memory-leak/large-file-memory-leak.test.ts
```diff
@@ -9,16 +8,20 @@ describe('Large file memory leak protection', { timeout: 20000 }, () => {
...
+    expect(res.valid).toBe(false);
+    expect(res.errors).toHaveLength(20);
+    expect(res.errors).toStrictEqual(
+      expect.arrayContaining([
+        {
+          message: expect.stringContaining('4xx is not expected to be here!'),
+        },
+      ]),
+    );
+    expect(res.additionalErrors).toBe(1016);
```
kanadgupta, line 25: could we just make all of this into a snapshot test?

```suggestion
    expect(res).toMatchSnapshot();
```

packages/parser/test/specs/real-world/real-world.test.ts
```diff
@@ -19,23 +19,38 @@ describe(
...
+        if (api.name in knownErrors) {
+          assert.fail(`${api.name} had known errors that are no longer a problem. Please remove them from the list.`);
+        }
+      } else if (api.name in knownErrors) {
+        expect(result.errors).toHaveLength(knownErrors[api.name].total);
+        expect(result.errors).toStrictEqual(knownErrors[api.name].errors);
+        expect(result.warnings).toHaveLength(0);
+      } else {
+        // API is invalid and does not have any known errors.
+        expect(result).toBeUndefined();
```
kanadgupta, line 53: interesting, this isn't something we'd want to fail on?

kanadgupta (changes requested):
mostly small marbles stuff, looks great overall. excited to ship this!

packages/parser/README.md
```diff
@@ -73,14 +73,73 @@ Internally this method invokes [`dereference()`](#dereference) so the returned o
...
+By default `validate` returns a `ValidationResult` which will contain an array of errors, if you would like to convert this shape into a human-readable error string you can do so by utilizing our `compileErrors` utility:
+
+```ts
+import { validate, compileErrors } from '@readme/openapi-parser';
+
+const result = await validate(myAPI);
+if (result.valid) {
+  console.log('🍭 The API is valid!');
+} else {
+  console.error(compileErrors(result));
```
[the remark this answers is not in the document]
erunion: added in 9de52f5b

packages/parser/README.md
```diff
@@ -73,14 +73,73 @@ Internally this method invokes [`dereference()`](#dereference) so the returned o
...
 
-try {
-  const api = await validate(myAPI);
+const result = await validate(myAPI);
+if (result.valid) {
   console.log('🍭 The API is valid!');
-} catch (err) {
-  console.error(err);
+} else {
+  console.error(result.errors);
```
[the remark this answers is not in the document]
erunion: added in 9de52f5b

packages/parser/src/types.ts
erunion: b2f98cc4

packages/parser/src/index.ts
erunion: b2f98cc4

packages/parser/test/specs/large-file-memory-leak/large-file-memory-leak.test.ts
erunion: it's for type narrowing. i could move this over to a snapshot but with 20 errors we're asserting here that'd be a **huge** snapshot

packages/parser/test/specs/deep-circular/deep-circular.test.ts
erunion: `validate` no longer returns an API schema so i can't do these assertions anymore

kanadgupta: oh, duh

packages/parser/test/specs/real-world/real-world.test.ts
erunion: yeah this assertion would trigger a failure because `result` won't ever be undefined.

packages/parser/test/specs/unknown/unknown.test.ts
```diff
@@ -31,18 +34,7 @@ describe('API with $refs to unknown file types', () => {
...
-    expect(api.paths['/files/text'].get.responses['200'].schema.default).toStrictEqual(
-      dereferencedAPI.paths['/files/text'].get.responses['200'].schema.default,
-    );
-    expect(api.paths['/files/html'].get.responses['200'].schema.default).toStrictEqual(
-      dereferencedAPI.paths['/files/html'].get.responses['200'].schema.default,
-    );
-    expect(api.paths['/files/blank'].get.responses['200'].schema.default).toStrictEqual(
-      dereferencedAPI.paths['/files/blank'].get.responses['200'].schema.default,
-    );
-    expect(api.paths['/files/binary'].get.responses['200'].schema.default).toBeInstanceOf(Buffer);
```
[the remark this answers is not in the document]
erunion: `validate` doesn't return a schema anymore

packages/parser/src/index.ts
```diff
@@ -1,16 +1,14 @@
...
 import { isSwagger, isOpenAPI } from './lib/index.js';
 import { convertOptionsForParser, normalizeArguments, repairSchema } from './util.js';
 import { validateSchema } from './validators/schema.js';
 import { validateSpec } from './validators/spec.js';
 
-export type { ParserOptions };
-
-export { ValidationError };
+export type { ParserOptions, ValidationResult, ErrorDetails, WarningDetails };
+export { ValidationError } from './errors.js';
```
kanadgupta, line 11: this doesn't look used now, can we delete it?

packages/parser/test/specs/real-world/real-world.test.ts
kanadgupta: oh duh

kanadgupta (approved):
last piece of feedback, otherwise LGTM!

packages/parser/src/index.ts
kanadgupta: we don't actually use it ourselves though and it's basically just the `Error` object with a different name

kanadgupta: i'm worried it might mislead folks into thinking it's a type that these functions actually return

erunion: `oas-normalize` uses it. should i just move it to that library instead?

erunion: yeah ill move it

kanadgupta: ohhh i see. yeah i'd prefer that, ty

erunion: moved in 564b4a09

packages/parser/package.json
```diff
@@ -11,6 +11,10 @@
       "require": "./dist/index.cjs",
       "import": "./dist/index.js"
     },
+    "./lib/errors": {
+      "require": "./dist/lib/errors.cjs",
+      "import": "./dist/lib/errors.js"
+    },
```
kanadgupta, line 17: this should be in `oas-normalize` right? wish knip caught this 😕

kanadgupta (changes requested):

packages/parser/package.json
erunion: `attw` catches it but we aren't running that in CI. will PR that after this

erunion: actually nvm we can't run `attw` in CI because of the node10 stuff and them not having a way to suppress errors from it.

Screenshot 2025-03-04 at 10 52 44 AM

kanadgupta: wait! i discovered a way we can support `node10` — check out how `swr` does it for `swr/infinite`: https://github.com/vercel/swr/blob/main/infinite/package.json

kanadgupta: i've been meaning to introduce this into our repos but i keep forgetting

kanadgupta: i think we can do what they do but omit the `types` field since that can be inferred thanks to `tsup`

kanadgupta (approved):

PR #948, opened 2025-03-04
ci: running `attw` as part of our ci processes

erunion:
## 🧰 Changes

This spins up `attw` as part of our build processes for asserting that the code we have set up to export in our `package.json` files is set up to function in Node.


.github/workflows/ci.yml
```diff
@@ -36,9 +36,22 @@ jobs:
...
+        with:
+          node-version: lts/*
+      - run: npm ci
+      - run: npm run attw
+
   lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
erunion, line 53: very surprised that linting and npm was working without this. i guess all boxes have node by default?

packages/oas-normalize/package.json
```diff
@@ -52,7 +52,7 @@
     "url": "https://github.com/readmeio/oas/issues"
   },
   "scripts": {
-    "attw": "attw --pack --format table-flipped",
+    "attw": "attw --pack --format table-flipped --profile node16",
```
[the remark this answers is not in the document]
kanadgupta: hell yeah this rules

.github/workflows/ci.yml
kanadgupta: yeah i don't ever do this in linting/etc steps b/c the node.js version isn't really a factor here, but up to you on if we should leave this in. the actions runner ships with an LTS version by default: https://github.com/actions/runner-images/blob/ubuntu24/20250223.1/images/ubuntu/Ubuntu2404-Readme.md

packages/oas-to-har/package.json
```diff
@@ -39,7 +39,7 @@
     "url": "https://github.com/readmeio/oas/issues"
   },
   "scripts": {
-    "attw": "attw --pack --format table-flipped",
+    "attw": "attw --pack --format table-flipped --profile node16",
```
kanadgupta, line 42: i'd probably remove these `--format` flags and defer to their auto-formatting, the output in github actions looks kinda screwy (and it looks even worse on narrower screens):

CleanShot 2025-03-04 at 14 14 03@2x


```suggestion
    "attw": "attw --pack --profile node16",
```

kanadgupta (approved):

packages/oas-to-har/package.json
erunion: changing this to `ascii` because they default to `auto` which still ends up using the table format on libraries that just have a single or two exports. makes the whole output look inconsistent with some having a table and others being a raw dump.

PR #949, opened 2025-03-04
fix: getting `jest-expect-har` and `jest-expect-openapi` functional

erunion:
## 🧰 Changes

I pulled in our `jest-expect-har` and `jest-expect-openapi` libraries into this repository today but without some work they were non-functional. This gets them working again and also updates `jest-expect-openapi` to support our new OpenAPI parser in `packages/parser`.


knip.ts
```diff
@@ -1,13 +1,15 @@
 import type { KnipConfig } from 'knip';
 
 const config: KnipConfig = {
-  // `attw` is used to ensure that our dists have proper typings.
-  ignoreBinaries: ['attw'],
+  ignore: ['packages/jest-expect-*/jest.config.js'],
```
erunion, line 4: Knip isn't picking up on the Jest configs being used in our two packages because (I believe) `jest` is installed as a peer dep.

knip.ts
```diff
@@ -1,13 +1,15 @@
 import type { KnipConfig } from 'knip';
 
 const config: KnipConfig = {
-  // `attw` is used to ensure that our dists have proper typings.
-  ignoreBinaries: ['attw'],
```
erunion, line 5: `attw` is a dev dep in the root now so this is unneeded.

package.json
```diff
@@ -14,7 +14,7 @@
     "publish": "npx lerna publish",
     "prelint": "npm run build",
     "pretest": "npm run build",
-    "test": "vitest --coverage"
+    "test": "vitest --coverage && npx lerna run test:jest --stream"
```
erunion, line 17: I hate this but these two libraries need to be run through Jest to ensure that that works even though Vitest actually uses Jest's `expect` library under the hood.

packages/jest-expect-har/.eslintrc
```diff
@@ -1,13 +1,4 @@
 {
-  "extends": [
-    "@readme/eslint-config",
-    "@readme/eslint-config/typescript"
-  ],
+  "extends": ["@readme/eslint-config", "@readme/eslint-config/typescript"],
   "root": true,
```
erunion, line 3: Retaining `root: true` in these two packages because the root loads in our ESM config and neither of these packages can be ESM because of Jest.

packages/jest-expect-har/__tests__/vitest.test.ts
erunion, line 1: Merged the two tests into a single file like `jest-expect-openapi` does.

packages/jest-expect-har/package.json
erunion, line 1: Lot of noise updates in these package files but I basically just sorted everything to match how the other packages are. My usual tidyness bullshit.

packages/jest-expect-openapi/src/index.ts
```diff
@@ -48,9 +48,14 @@ export default async function toBeAValidOpenAPIDefinition(
     cloned = transformer(cloned);
   }
 
-  const { pass, error } = await openapiParser
-    .validate(cloned)
-    .then(() => ({ pass: true, error: null }))
+  const { pass, error } = await validate(cloned)
```
erunion, line 51: new parser usage

kanadgupta (approved):

PR #950, opened 2025-03-04
chore: getting the `oas-examples` package working again

erunion:
## 🧰 Changes

Like with https://github.com/readmeio/oas/pull/949 this gets our `oas-examples` package working again after I migrated it into this repository from https://github.com/readmeio/oas-examples


.github/workflows/validate.yml
```diff
@@ -24,29 +24,29 @@ jobs:
         with:
           # See our `rdme` docs for more GitHub Actions usage examples:
           # https://docs.readme.com/docs/rdme
-          rdme: openapi:validate 2.0/json/petstore.json
+          rdme: openapi:validate 2.0/json/petstore.json --workingDirectory=packages/oas-examples
```
erunion, line 27: merging this all in as-is but fyi @kanadgupta in case you have a better way to do this

PR #953, opened 2025-03-05
feat: dropping support for node 18

erunion:
## 🧰 Changes

Node 18 hits EOL at the end of next month and with the upcoming ESM improvements in Node 20+, with which we'll be able to simplify our build processes, we're getting ahead of the gate and dropping Node 18 now.


kanadgupta (approved):
lol we haven't been running tests against node 18 for a while anyways. ope

PR #958, opened 2025-03-06
feat(jest-expect-har): support for CJS and ESM builds

erunion:
## 🧰 Changes

This updates `jest-expect-har` to support CJS and ESM dists, as well as consolidating its various TS configs to use the core one in this repository.

### DIFF
--- a/packages/jest-expect-har/package.json
+++ b/packages/jest-expect-har/package.json
@@ -4,8 +4,17 @@
   "description": "Jest/Vitest matcher for asserting valid HAR definitions",
   "license": "MIT",
   "author": "Jon Ursenbach <jon@readme.io>",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "sideEffects": false,
+  "type": "module",
+  "exports": {
+    ".": {
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
+  },
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
   "engines": {
     "node": ">=20"
   },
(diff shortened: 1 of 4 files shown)


packages/jest-expect-har/package.json
```diff
@@ -4,8 +4,17 @@
...
+  "type": "module",
+  "exports": {
+    ".": {
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
+  },
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
```
erunion, line 17: I don't _think_ that this is a breaking release since these all should be backwards compatible but happy to tag another major on this library if you think its worth it.

kanadgupta: if it’s a single default export then yeah probably not breaking

packages/jest-expect-har/package.json
```diff
@@ -33,7 +42,8 @@
     "url": "https://github.com/readmeio/oas/issues"
   },
   "scripts": {
-    "build": "tsc",
+    "attw": "attw --pack --format ascii --profile node16",
```
kanadgupta, line 45: would it pass without this flag? might be a good check to have to ensure this is a non-breaking change

```suggestion
    "attw": "attw --pack --format ascii",
```

packages/jest-expect-har/test/tsconfig.json
```diff
@@ -1,9 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "module": "es6",
-    "moduleResolution": "node",
-    "strict": false
+    "resolveJsonModule": true
```
kanadgupta, line 4: huh kinda surprising this isn’t in our base config

erunion: we only ever load JSON files in tests

packages/jest-expect-har/package.json
erunion: yep `node10` still works, will leave `--profile` off

packages/jest-expect-har/package.json
erunion: confirmed `node10` still works so im not going to treat this as a breaking release

PR #959, opened 2025-03-06
feat(jest-expect-openapi): support for CJS and ESM builds

erunion:
## 🧰 Changes

Like in https://github.com/readmeio/oas/pull/958 this updates `jest-expect-openapi` to support CJS and ESM dists, as well as consolidating its various TS configs to use the core one in this repository.

### DIFF
--- a/packages/jest-expect-openapi/package.json
+++ b/packages/jest-expect-openapi/package.json
@@ -4,8 +4,17 @@
   "description": "Jest/Vitest matcher for asserting valid OpenAPI definitions",
   "license": "MIT",
   "author": "Jon Ursenbach <jon@readme.io>",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "sideEffects": false,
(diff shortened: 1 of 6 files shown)


packages/jest-expect-openapi/package.json
```diff
@@ -4,8 +4,17 @@
   "description": "Jest/Vitest matcher for asserting valid OpenAPI definitions",
   "license": "MIT",
   "author": "Jon Ursenbach <jon@readme.io>",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "sideEffects": false,
```
erunion, line 7: Will bump this to a major release if you also think https://github.com/readmeio/oas/pull/958#discussion_r1982408949 should be too.

erunion: confirmed `node10` still works so im not going to treat this as a breaking release

PR #960, opened 2025-03-06
feat(har-examples): dual ESM and CJS build support

erunion:
## 🧰 Changes

I would like to enable `strict` and `noImplicitAny` across this repository but this `har-examples` package of ours has no typings so it always gets picked up by TS as not being able to be imported unless we do a module declaration.

This resolves all of that by migrating the library to full TS, with actual Har typing enforcement, and a dual CJS and ESM build process.


packages/har-examples/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  | 'multipart-form-data'
+  | 'query-encoded'
+  | 'query'
+  | 'short'
+  | 'text-plain'
+  | 'xml',
+  Har
+>;
+
+const examples: Examples = {
```
erunion, line 49: This `Examples` type I had to do is nuts but `isolatedDeclarations` loses its marbles without it:

Screenshot 2025-03-05 at 6 29 11 PM

erunion: Alternatively, if I do ` satisfies Record<string, Har> as Record<string, Har>` like it suggests then the export loses all type inference of what is actually in the `examples` object. Feels like a lose-lose.

erunion: not sure if i _like like_ that, but i like that.

kanadgupta: yeahhh the ergonomics of `isolatedDeclarations` is still pretty rough atm. hopefully it's worth it 🫠

package-lock.json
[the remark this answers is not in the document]
erunion: its because `tsup` got bumped to 8.4.0

packages/har-examples/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import multipart_file from './multipart-file.har.js';
+import multipart_form_data from './multipart-form-data.har.js';
+import query_encoded from './query-encoded.har.js';
+import query from './query.har.js';
+import short from './short.har.js';
+import text_plain from './text-plain.har.js';
+import xml from './xml.har.js';
+
+const examples = {
+  'application-form-encoded': application_form_encoded satisfies Har as Har,
```
kanadgupta, line 26: it's still rather ugly but i think this is a step in the right direction and i'm actually quite glad TS suggested this:

CleanShot 2025-03-06 at 12 00 13@2x

with `satisfies T as T` we can at least assert that the object in question meets the requirements of `T` before setting it with `as`

kanadgupta (approved):

PR #971, opened 2025-04-07
fix(oas-normalize): `enablePaths` wasn't properly being configured

erunion:
## 🧰 Changes

This resolves a problem in `oas-normalize` where the `enablePaths` option wasn't being supplied down into `@readme/openapi-parser`.


kellyjosephprice (approved):
Not familiar with this repo, but the test looks good.

PR #973, opened 2025-04-14
fix(parser): internal cache improvements to dereferencing circular refs

erunion:
## 🧰 Changes

We uncovered a problem with the dereferencing functionality within our OpenAPI parser on schemas with an extensive amount of circular references where under these circumstances it could take upwards to an hour to dereference a schema. I was able to trace it down to being a problem within [@apidevtools/json-schema-ref-parser](https://npm.im/@apidevtools/json-schema-ref-parser) where because it was not always using its internal cache it would get into situations where, when parsing a circular reference, it would often attempt to re-parse that same reference multiple times.

I submitted a fix for this over in https://github.com/APIDevTools/json-schema-ref-parser/pull/380, and this pulls that in along with builds a unit test around one of these problematic schemas.

### DIFF
--- a/packages/oas/test/analyzer/queries/openapi.test.ts
+++ b/packages/oas/test/analyzer/queries/openapi.test.ts
@@ -89,13 +89,8 @@ describe('analyzer queries (OpenAPI)', () => {
   describe('circularRefs', () => {
     it('should determine if a definition has circular refs when it does', async () => {
       await expect(QUERIES.circularRefs(circular)).resolves.toStrictEqual([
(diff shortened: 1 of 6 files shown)


packages/oas/test/analyzer/queries/openapi.test.ts
erunion, line 1: Because of the cache improvements I did in `@apidevtools/json-schema-ref-parser` we are now calling `onCircular` less times in our parser because we are no longer processing schemas that have already been processed.

PR #974, opened 2025-04-16
feat(extensions): add support for template engine in code samples

anandkumarpatel:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

- Introduced a new optional property `templateEngine` in the Extensions interface to allow dynamic code samples.
- Added a new templated code snippet example using Mustache in both JSON and YAML readme files for better demonstration of the feature.

for: https://github.com/readmeio/readme/pull/14297
## 🧬 QA & Testing

* updated test fixtures

### DIFF
--- a/packages/oas-examples/3.0/json/readme-extensions.json
+++ b/packages/oas-examples/3.0/json/readme-extensions.json
@@ -145,6 +145,12 @@
               "code": "import something from 'requestidk';\n\nconsole.log('do something idk');",
               "install": "npm install requestidk",
               "correspondingExample": "curlExample"
+            },
+            {
+              "name": "Templated code snippet",
+              "language": "node",
+              "templateEngine": "mustache",
+              "code": "sdk.put().data({{#harRequest.postData.jsonObj}}{\n{{/harRequest.postData.jsonObj}}{{#harRequest.postData.jsonObj.name}}  name: {{harRequest.postData.jsonObj.name}}\n{{/harRequest.postData.jsonObj.name}}{{#harRequest.postData.jsonObj.id}}  id: {{harRequest.postData.jsonObj.id}}\n{{/harRequest.postData.jsonObj.id}}{{#harRequest.postData.jsonObj}}\n}{{/harRequest.postData.jsonObj}})"
             }
           ],
           "samples-languages": ["shell", "python", "node"]
--- a/packages/oas/src/extensions.ts
+++ b/packages/oas/src/extensions.ts
@@ -228,6 +228,11 @@ export interface Extensions {
      * @example "Custom cURL snippet"
      */
     name?: string;
+    /**
+     * Optional template engine for dynamic code samples
+     * @example "mustache"
+     */
+    templateEngine?: string;
   }[];
   [DISABLE_TAG_SORTING]: boolean;
   [EXPLORER_ENABLED]: boolean;
--- a/packages/oas/src/extensions.ts
+++ b/packages/oas/src/extensions.ts
@@ -232,7 +232,7 @@ export interface Extensions {
      * Optional template engine for dynamic code samples
      * @example "mustache"
      */
-    templateEngine?: string;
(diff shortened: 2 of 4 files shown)


packages/oas/src/extensions.ts
```diff
@@ -228,6 +228,11 @@ export interface Extensions {
      * @example "Custom cURL snippet"
      */
     name?: string;
+    /**
+     * Optional template engine for dynamic code samples
+     * @example "mustache"
+     */
+    templateEngine?: string;
```
anandkumarpatel, line 235: debating making this an enum / hardcode values. thoughts?

erunion: Definitely make this an enum.

erunion (changes requested):

packages/oas/src/extensions.ts
anandkumarpatel: done!

packages/oas-examples/3.0/json/readme-extensions.json
```diff
@@ -145,6 +145,12 @@
               "code": "import something from 'requestidk';\n\nconsole.log('do something idk');",
               "install": "npm install requestidk",
               "correspondingExample": "curlExample"
+            },
+            {
+              "name": "Templated code snippet",
+              "language": "node",
+              "templateEngine": "mustache",
```
kanadgupta, line 152: sorry i'm coming in a little late with this question, but is there a reason we're going with mustache here? for context, readme already has a few templating engines in place:

- jade/pug in our main app (though a lot less since we're trying to move away from this)
- liquid in our legacy global landing page ([see docs](https://docs.readme.com/ent/docs/creating-a-global-landing-page#/))
- JSX in our main app
- MDX (used by our customers going forward)

i would much rather use one of these existing ones (with a strong preference towards JSX or MDX so we have the ability to add type safety around these template variables) rather than introducing a new templating engine + dependency into the mix!

anandkumarpatel: I just picked mustache because it was one that one of our partners recommended and after review, it seemed to be the "safest" because it did not allow any executions. 

I would actually prefer MDX but didn't think it would work here will look into that.

rafegoldberg: Yeah, I agree with @anandkumarpatel: MDX feels like overkill for this scenario. I'm…_fine_ with Mustache, but if we want to stick with one of our current templating languages I'd probably have gone with Liquid, since it's relatively lightweight.

rafegoldberg: MDX it comes with a lot of server side security implications and is pretty inefficient just for a single code snippet, I’d imagine?

kanadgupta: yeah i would only suggest this if we could do something lightweight like `ReactDOMServer.renderToString` here... tbh i'd even prefer liquid just b/c we're adding the `mustache` dep to our main app just for the sake of supporting this and i don't _love_ that

kanadgupta: the mustache snippets are SSR'd right? no rendering logic is executed place on the client?

kanadgupta: honestly thinking about it more, i'm less tied to JSX/MDX given the fact that this is all being written as JSON strings which makes the DX around writing these snippets less than ideal regardless. cool with whatever you run with here, but if i had to choose between mustache and liquid i'd probably go with liquid!

anandkumarpatel: I think i'll close this for now because we already have partner committed to writing `httpSnippet Plugin` but... they are going to be able to import any library, so we will have a little less control on what they use so might re-open in the future.

PR #992, opened 2025-08-15
chore: upgrading our coding standards

erunion:
## 🧰 Changes

Biome [shipped a pretty large release earlier this week](https://github.com/biomejs/biome/releases/tag/%40biomejs%2Fbiome%402.2.0) so this upgrades us, and our adjoining shared Biome config in `@readme/standards`, to support it.

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -15,9 +15,6 @@
     },
     "rules": {
       "recommended": true,
-      "correctness": {
-        "useJsonImportAttributes": "error",
-      },
     },
   },
   "overrides": [
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -1,6 +1,6 @@
 {
   "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
-  "extends": ["@readme/standards/biome"],
+  "extends": ["@readme/standards/biome", "@readme/standards/biome/esm"],
   "files": {
     "includes": [
       // This array is extended in each package. If we don't include this then Biome will scan
@@ -15,9 +15,21 @@
     },
     "rules": {
       "recommended": true,
-      "nursery": {
-        "useJsonImportAttribute": "error",
+      "correctness": {
+        "useJsonImportAttributes": "error",
       },
     },
   },
+  "overrides": [
+    {
+      "includes": ["**/tsup.config.ts"],
+      "linter": {
+        "rules": {
+          "style": {
+            "noDefaultExport": "off",
+          },
+        },
+      },
+    },
+  ],
 }
--- a/packages/oas-normalize/src/index.ts
+++ b/packages/oas-normalize/src/index.ts
@@ -19,6 +19,7 @@ import {
   stringToJSON,
 } from './lib/utils.js';
 
+// biome-ignore lint/style/noDefaultExport: This is fine for now.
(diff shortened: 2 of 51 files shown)


packages/oas-normalize/src/index.ts
```diff
@@ -19,6 +19,7 @@ import {
   stringToJSON,
 } from './lib/utils.js';
 
+// biome-ignore lint/style/noDefaultExport: This is fine for now.
```
erunion, line 22: Every place I'm suppressing this would be a breaking change that I don't really feel like undertaking right now because upgrading `oas` a major version across our stack is a real pain and involves bumping `api` along the way.

kanadgupta: agreed w this assessment

packages/oas-to-har/src/lib/style-formatting/index.ts
```diff
@@ -225,7 +225,7 @@ function shouldExplode(parameter: ParameterObject) {
   );
 }
 
-export default function formatStyle(value: unknown, parameter: ParameterObject): any {
+export function formatStyle(value: unknown, parameter: ParameterObject): any {
```
kanadgupta, line 228: just checked and this one is fine, but we just wanna be careful that we're not making this change in any file that's exposed in [our `exports` field in the `package.json`](https://github.com/readmeio/oas/blob/4bccb292663cb3945062633cb48f66aa57a8dedd/packages/oas-to-har/package.json#L9) since that would be a breaking change (per what you mentioned [here](https://github.com/readmeio/oas/pull/992/files#r2280009596))

packages/oas-to-har/src/index.ts
```diff
@@ -547,7 +548,7 @@ export default function oasToHar(
                   jsonTypes.forEach((prop: boolean | string) => {
                     try {
                       set(cleanBody, String(prop), JSON.parse(get(cleanBody, String(prop))));
-                    } catch (e) {
```
kanadgupta, line 550: still blown away that we've been able to do this this entire time lmao

packages/oas/src/analyzer/index.ts
kanadgupta, line 1: woof this must have been cumbersome but ty for this

biome.jsonc
```diff
@@ -15,9 +15,21 @@
     },
     "rules": {
       "recommended": true,
-      "nursery": {
-        "useJsonImportAttribute": "error",
+      "correctness": {
+        "useJsonImportAttributes": "error",
```
kanadgupta, line 19: i think we should add this to our ESM config (if not the main one)

kanadgupta (approved):
one comment re: the `useJsonImportAttributes` rule that might be nice to address before merge but no blocking thoughts otherwise

packages/oas-to-har/src/lib/style-formatting/index.ts
erunion: yeah i definitely checked that before making any of these default changes i did

biome.jsonc
erunion: Oh this is actually part of the `project` domain now and is enabled by default in all of our shared configs. Will remove this.

PR #998, opened 2025-09-10
feat(oas-to-har): retain null values in array

eaglethrost:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

This PR is part of the work for the readme ticket [RM-9906 Not displaying null values in an array](https://linear.app/readme-io/issue/RM-9906/not-displaying-null-values-in-an-array) and should get merged ONLY after the corresponding PR changes in the `remove-undefined-objects` repo is shipped

The aim of the ticket is we want user-added null values in arrays to be visible in requests.

This requires adjustments in the oas-to-har package, with two separate causes of why the nulls in array are being dropped
1. For postData (body):
- Issue: remove-undefined-objects was filtering out null values in arrays.
- Fix: Replaced JSON.parse/stringify approach with a custom function to distinguish actual null vs. undefined.
   Updated stripEmptyObjects to stop removing null values in arrays, while still removing undefined, empty objects, and   empty arrays.
2. For Non-postData (parameters, cookies, etc.):
- Issue: encodeArray in style-serializer used .join, which dropped null values.
- Fix: Convert null values to the string "null" before joining, so they’re preserved.

TODO Before Merging:
- [ ] Update the remove-undefined-object package version once this PR is shipped: [https://github.com/readmeio/remove-undefined-objects/pull/149](https://github.com/readmeio/remove-undefined-objects/pull/149)

## 🧬 QA & Testing

- Temporarily migrated tests for remove-undefined-objects to show the changes works
- Run oasToHar or oasToSnippet in a script with an OAS object containing arrays of null, undefined, and undefined objects


packages/oas-to-har/src/lib/remove-undefined-objects.ts
erunion, line 1: why are we vendoring and forking remove-undefined-objects in here instead of adding a new option to that library for preserving empty arrays? thats what @mjcuva and i talked about out what i thought needed to happen for this ticket

eaglethrost: Sorry this PR is not for that [ticket](https://linear.app/readme-io/issue/CX-1151/add-support-for-allowemptyvalue-for-non-primitive-types), but a different [one](https://linear.app/readme-io/issue/RM-9906/not-displaying-null-values-in-an-array), which tries to solve a different issue of null values in arrays getting dropped in the request snippet. 

Still working on the ticket you & marc mentioned, I'll create another PR's for it.

kanadgupta (changes requested):
can you use the latest release of [`remove-undefined-objects@7.0.0`](https://github.com/readmeio/remove-undefined-objects/releases/tag/7.0.0) here?

erunion (approved):

PR #1000, opened 2025-09-20
fix(oas): server URL mapping edge case

kanadgupta:
## 🧰 Changes

jack from lyra (via @anandkumarpatel) pointed out a little edge case that we weren't properly accounting for in our server variable matching logic. this PR fixes that and adds a little test case just in case.

### DIFF
--- a/packages/parser/test/specs/real-world/real-world.test.ts
+++ b/packages/parser/test/specs/real-world/real-world.test.ts
@@ -32,7 +32,11 @@ describe(
         });
       } catch (err) {
         // If we have errors pulling the API definition down then don't fail out.
-        if (err.message.includes('Error downloading https://') || err.message.includes('socket hang up')) {
+        if (
+          err.message.includes('Error downloading https://') ||
+          err.message.includes('Error reading file "https://') ||
+          err.message.includes('socket hang up')
+        ) {
           return;
         }
 


packages/oas/test/index.test.ts
```diff
@@ -1322,6 +1322,36 @@ describe('Oas', () => {
         expect(operation).toBeUndefined();
       });
 
+      it('should be able to find a match on a url with a non-default port', () => {
```
anandkumarpatel, line 1325: Do you mind adding a few more hard test here ? I.E.  a file that has 2 servers and like 3 different vars? One in domain, on for port, and one for path

anandkumarpatel (changes requested):

packages/oas/test/index.test.ts
kanadgupta: we do have a lot of those kinds of tests in this file already, feel free to take a look in these places and let me know if you can think of any gaps:

https://github.com/readmeio/oas/blob/f2d3b32595993807e4f65232dd288adec6310871/packages/oas/test/index.test.ts#L867-L945 

https://github.com/readmeio/oas/blob/f2d3b32595993807e4f65232dd288adec6310871/packages/oas/test/index.test.ts#L1242-L1397

i'm truthfully still kinda stumped how the example you provided wasn't working given how thorough our tests already are 🧐

anandkumarpatel: ahh perf!

anandkumarpatel (approved):

anandkumarpatel:
@kanadgupta are the deps we use for the test gone now?

kanadgupta:
@anandkumarpatel not sure what you mean!

kanadgupta:
> do we want to bring those into this repo so it does not have to make network request in the test? I fear depending on a 3rd party network request makes these test flaky?

@anandkumarpatel ah yeah good question. digging into this a bit i think the network request is technically part of the test since the parser has built-in fetching but agreed that it could be a source of flakiness. i don't really care to revisit that whole process rn (and this would be good to discuss with jon when he's back) but in the interim i opened https://github.com/readmeio/oas/pull/1001 which should address the flakiness!

PR #1002, opened 2025-10-01
feat(pkce): add usePkce custom extension

maximilianfalco:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

This PR adds support for PKCE (Proof Key for Code Exchange) in OAuth authorization code flows by introducing a new `usePkce` property within the `oauth-options` extension.

- Added `usePkce?: boolean` property to the `OAUTH_OPTIONS` extension interface
- When enabled, OAuth authorization code flows should use PKCE with auto-generated code verifier and challenge
- When disabled (default), uses standard OAuth 2.0 authorization code flow with client credentials

### Usage:
```json
{
  "x-readme": {
    "oauth-options": {
      "usePkce": true
    }
  }
}
```

## 🧬 QA & Testing

- Added unit tests for root-level and operation-level `usePkce` configuration
- Added tests for multiple OAuth options working together
- All existing extension tests continue to pass

### DIFF
--- a/packages/oas/src/extensions.ts
+++ b/packages/oas/src/extensions.ts
@@ -104,7 +104,8 @@ export const METRICS_ENABLED = 'metrics-enabled';
  *  "x-readme": {
  *    "oauth-options": {
  *      "scopeSeparator": ",",
- *      "useInsecureClientAuthentication": true
+ *      "useInsecureClientAuthentication": true,
+ *      "usePkce": false
  *    }
  *  }
  * }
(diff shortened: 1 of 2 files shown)


packages/oas/src/extensions.ts
```diff
@@ -257,6 +258,19 @@ export interface Extensions {
...
+     * The client secret is replaced with an auto-generated code verifier and code challenge.
+     * When disabled, uses the standard OAuth 2.0 authorization code flow with client credentials.
+     *
+     * @see {@link https://datatracker.ietf.org/doc/html/rfc7636#section-4.1}
+     * @see {@link https://datatracker.ietf.org/doc/html/rfc7636#section-4.2}
+     *
+     * @example true
+     * @default false
+     */
+    usePkce?: boolean;
```
erunion, line 273: Can this be renamed to `usePKCE`?

erunion (changes requested):
very minor request but otherwise lgtm

packages/oas/src/extensions.ts
maximilianfalco: I specifically named it to `usePkce` here because we also wanted to piggy back off of redocly's [custom extension](https://redocly.com/docs-legacy/api-reference-docs/specification-extensions/x-use-pkce). (see linear thread [here](https://linear.app/readme-io/issue/RM-10969/pkce-support#comment-0d20338c))

the difference here was that redocly stores their flag within the oauth authorization code flow object while we plan to store them globally in `x-readme.oauth_options`. the frontend also checks them both (in the `x-readme` custom extension first and then the `x-usePkce` from redocly) so I opted to store them in the same name, hence I went for `usePkce` so that we dont have both `usePKCE` and `usePkce` in our codebase

the frontend component that uses these flags can be found here:
https://github.com/readmeio/readme/blob/edf513ee012e327a5c7149af90eb20b54a7d7e0c/packages/react/src/ui/API/Auth/LegacyAuth/components/input.tsx#L275-L282

maximilianfalco: on that note, we should also probably update our docs to add a section for `usePkce` and `PKCE` support in general

erunion (approved):

PR #1009, opened 2025-10-24
feat(parser): improved error message when paths are missing a leading slash

maximilianfalco:
| 🚥 Resolves CX-2158 |
| :------------------- |

## 🧰 Changes

This PR aims to show a better and more informative error message in cases where OAS paths have no leading slash (`/`)

Before | After
-- | --
Screenshot 2025-10-24 at 19 14 05 | Screenshot 2025-10-28 at 15 27 57

## 🧬 QA & Testing

Checkout the linked linear ticket, steps to repro this bug is placed there.

### DIFF
--- a/packages/parser/src/lib/detectNoSlashPaths.ts
+++ b/packages/parser/src/lib/detectNoSlashPaths.ts
@@ -7,11 +7,11 @@ import type { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
  * @param api - The API definition to check
  * @returns Array of paths that are missing leading slashes
  */
-export function detectNoSlashPaths(api: OpenAPIV2.Document | OpenAPIV3_1.Document | OpenAPIV3.Document): string[] {
+export function hasInvalidPaths(api: OpenAPIV2.Document | OpenAPIV3_1.Document | OpenAPIV3.Document): boolean {
   if (!api.paths || typeof api.paths !== 'object' || Array.isArray(api.paths)) {
     return [];
   }
 
   // Return paths that do not start with a leading slash
-  return Object.keys(api.paths).filter(path => !path.startsWith('/'));
+  return Object.keys(api.paths).some(path => !path.startsWith('/'));
 }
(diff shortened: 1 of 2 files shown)


packages/parser/src/lib/detectNoSlashPaths.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export function detectNoSlashPaths(api: OpenAPIV2.Document | OpenAPIV3_1.Document | OpenAPIV3.Document): string[] {
+  const noSlashPaths: string[] = [];
+
+  if (!api.paths) {
+    return noSlashPaths;
+  }
+
+  // Check each path in the paths object
+  Object.keys(api.paths).forEach(path => {
```
erunion: As an additional level of protection you might want to only run `Object.keys()` if `api.paths` is an object. If they've given us anything else for this, like a boolean or an array, whatever, we should avoid crashing here and let the AJV parsing handle and return an appropriate error back to them.

packages/parser/src/lib/detectNoSlashPaths.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @returns Array of paths that are missing leading slashes
+ */
+export function detectNoSlashPaths(api: OpenAPIV2.Document | OpenAPIV3_1.Document | OpenAPIV3.Document): string[] {
+  if (!api.paths || typeof api.paths !== 'object' || Array.isArray(api.paths)) {
+    return [];
+  }
+
+  // Return paths that do not start with a leading slash
+  return Object.keys(api.paths).filter(path => !path.startsWith('/'));
+}
```
kanadgupta: might want to also rename the file accordingly

```suggestion
export function hasInvalidPaths(api: OpenAPIV2.Document | OpenAPIV3_1.Document | OpenAPIV3.Document): boolean {
  if (!api.paths || typeof api.paths !== 'object' || Array.isArray(api.paths)) {
    return [];
  }

  // Return paths that do not start with a leading slash
  return Object.keys(api.paths).some(path => !path.startsWith('/'));
}
```

kanadgupta (changes requested):
sorry one more piece of feedback — i don't think we need to scan the whole `paths` array, we can just find the first invalid one and call it a day!

PR #1010, opened 2025-10-28
feat(parser): support for downgrading Swagger validation errors to warnings

erunion:
## 🧰 Changes

We have a number of Swagger API definitions which contain issues that we have begun to consider as fixable warnings in OpenAPI files. Because we have a method for downgrading these validation errors to warnings for OpenAPI definitions and not Swagger this ports that work over to support it.

### DIFF
--- a/packages/parser/src/validators/spec.ts
+++ b/packages/parser/src/validators/spec.ts
@@ -11,6 +11,9 @@ import { SwaggerSpecificationValidator } from './spec/swagger.js';
  * Validates either a Swagger 2.0 or OpenAPI 3.x API definition against cases that aren't covered
  * by their respective JSON Schema definitions.
  *
+ * Some specification-level cases can be treated as warnings, instead of hard validation erros, by
+ * supplying a `rules` configuration to the parsers `validate` option.
+ *
  */
 export function validateSpec(
   api: OpenAPIV2.Document | OpenAPIV3_1.Document | OpenAPIV3.Document,


packages/parser/README.md
```diff
@@ -160,17 +163,18 @@ if (result.valid) {
...
-| `path-parameters-not-in-path` | Path parameters defined in `parameters` must also be specified in the path URI with path templating. |
+| Rule | Description | Supports OpenAPI? | Supports Swagger? |
+| :--- | :--- | :--- | :--- |
+| `array-without-items` | Schemas that are defined as `type: array` must also have an `items` schema. | ✓ | ✓ |
+| `duplicate-non-request-body-parameters` | Parameters must be unique. | ✓ | ✓ |
+| `duplicate-operation-id` | The `operationId` definition in a path object must be unique. | ✓ | ✓ |
+| `non-optional-path-parameters` | Parameters that are defined within the path URI must be specified as being `required`. | ✓ | ✓ |
+| `path-parameters-not-in-parameters` | Path parameters defined in a path URI path template must also be specified as part of that paths `parameters`. | ✓ | ✓ |
+| `path-parameters-not-in-path` | Path parameters defined in `parameters` must also be specified in the path URI with path templating. | ✓ | ✓ |
+| `unknown-required-schema-property` | Schema properties that are listed as being required but don't exist within the schema. | ✕ | ✓ |
```
erunion: We don't support this rule on OpenAPI because with the way JSON Schema is composed in OpenAPI it's too difficult and slow for us to recursively validate these.

packages/parser/src/validators/spec.ts
```diff
@@ -9,13 +9,14 @@ import { SwaggerSpecificationValidator } from './spec/swagger.js';
 
 /**
  * Validates either a Swagger 2.0 or OpenAPI 3.x API definition against cases that aren't covered
- * by their JSON Schema definitions.
+ * by their respective JSON Schema definitions.
  *
```
julshotal: Do we want to call out we've relaxed validation to make some errors warnings in this comment?

julshotal (approved):

PR #1011, opened 2025-10-30
feat(oas): creating tests for dereferencing API defs that have invalid component names

erunion:
## 🧰 Changes

We have a number of customers that have invalid API definitions that contain invalid component names. For example:

```
OpenAPI schema validation failed.

`/components/securitySchemes/bearer token` has an invalid name. Component names should match against: /^[a-zA-Z0-9.-_]+$/

We have also found some additional warnings:

`/paths/api/rating/v2/batch-rater/jobs/{id}/start/post` has a path parameter named `Id`, but there is no corresponding `{Id}` in the path string.

`/paths/api/rating/v2/rates/audit/{id}/get` has a path parameter named `Id`, but there is no corresponding `{Id}` in the path string.

`/paths/api/rating/v2/rate-queries/rate-query/{id}/get` has a path parameter named `Id`, but there is no corresponding `{Id}` in the path string.

`/paths/api/rating/v2/batch-rater/jobs/{id}/get` has a path parameter named `Id`, but there is no corresponding `{Id}` in the path string.

`/paths/api/rating/v2/batch-rater/jobs/{id}/download/json/get` has a path parameter named `Id`, but there is no corresponding `{Id}` in the path string.
```

`/components/securitySchemes/bearer token` is invalid because schema names shouldn't contain spaces.

We already have a number of pieces within `@readme/openapi-parser` to downgrade certain errors to warnings (as visible above) and I've recently discovered that though Swagger Editor sees these as validation errors it calls them "semantic errors":

Screenshot 2025-10-30 at 2 39 32 PM

Because of this I am debating downgrading this invalid component schema name case to a "semantic" error that could be optionally treated as a validation warning. Before I do this however, I wanted to make sure that our general `oas` library was still able to handle these kinds of API definitions and have written tests for this against the following methods:

* [x] `Oas.dereference()`
* [x] `Oas.getAuth()`
* [x] `Operation.getParametersAsJSONSchema()`
* [x] `Operation.getResponseAsJSONSchema()`


packages/oas/test/operation/lib/get-parameters-as-json-schema.test.ts
erunion: Best to review these with white space changes off.

PR #1018, opened 2025-11-11
feat(oas-normalize): new option to return validation errors instead of throwing

erunion:
## 🧰 Changes

This expands the `.validate()` method within `oas-normalize` to create a new `throwIfInvalid` option that will allow us to always return a `ValidationResult` back for invalid API definitions, instead of always throwing a `ValidationError`. The reason I am adding this is because I need to be able to separate errors from warnings in this error as the current `ValidationError` message is being run through the `compileErrors` utility from `@readme/openapi-parser` that transforms errors and warnings into a single human-readable string.

To complement this new `throwIfInvalid` option I am also extending the existing `ValidationError` exception that's thrown to now contain an `errors` and `warnings` property for folks who do want exceptions thrown but still want separate access to this data.

### DIFF
--- a/packages/oas-normalize/src/index.ts
+++ b/packages/oas-normalize/src/index.ts
@@ -226,7 +226,7 @@ export default class OASNormalize {
       parser?: ParserOptions;
 
       /**
-       * Determines if `.validate()` throws a `ValidationError` if the API definition is invalid
+       * Determines if `.validate()` throws a `ValidationError` if the API definition is invalid.
        *
        * By default this will always happen however if you want to always instead receive a
        * `ValidationResult` object back, which will still allow you to determine if the supplied


packages/oas-normalize/README.md
kanadgupta: (i know it's not related to this PR per se but i was reading the whole thing and noticed this) on line 131, i think that `validate` object needs to be nested within a `parser` object, correct? sort of like how we do here:

https://github.com/readmeio/oas/blob/f6ec6fcb02cc6de9851fc99cfde4000d9da7b0a2/packages/oas-normalize/test/index.test.ts#L571-L573

packages/oas-normalize/README.md
```diff
@@ -148,7 +148,45 @@ try {
 }
 ```
 
-For full documentation on the available rulesets, as well as tooling to transform the the `.validate()` responses into a human-readable strings, check out the documentation for [`@readme/openapi-parser`](https://npm.im/@readme/openapi-parser).
+If you want to always receive a `ValidationResult` object back, which will still allow you to determine if the supplied API definition is valid or not, you can set the `throwIfInvalid` option to `false`:
```
kanadgupta:
```suggestion
If you want to always receive a `ValidationResult` object back (which includes a flag that indicates if the supplied API definition is valid or not), you can set the `throwIfInvalid` option to `false`:
```

kanadgupta (commented):
a few documentation suggestions otherwise LGTM!

packages/oas-normalize/README.md
erunion: will fix

packages/oas-normalize/README.md
```diff
@@ -148,7 +148,45 @@ try {
...
+
+Additionally, when receiving a `ValidationResult` object back you can also transform it into a human-readable string by utilizing our `compileErrors` utility:
+
+```ts
+import { compileErrors } from 'oas-normalize/lib/utils';
+
+const result = await oas.validate({ throwIfInvalid: false });
+if (result.valid) {
+  if (result.warnings.length) {
+    console.warn(compileErrors(result));
```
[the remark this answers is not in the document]
erunion: actually there's better docs on that in `openapi-parser`, going to rework the note i removed and link out to those docs instead

packages/oas-normalize/README.md
```diff
@@ -148,7 +150,25 @@ try {
...
 ```
 
-For full documentation on the available rulesets, as well as tooling to transform the the `.validate()` responses into a human-readable strings, check out the documentation for [`@readme/openapi-parser`](https://npm.im/@readme/openapi-parser).
+If you want to always receive a `ValidationResult` object back (which includes a flag that indicates if the supplied API definition is valid or not), you can set the `throwIfInvalid` option to `false`:
+
+```ts
+const result = await oas.validate({ throwIfInvalid: false });
+console.log(result);
+
+/* {
```
kanadgupta: i slightly prefer this but don't feel strongly:

```suggestion
/*
{
```

packages/oas-normalize/src/index.ts
```diff
@@ -224,8 +224,18 @@ export default class OASNormalize {
        * @deprecated
        */
       parser?: ParserOptions;
+
+      /**
+       * Should `.validate()` throw a `ValidationError` if the API definition is invalid?
```
julshotal:
```suggestion
       * Determines if `.validate()` throws a `ValidationError` if the API definition is invalid
```
Nit: can we word this less like a question? I got a little confused reading this comment before realizing this was a boolean prop

erunion: sure

PR #1020, opened 2025-11-25
perf: introduced caching for components within the `Operation` class

maximilianfalco:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

### Problem
The `getParametersAsJSONSchema` function was experiencing performance issues, with calls sometimes taking >1 second. The main bottleneck occurs when transforming components to handle circular references, particularly when `transformComponents` is called.

This PR makes `Oas` class maintains an internal `componentCache` that is automatically passed to all operations. When `getParametersAsJSONSchema` is called:

1. First call: Transforms all component schemas and caches the result (~1-2ms)
2. Subsequent calls: Reuses cached components (~0.01-0.02ms)

The cache uses the `OASDocument` object reference as the key to make sure it works across multiple operations

> [!NOTE]
> The caching introduced will only trigger and improve performance when circular refs are present. And when it is called multiple times.

## 🧬 QA & Testing
1. Use a OAS definition that has tons of refs and circular refs
2. Call the `getParametersAsJSONSchema` multiple times and check performance

## Some Benchmark Screenshots

With Cache | No Cache
-- | --
Screenshot 2025-12-04 at 17 47 28 | Screenshot 2025-12-04 at 17 47 44

### DIFF
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -30,6 +30,7 @@ import {
   SAMPLES_LANGUAGES,
   validateParameterOrdering,
 } from './extensions.js';
+import { buildDiscriminatorOneOf, findDiscriminatorChildren } from './lib/build-discriminator-one-of.js';
 import { getAuth } from './lib/get-auth.js';
 import getUserVariable from './lib/get-user-variable.js';
 import { isPrimitive } from './lib/helpers.js';
@@ -1083,6 +1084,12 @@ export default class Oas {
 
     this.dereferencing.processing = true;
 
+    // Discriminator Phase 1: Find discriminator schemas and their children before dereferencing (allOf $refs are resolved
+    // during dereferencing). For schemas with a discriminator using allOf inheritance, we build a
+    // oneOf array from the discovered child schemas so consumers can see the polymorphic options.
+    // (see https://spec.openapis.org/oas/v3.0.0.html#fixed-fields-20)
+    const discriminatorChildrenMap = findDiscriminatorChildren(this.api);
+
     const { api, promises } = this;
 
     // Because referencing will eliminate any lineage back to the original `$ref`, information that
@@ -1136,6 +1143,12 @@ export default class Oas {
     })
       .then((dereferenced: OASDocument) => {
         this.api = dereferenced;
+
+        // Discriminator Phase 2: Build oneOf arrays for discriminator schemas using dereferenced child schemas.
+        // This must be done after dereferencing so we have the fully resolved child schemas.
+        if (discriminatorChildrenMap && discriminatorChildrenMap.size > 0) {
+          buildDiscriminatorOneOf(this.api, discriminatorChildrenMap);
+        }
 
         this.promises = promises;
         this.dereferencing = {
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,11 @@ updates:
         update-types:
           - minor
           - patch
+      minor-production-deps:
+        dependency-type: 'production'
+        update-types:
+          - minor
+          - patch
     commit-message:
       prefix: chore(deps)
       prefix-development: chore(deps-dev)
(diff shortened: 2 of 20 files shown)


packages/oas-to-har/src/index.ts
```diff
@@ -253,6 +253,7 @@ export default function oasToHar(
       operationSchema?.path || '',
       operationSchema?.method || ('' as HttpMethods),
       (operationSchema as unknown as OperationObject) || { path: '', method: '' },
+      {},
```
erunion:
```suggestion
```

packages/oas/src/operation/index.ts
```diff
@@ -840,7 +875,7 @@ export class Callback extends Operation {
     identifier: string,
     parentPathItem: PathItemObject,
   ) {
-    super(oas, path, method, operation);
+    super(oas, path, method, operation, {});
```
erunion:
```suggestion
    super(oas, path, method, operation);
```

packages/oas/src/operation/lib/get-parameters-as-json-schema.ts
```diff
@@ -215,11 +228,30 @@ export function getParametersAsJSONSchema(
...
+   * @returns Transformed components object, or false if no components exist
+   */
   function transformComponents(): ComponentsObject {
     if (!('components' in api)) {
       return false;
     }
 
+    // Check cache first if caching is enabled
+    if (componentCache) {
+      const cached = componentCache.get(api);
```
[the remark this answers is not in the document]
erunion: Will this `.get()` pull anything back? I know that `Map` supports objects as keys but, at least in some cases, but this code doesn't seem to do that unless I'm missing something.

```
const map = new Map();
map.set({ doug: true }, { doug: true });
console.log({ 
  has: map.has({ doug: true }), 
  get: map.get({ doug: true })
});

// {has: false, get: undefined}
```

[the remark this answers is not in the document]
maximilianfalco: Ah I can respond directly now... but yea I think this case specifically works because we would be passing and using the same reference throughout an `Oas` instance

From my understanding the flow would be like this. `Oas.init(spec)` stores the reference in `this.api` and it passes that to `oas.operation()` which is then finally used in `operation.getParametersAsJSONSchema`.

packages/oas/src/operation/lib/get-parameters-as-json-schema.ts
maximilianfalco: @erunion It looks like we are always traversing ALL `$ref` components regardless if they are used within this operation or not. I managed to make a prototype where we only evaluate the once we actually need and dont have to resolve everything.

Just wanted to make sure here before I push my changes, was this behaviour (solving ALL components) intended and done on purpose? I guess we can make it selectively transform only referenced components but idk how widespread the effect of that may be. 

Idk how well this library is being used in the main ReadMe repo and that if it expects the full transformed schema or not when running `getParametersAsJSONSchema`. I would assume not since if users wanted that, they would just go to `operation.dereference()` instead.

maximilianfalco: Sounds good! Will look into this today!

maximilianfalco: Yea I think its a bit experimental to do that and not that big of an improvement

packages/oas/src/index.ts
```diff
@@ -291,6 +292,12 @@ export default class Oas {
     processing: boolean;
   };
 
+  /**
+   * Cache for transformed components, used to improve performance when processing multiple
+   * operations on the same API document. Limited to 100 entries with LRU eviction.
+   */
+  private componentCache?: Map<OASDocument, ComponentsObject | false>;
```
erunion: I'm a little confused on the shape of this `Map` now because its primary key is the API definition that the instance of `Oas` has. Won't this always just store one component object?

PR #1022, opened 2025-11-28
feat(dereference): build oneOf arrays for discriminator

maximilianfalco:
| 🚥 Resolves CX-2235 |
| :------------------- |

## 🧰 Changes

When OpenAPI schemas use the discriminator pattern with `allOf` inheritance, the polymorphic options become invisible to JSON Schema consumers after dereferencing.

Consider this OpenAPI schema:

```yaml
components:
  schemas:
    Pet:
      type: object
      discriminator:
        propertyName: pet_type
      properties:
        pet_type: { type: string }
    
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            name: { type: string }
    
    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            bark: { type: string }
```

After dereferencing, the Pet schema only has:
- ✅ `discriminator` property (tells us to look at `pet_type`)
- ❌ **No `oneOf` array** (we don't know what subtypes exist)

As a result, the ReadMe UI displays a malformed and incorrect UI state.

Screenshot 2025-11-28 at 15 47 19

This PR introduces a new feature (`buildDiscriminatorOneOfFromAllOf`) that automatically builds a `oneOf` array for discriminator schemas by discovering all child schemas that extend them via `allOf`.

<table>
<tr>
<th>
Before
</th>
<th>
After
</th>
</tr>
<tr>
<td>
Notice how it lacked the <code>oneOf</code> array

```json
{
  "Pet": {
    "type": "object",
    "discriminator": { "propertyName": "pet_type" },
    "properties": {
      "pet_type": { "type": "string" }
    }
  }
}
```
</td>
<td>
The same schema now produces:

```json
{
  "Pet": {
    "type": "object",
    "discriminator": { "propertyName": "pet_type" },
    "properties": {
      "pet_type": { "type": "string" }
    },
    "oneOf": [
      {
        "type": "object",
        "properties": {
          "pet_type": { "type": "string" },
          "name": { "type": "string" }
        },
        "x-readme-ref-name": "Cat"
      },
      {
        "type": "object",
        "properties": {
          "pet_type": { "type": "string" },
          "bark": { "type": "string" }
        },
        "x-readme-ref-name": "Dog"
      }
    ]
  }
}
```

</td>

</tr>
</table>

The implementation works in two phases:

1. **Phase 1 (Pre-dereference):** Discover discriminator schemas and their children
   - Find all schemas with a `discriminator` but no existing `oneOf`/`anyOf`
   - Scan for child schemas that reference the base via `allOf` 
   - Respect explicit `discriminator.mapping` if present

2. **Phase 2 (Post-dereference):** Build `oneOf` arrays
   - After dereferencing merges all `allOf` references
   - Add `oneOf` to the base discriminator schema containing all discovered child schemas
   - Each child schema is cloned to avoid circular reference issues

Now JSON Schema consumers can:
- see all polymorphic options (`Cat` and `Dog`)
- render proper UI for selecting subtypes
- generate accurate form fields based on the discriminator

## Breaking Changes

**None.** This feature is opt-in via a new option that defaults to `true`. To disable:

```typescript
await oas.dereference({ buildDiscriminatorOneOfFromAllOf: false });
```

## 🧬 QA & Testing
[openapi-allof.yaml](https://github.com/user-attachments/files/23814328/openapi-allof.yaml)
Use the file above to test. A `oneOf` array should be created.

### DIFF
--- a/packages/oas/test/operation/lib/get-response-as-json-schema.test.ts
+++ b/packages/oas/test/operation/lib/get-response-as-json-schema.test.ts
@@ -4,8 +4,8 @@ import { validate } from '@readme/openapi-parser';
 import { beforeAll, describe, expect, it } from 'vitest';
 
 import Oas from '../../../src/index.js';
-import discriminatorAllOfInheritance from '../../__datasets__/discriminator-allof-inheritance.json' with { type: 'json' };
-import petDiscriminatorAllOf from '../../__datasets__/pet-discriminator-allof.json' with { type: 'json' };
+import discriminatorAllOfInheritance from '../../__datasets__/discriminator-allof-inheritance.json';
+import petDiscriminatorAllOf from '../../__datasets__/pet-discriminator-allof.json';
 import { createOasForOperation } from '../../__fixtures__/create-oas.js';
 
 let circular: Oas;


packages/oas/test/__datasets__/pet-discriminator-allof.json
```diff
@@ -0,0 +1,96 @@
...
+        "required": ["pet_type"],
+        "properties": {
+          "pet_type": {
+            "type": "string"
+          }
+        },
+        "discriminator": {
+          "propertyName": "pet_type"
+        }
+      },
```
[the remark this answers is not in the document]
erunion: because this is from my understanding deviates from the json schema specification on discriminators i am very hesitant to merge this in because will create some unexpected behaviors with how `$ref` pointers and discriminators generally work elsewhere

im not sure what customer asked for this or what ticket prompted it but id rather push back and have them adjust their api definition to use discriminators the right way

packages/oas/src/lib/build-discriminator-one-of.ts
```diff
@@ -0,0 +1,155 @@
+import type { OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
+import type { OASDocument, SchemaObject } from '../types.js';
+
+import { isRef } from '../types.js';
+
+type DiscriminatorObject = OpenAPIV3.DiscriminatorObject | OpenAPIV3_1.DiscriminatorObject;
```
erunion: Can you move this into `lib/types.js`?

packages/oas/test/operation/lib/get-response-as-json-schema.test.ts
```diff
@@ -457,4 +458,102 @@ describe('#getResponseAsJSONSchema()', () => {
       });
     });
   });
+
+  describe('discriminator + allOf inheritance', () => {
+    async function loadDiscriminatorSpec() {
+      const data = await import('../../__datasets__/discriminator-allof-inheritance.json').then(r => r.default);
+      // Deep clone to avoid mutation between tests
+      return Oas.init(JSON.parse(JSON.stringify(data)));
+    }
```
erunion: I'm assuming the `useImportExtensions` ignore at the top of the file is because of the JSON import in here? Instead of that, can we do a common import of this JSON file at the top and then replace this method with the following?

```suggestion
    function loadDiscriminatorSpec() {
      return Oas.init(structuredClone(discrimnatorAllOfInheritance));
    }
```

packages/oas/test/operation/lib/get-response-as-json-schema.test.ts
```diff
@@ -457,4 +458,102 @@ describe('#getResponseAsJSONSchema()', () => {
...
+      const responseSchema = jsonSchema[0].schema;
+      const itemsSchema = responseSchema.properties.pets.items;
+
+      // Should not have oneOf since we disabled the feature
+      expect(itemsSchema.oneOf).toBeUndefined();
+      expect(itemsSchema.discriminator).toBeDefined();
+    });
+
+    it('should build oneOf from user-provided spec with Cat, Dog, and Lizard', async () => {
+      const userSpec = await import('../../__datasets__/pet-discriminator-allof.json').then(r => r.default);
```
erunion: Would also add this to the top of the file as a common import instead as well.

erunion (approved):

PR #1030, opened 2025-12-17
feat(oas-to-har): support content in params

azyl20:
| 🚥 Resolves RM-3225 |
| :------------------- |

## 🧰 Changes

- add support for content in params

## 🧬 QA & Testing

### DIFF
--- a/packages/parser/test/specs/validate-spec/__snapshots__/validate-spec.test.ts.snap
+++ b/packages/parser/test/specs/validate-spec/__snapshots__/validate-spec.test.ts.snap
@@ -44,6 +44,28 @@ exports[`Invalid APIs (specification validation) > components / definitions > sh
 }
 `;
 
+exports[`Invalid APIs (specification validation) > parameter content validation > given a parameter with both schema and content > OpenAPI 3.1 1`] = `
+{
+  "additionalErrors": 0,
+  "errors": [
+    {
+      "message": "IF must match "else" schema
+
+   9 |       "get": {
+  10 |         "parameters": [
+> 11 |           {
+     |           ^ if must match "else" schema
+  12 |             "name": "filter",
+  13 |             "in": "query",
+  14 |             "schema": {",
+    },
+  ],
+  "specification": "OpenAPI",
+  "valid": false,
+  "warnings": [],
+}
+`;
+
 exports[`Invalid APIs (specification validation) > parameter content validation > given a parameter with both schema and content > OpenAPI 3.x > should catch an error by default 1`] = `
 {
   "additionalErrors": 0,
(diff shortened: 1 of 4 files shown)


packages/parser/src/validators/spec/openapi.ts
```diff
@@ -219,23 +219,49 @@ export class OpenAPISpecificationValidator extends SpecificationValidator {
...
+  /**
+   * Validates parameter content object.
+   * Note: The requirement for exactly one media type is already enforced by the OpenAPI JSON schema.
+   */
+  private validateParameterContent(
+    content: OpenAPIV3_1.ParameterObject['content'] | OpenAPIV3.ParameterObject['content'],
+    parameterId: string,
+  ) {
+    const mediaTypes = Object.keys(content);
+    if (mediaTypes.length !== 1) {
```
azyl20: see note line 241. This is more for peace of mind

packages/parser/src/validators/spec/openapi.ts
```diff
@@ -219,23 +219,49 @@ export class OpenAPISpecificationValidator extends SpecificationValidator {
...
 
+      const parameterId = `${operationId}/parameters/${param.name}`;
+
       /**
-       * @todo add better handling when `content` is present instead of `schema`.
        * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#fixed-fields-10}
        * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-10}
        */
       if (!param.schema && param.content) {
+        this.validateParameterContent(param.content, parameterId);
```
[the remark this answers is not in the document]
erunion: Actually, it looks like you aren't testing OpenAPI 3.1 with the definition additions to `invalid/3.x`. Can you add one for that too? If the validation error is the same schema-level validation error as OpenAPI 3.0 then I'd say to definitely remove this `validateParameterContent` check and offload this work to the existing AJV schema validation.

packages/parser/src/validators/spec/openapi.ts
```diff
@@ -219,23 +219,49 @@ export class OpenAPISpecificationValidator extends SpecificationValidator {
...
+   * Validates parameter content object.
+   * Note: The requirement for exactly one media type is already enforced by the OpenAPI JSON schema.
+   */
+  private validateParameterContent(
+    content: OpenAPIV3_1.ParameterObject['content'] | OpenAPIV3.ParameterObject['content'],
+    parameterId: string,
+  ) {
+    const mediaTypes = Object.keys(content);
+    if (mediaTypes.length !== 1) {
+      this.reportError(
```
erunion: If this test case is always getting picked up by the schema-level validation for OpenAPI 3.0 and OpenAPI 3.1 it's probably safe to not worry about this. I'm cool with it as-is but will leave it up to you.

packages/oas/src/operation/lib/get-parameters-as-json-schema.ts
```diff
@@ -304,21 +304,12 @@ export function getParametersAsJSONSchema(
           } else if ('content' in current && typeof current.content === 'object') {
             const contentKeys = Object.keys(current.content);
             if (contentKeys.length) {
-              let contentType: string;
```
erunion: Thanks for moving this block into a testable utility.

packages/parser/src/validators/spec/openapi.ts
azyl20: I think I will leave it there :)

PR #1031, opened 2025-12-19
feat: Ability to get response schema by preferred header content type

Jadenzzz:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

Added `preferContentType` to Operation so we can fetch response schema based on different header 'accept' content type

## 🧰 Changes

Describe in detail what this PR is for.

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -64,6 +64,13 @@ export class Operation {
   contentType: string;
 
   /**
+   * Selected content type for this operation.
+   *
+   * @example 'application/json'
+   */
+  selectedContentType: string;
+
+  /**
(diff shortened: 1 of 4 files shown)


packages/oas/src/operation/index.ts
```diff
@@ -490,11 +518,6 @@ export class Operation {
     });
   }
 
-  /**
-   * Get a single response for this status code, formatted as JSON schema.
-   *
-   * @param statusCode Status code to pull a JSON Schema response for.
-   */
```
erunion: Can you add this back in?

erunion (changes requested):

PR #1113, opened 2026-04-09
fix(oas-to-har): improved `format: json` encoding support

erunion:
| 🚥 Resolves CX-3182 |
| :------------------- |

## 🧰 Changes

This addresses a bug in `oas-to-har` where we were always decoding form data as if it were JSON even if the schema properties didn't have `format: json` set.

### DIFF
--- a/packages/oas-to-har/src/index.ts
+++ b/packages/oas-to-har/src/index.ts
@@ -30,7 +30,7 @@ import {
   getSafeRequestBody,
   getTypedFormatsInSchema,
   hasSchemaType,
-  parseJsonStringsInBody,
+  parseJSONStringsInBodyWithSchema,
 } from './lib/utils.js';
 
 function formatter(
@@ -598,7 +598,7 @@ export default function oasToHar(
                 // values that are valid JSON so `format: json` is still resolved for our
                 // `application/json` payload.
                 try {
-                  const parsed = parseJsonStringsInBody(cleanBody) as Record<string, unknown>;
+                  const parsed: any = parseJSONStringsInBodyWithSchema(cleanBody, requestBodySchema, operation.api);
                   if (typeof parsed?.RAW_BODY !== 'undefined') {
                     har.postData.text = isPrimitive(parsed.RAW_BODY)
                       ? String(parsed.RAW_BODY)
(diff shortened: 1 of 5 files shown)


packages/oas-to-har/src/index.ts
```diff
@@ -598,7 +598,7 @@ export default function oasToHar(
                 // values that are valid JSON so `format: json` is still resolved for our
                 // `application/json` payload.
                 try {
-                  const parsed = parseJsonStringsInBody(cleanBody) as Record<string, unknown>;
+                  const parsed: any = parseJSONStringsInBodyWithSchema(cleanBody, requestBodySchema, operation.api);
```
kevinports, line 601: nit: Was there a reason to type this as `any` over at least asserting it to a record like before? Lesser of two evils i guess 🤷

kevinports (approved):
I don't know the oas library enough to really understand these changes, but code looks good to me.

packages/oas-to-har/src/lib/utils.ts
```diff
@@ -301,24 +301,113 @@ export function getParameterContentSchema(param: ParameterObject, contentType: s
...
+  return obj;
+}
+
+/**
+ * Recursively runs through a schema, parsing any values that have `format: json` attached and
+ * deserializing them into their JSON representations.
+ *
+ * @see {@link parseJSONStrings}
+ */
+export function parseJSONStringsInBodyWithSchema(
```
maximilianfalco, line 336: i think the changes in this function makes sense, Im just curious on whether its safe to keep passing the `seenRefs` set across different subtrees (branching sibling subtrees)? like for example

```
  components:
    schemas:
      Thing:
        type: object
        properties:
          label:
            type: string   # plain string, no format: json
  requestBody:
    content:
      application/json:
        schema:
          type: object
          properties:
            left:  { $ref: '#/components/schemas/Thing' }
            right: { $ref: '#/components/schemas/Thing' }

  With body { "left": { "label": "[1,2,3]" }, "right": { "label": "[4,5,6]" } }:
``` 

the `left` and `right` as different in terms of `body` but then after `left` is parsed, `Thing` would be in `seenRefs` right? so the `Thing` in `right` wont be properly parsed

maybe we can clone the `seenRefs` before we branch off subtrees

maximilianfalco (approved):
generally approve with just one very small question

packages/oas-to-har/src/index.ts
erunion: Yeah technically `any` is the better typing here because `cleanBody` is `unknown` and `parsed` isn't guaranteed to be a record.

packages/oas-to-har/src/lib/utils.ts
erunion: yeah good point, i put in a fix

PR #1114, opened 2026-04-09
fix(oas): polymorphic merging can sometimes drop schema properties

erunion:
| 🚥 Resolves CX-3194, CX-3195 |
| :------------------- |

## 🧰 Changes

This addresses a race condition where if an `allOf` and `oneOf`, that itself had a nested `allOf` are merged together we would sometimes lose schema properties of the `oneOf` and its nested `allOf`. For example, with the following schema:

```js
allOf: [
  {
    type: 'object',
    properties: {
      testiks: {
        type: 'string'
      }
    }
  }
],
oneOf: [
  {
    title: 'Test 1',
    allOf: [
      {
        properties: {
          testik: {
            type: 'string'
          }
        }
      },
      {
        $ref: '#/components/schemas/Testsasffqefeq'
      }
    ]
  }
]
```

When we resolve and generate a schema for this it would be generated to the following:

```js
{
  type: 'object',
  oneOf: [
    {
      type: 'object',
      properties: { 
        test: { type: 'string' }, 
        testiks: { type: 'string' }
      },
      'x-readme-ref-name': 'Testsasffqefeq',
    }
  ]
}
```

Not only would we lose the `title` property that the `oneOf` had but we also would lose the `testik` object property. With the fix I have here we're now able to fully preserve this object as intended:

```js
{
  title: 'Test 1',
  type: 'object',
  properties: {
    test: {
      type: 'string',
    },
    testik: {
      type: 'string',
    },
    testiks: {
      type: 'string',
    },
  },
  'x-readme-ref-name': 'Testsasffqefeq',
}
```

### DIFF
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -1,4 +1,5 @@
-import type { JSONSchema7TypeName } from 'json-schema';
+import type { JSONSchema4, JSONSchema7TypeName } from 'json-schema';
+import type { Options as JSONSchemaMergeAllOfOptions } from 'json-schema-merge-allof';
 import type {
   ExampleObject,
   JSONSchema,
@@ -29,6 +30,38 @@ const UNSUPPORTED_SCHEMA_PROPS = [
   'externalDocs',
   'xml',
 ] as const;
+
+const mergeAllOfSchemasOptions: JSONSchemaMergeAllOfOptions = {
+  ignoreAdditionalProperties: true,
+  resolvers: {
+    // `merge-json-schema-allof` by default takes the first `description` when you're merging an
+    // `allOf` but because generally when you're merging two schemas together with an `allOf` you
+    // want data in the subsequent schemas to be applied to the first and `description` should be a
+    // part of that.
+    description: (obj: string[]) => {
+      return obj.slice(-1)[0];
+    },
+
+    // `merge-json-schema-allof` doesn't support merging enum arrays but since that's a safe and
+    // simple operation as enums always contain primitives we can handle it ourselves with a custom
+    // resolver. We intersect the arrays so that child schemas can narrow a parent's broad enum
+    // (e.g. [1,2,20,50] ∩ [1] = [1]).
+    //
+    // We unfortunately need to cast our return value as `any[]` because the internal types of
+    // `merge-json-schema-allof`'s `enum` resolver are not portable.
+    enum: (obj: unknown[]) => {
+      const arrays = obj as any[][];
+      const intersection = arrays.reduce((acc, e) => acc.filter(v => e.includes(v)));
+      return intersection.length > 0 ? intersection : arrays.reduce((acc, e) => acc.concat(e), []);
+    },
+
+    // For any unknown keywords (e.g., `example`, `format`, `x-readme-ref-name`), we fallback to
+    // using the `title` resolver (which uses the first value found).
+    // https://github.com/mokkabonna/json-schema-merge-allof/blob/ea2e48ee34415022de5a50c236eb4793a943ad11/src/index.js#L292
+    // https://github.com/mokkabonna/json-schema-merge-allof/blob/ea2e48ee34415022de5a50c236eb4793a943ad11/README.md?plain=1#L147
+    defaultResolver: mergeJSONSchemaAllOf.options.resolvers.title,
+  },
+};
 
 export interface toJSONSchemaOptions {
   /**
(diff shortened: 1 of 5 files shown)


packages/oas/test/operation/transformers/get-parameters-as-json-schema.test.ts
```diff
@@ -495,6 +496,55 @@ describe('.getParametersAsJSONSchema()', () => {
...
+                    },
+                  },
+                  'x-readme-ref-name': 'Testsasffqefeq',
+                },
+              ],
+            },
+          },
+          components: {
+            schemas: {
+              Testsasffqefeq: {
```
erunion, line 535: This component schema shouldn't have been preserved because we resolved the `$ref` that guided us to it, however I'm nervous about causing more regressions right now with schema preservation so I'm going to say this is ok for now and will address it later.

I'd rather preserve too much now than lose required data.

kevinports (approved):

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -565,6 +598,35 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
...
+              currentLocation: `${currentLocation}/${encodePointer(prop)}`,
+              prevDefaultSchemas,
+              prevExampleSchemas,
+            });
+          } else {
+            preprocessed[prop] = val;
+          }
+        }
+
+        schema = { ...schema, properties: preprocessed as SchemaObject['properties'] } as SchemaObject;
```
maximilianfalco:
```suggestion
        schema = { ...schema, properties: preprocessed } as SchemaObject;
```
the `preprocessed` should be typed already above in L612 right?

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -575,19 +637,55 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
...
+            }
+
+            const siblingSchema = toJSONSchema(siblings as SchemaObject, allOfOptions);
+            try {
+              return mergeJSONSchemaAllOf(
+                { allOf: [resolved as JSONSchema4, siblingSchema as JSONSchema4] },
+                mergeAllOfSchemasOptions,
+              ) as SchemaObject;
+            } catch {
+              return resolved;
```
maximilianfalco, line 684: should we log an error here just so we dont silently swallow errors. from my understanding, `resolved` here drops title (maybe other properties as well) which I think might be good to log if we expect something to be missing

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -575,19 +637,55 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
         // unwrapping the schema, so `$ref` pointers _do_ appear in the output then we **should**
         // log those.
         const allOfOptions: toJSONSchemaOptions =
-          schema.allOf.length > 1 ? { ...polyOptions, refLogger: () => {} } : polyOptions;
+          // biome-ignore lint/style/noNonNullAssertion: We've narrowed above to have an `allOf` array.
+          schema.allOf!.length > 1 ? { ...polyOptions, refLogger: () => {} } : polyOptions;
```
maximilianfalco: this is just a nit but hoisting `schema.allOf` into a const should allow us to remove the biome ignores right? maybe im missing something...

packages/oas/test/operation/transformers/get-parameters-as-json-schema.test.ts
```diff
@@ -495,6 +497,97 @@ describe('.getParametersAsJSONSchema()', () => {
...
+                },
+                'x-readme-ref-name': 'Testsasffqefeq',
+              },
+            },
+          },
+        });
+      });
+
+      it('should preserve siblings if they are `$ref` pointers', async () => {
+        const oas = Oas.init(cx3195);
```
maximilianfalco: should we be `structuredClone`-ing just like the new test above?

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -565,6 +598,35 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
...
+        schema.properties !== undefined &&
+        typeof schema.properties === 'object' &&
+        schema.properties !== null &&
+        !Array.isArray(schema.properties)
+      ) {
+        const preprocessed: SchemaObject['properties'] = {};
+        for (const prop of Object.keys(schema.properties)) {
+          const val = schema.properties[prop];
+          if (Array.isArray(val) || (typeof val === 'object' && val !== null)) {
+            preprocessed[prop] = toJSONSchema(val as SchemaObject, {
```
maximilianfalco, line 616: i might be overthinking this but can this ever clash with the toJSONSchema call we are making to the same `schema.properties[prop]` over in [L1113](https://github.com/readmeio/oas/pull/1114/changes#diff-a960342c2fc65f30371e7d8daac7c12cc75220e6d1430ff803ea16c4c94daa9dL1045)

maximilianfalco (approved):
overall looks good! just some non blocking nits!

packages/oas/src/lib/openapi-to-json-schema.ts
erunion: it is. this is remnants from me fiddling with the types before adding `as SchemaObject` to the end. will remove.

packages/oas/src/lib/openapi-to-json-schema.ts
erunion: We unfortunately have no mechanisms for logging errors within this library currently. I've thought in the past about introducing the [debug](https://npm.im/debug) library for this but am hesitant to introduce that into our compiled dists.

packages/oas/src/lib/openapi-to-json-schema.ts
erunion: Yep, I missed that we already have that set to a local variable on line 633. Resovled.

packages/oas/test/operation/transformers/get-parameters-as-json-schema.test.ts
erunion: yeah might as well

packages/oas/src/lib/openapi-to-json-schema.ts
erunion: We are doing some double work for polymorphic schemas with this, yeah. Refactoring this is going to be a bit hairy so I'm going to hold off on that for now because it seems fine for now, we're just wasting cycles recalculating the same schema twice sometimes.

PR #1115, opened 2026-04-09
fix(oas): support `$ref` pointers in `allOf` during example generation

erunion:
| 🚥 Resolves CX-3191 |
| :------------------- |

## 🧰 Changes

This fixes a bug in our `getResponseExamples` method where we weren't pre-resolving `$ref` pointers in `allOf` schemas prior to merging them together with `json-schema-merge-allof`. For example the following schema:

```js
{
  allOf: [
    {
      $ref: '#/components/schemas/UserStatus'
    },
    {
      $ref: '#/components/schemas/UserInfo'
    }
  ]
}
```

When generating an example from this we'd only pick the first entry, `UserStatus`, and generate an example for our response from that:

```js
{
  active: true,
  email: 'john.doe@example.com',
}
```

With this work we're now able to fully handle _both_ branches of this `allOf` and generate the following:

```js
{
  id: 123,
  name: 'John Doe',
  active: true,
  email: 'john.doe@example.com',
}
```

### DIFF
--- a/packages/oas/src/samples/index.ts
+++ b/packages/oas/src/samples/index.ts
@@ -140,14 +140,32 @@ function sampleFromResolvedSchema(
   const hasPolymorphism = usesPolymorphism(schema);
   if (hasPolymorphism === 'allOf') {
     try {
+      const definition = opts.definition;
+      const resolvedAllOf = schema.allOf.map((subSchema: SchemaObject) => {
+        let sub = objectify(subSchema);
+        if (definition && isRef(sub)) {
+          // We need to dereference `$ref` pointers before mmerging our schemas together because
+          // `json-schema-merge-allof` does not support `$ref` resolutions.
+          const resolved = dereferenceRef(sub, definition, new Set<string>());
+          if (resolved && !isRef(resolved)) {
+            sub = resolved as SchemaObject;
+          }
+        }
+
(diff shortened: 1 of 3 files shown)


dannobytes (approved):

packages/oas/src/samples/index.ts
```diff
@@ -140,14 +140,32 @@ function sampleFromResolvedSchema(
   const hasPolymorphism = usesPolymorphism(schema);
   if (hasPolymorphism === 'allOf') {
     try {
+      const definition = opts.definition;
+      const resolvedAllOf = schema.allOf.map((subSchema: SchemaObject) => {
+        let sub = objectify(subSchema);
+        if (definition && isRef(sub)) {
+          // We need to dereference `$ref` pointers before mmerging our schemas together because
+          // `json-schema-merge-allof` does not support `$ref` resolutions.
+          const resolved = dereferenceRef(sub, definition, new Set<string>());
```
maximilianfalco, line 149: wait is there a reason why we arent passing in `seenRefs` and instead passing in a new Set?

erunion: Example generation is separate from JSON Schema generation so we don't have the same `seenRefs` dictionary

PR #1119, opened 2026-04-14
fix(oas): addressing more issues with `$ref` pointers being excluded

erunion:
| 🚥 Resolves CX-3205 |
| :------------------- |

## 🧰 Changes

This addresses more issues in JSON Schema generation where `$ref` pointers weren't always preserved, resulting in a pointer that could not be resovled.


ashacs (approved):
once again i believe u

darrenyong (approved):

PR #1120, opened 2026-04-16
fix(oas-to-har): polymorphic `multipart/form-data` schema support

erunion:
| 🚥 Resolves CX-3220 |
| :------------------- |

## 🧰 Changes

This fixes a number of quirky issues with how we do payload matching against polymorphic schemas.

Currently the way it works is that if you have a polymorphic schema, meaning you have a `oneOf` or an `anyOf` we will just pick the first an say good luck, we hope the payload you have matches that. The way it'll work now is that we'll do branch-level scoring within each `oneOf` and `anyOf` schema in order to determine which branch best matches the payload given. This means we'll now be able to generate HAR files, and by extension code snippets in `oas-to-snippet`, for polymirphic branches that _aren't_ the first schema.

### DIFF
--- a/packages/oas-to-har/test/requestBody.test.ts
+++ b/packages/oas-to-har/test/requestBody.test.ts
@@ -11,6 +11,7 @@ import oasToHar from '../src/index.js';
 import deeplyNestedJsonFormats from './__datasets__/deeply-nested-json-formats.json' with { type: 'json' };
 import formdataNestedObject from './__datasets__/formData-nested-object.json' with { type: 'json' };
 import cx3182 from './__datasets__/issues/CX-3182.json' with { type: 'json' };
+import cx3220 from './__datasets__/issues/CX-3220.json' with { type: 'json' };
 import multipartFormData from './__datasets__/multipart-form-data.json' with { type: 'json' };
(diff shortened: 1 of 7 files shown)


packages/oas-to-har/test/requestBody.test.ts
```diff
@@ -519,7 +520,28 @@ describe('request body handling', () => {
...
+
+          const har = oasToHar(oas, operation, values, {});
+
+          expect(har.log.entries[0].request.headers).toStrictEqual([
+            { name: 'content-type', value: 'multipart/form-data' },
+          ]);
+
+          expect(har.log.entries[0].request.postData).toStrictEqual({
+            mimeType: 'multipart/form-data',
+            params: [{ name: 'grant_type2', value: 'two' }],
```
erunion, line 540: Prior to all of the fixes in this work, this payload was being generated as a JSON-encoded string to the `text` field of the HAR, and because our code snippet generation only supports `params` for `multipart/form-data` requests we were generating code snippets without payloads.

All of this is a regression from the dereferencing refactors I did in https://github.com/readmeio/oas/pull/1082.

darrenyong (approved):

PR #1121, opened 2026-04-22
fix: resolve deep self-ref through allOf

maximilianfalco:
| 🚥 Resolves CX-3218 |
| :------------------- |

## 🧰 Changes

When a schema contains a deep self-referencing `$ref` through an `allOf` (e.g. `#/components/schemas/Foo/allOf/1/properties/bar`), `toJSONSchema` merges the `allOf` flat but preserves the `$ref` stub in the output. The resolved schema is stored in `usedSchemas` and later placed into the root schema by `mergeReferencedSchemasIntoRoot`.

The bug: `mergeReferencedSchemasIntoRoot` iterated the ref map in insertion order. When the child ref (the deep `allOf` pointer) was inserted before the parent ref (the component schema), the child was placed first — creating the `allOf` path — and then the parent overwrote the entire component schema with the merged flat object, destroying the child's path. This left a dangling `$ref` that crashed `findSchemaDefinition` at render time.

The fix: sort entries by path depth (shallowest first) before placing them so parent schemas are always written before their children.

## 🧬 QA & Testing

[minimal-self-ref.json](https://github.com/user-attachments/files/26952387/minimal-self-ref.json)

Use this file to test.

### Before (broken)

The `allOf` is merged flat but `nested.properties.rate` still has a `$ref` pointing into the old `allOf` path which no longer exists. Dangling pointer → crash.

```json
{
  "$ref": "#/components/schemas/Payload",
  "components": {
    "schemas": {
      "Payload": {
        "type": "object",
        "properties": {
          "id": { "type": "string" },
          "rate": { "type": "number", "format": "double" },
          "nested": {
            "type": "object",
            "properties": {
              "rate": {
                "$ref": "#/components/schemas/Payload/allOf/1/properties/rate"
              }
            }
          }
        }
      }
    }
  }
}
```

### After (fixed)

`mergeReferencedSchemasIntoRoot` now places the parent first, then the child creates the `allOf` path on top. The `$ref` resolves.

```json
{
  "$ref": "#/components/schemas/Payload",
  "components": {
    "schemas": {
      "Payload": {
        "type": "object",
        "properties": {
          "id": { "type": "string" },
          "rate": { "type": "number", "format": "double" },
          "nested": {
            "type": "object",
            "properties": {
              "rate": {
                "$ref": "#/components/schemas/Payload/allOf/1/properties/rate"
              }
            }
          }
        },
        "allOf": [
          null,
          {
            "properties": {
              "rate": { "type": "number", "format": "double" }
            }
          }
        ]
      }
    }
  }
}
```

### DIFF
--- a/.changeset/fix-deep-self-ref-allof.md
+++ b/.changeset/fix-deep-self-ref-allof.md
@@ -1,0 +1,5 @@
+---
+"oas": patch
+---
+
+Fixed a crash when a schema contains a deep self-referencing `$ref` through an `allOf` (e.g. `#/components/schemas/Foo/allOf/1/properties/bar`) by sorting ref entries by path depth before merging into the root schema.
--- a/packages/oas/src/lib/refs.ts
+++ b/packages/oas/src/lib/refs.ts
@@ -313,7 +313,16 @@ function childShouldBeSchemaArray(parentKey: string, childSeg: string | undefine
  * @example `#/components/schemas/MySchema` -> `root.components.schemas.MySchema`
  */
 export function mergeReferencedSchemasIntoRoot(root: SchemaObject, refToSchema: Map<string, unknown>): void {
-  for (const [ref, schema] of refToSchema) {
+  // we need to sort this so parent schemas are placed before child schemas to
+  // prevent a parent from overwriting a child's synthetically created path
+  // sorting by path depth (shallowest first)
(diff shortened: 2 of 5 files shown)


erunion (changes requested):
can you add a patch changeset for this for the changelog? you can do that with `npx changeset`. this is a new addition i rolled out last week to help clean up the release flow

erunion (approved):

PR #1125, opened 2026-04-28
feat(oas): getParametersAsJsonSchema supports choosing content-type

azyl20:
| 🚥 Resolves RM-4392 |
| :------------------- |

## 🧰 Changes

- allow getParametersAsJsonSchema to return the correct schema based on inputted content-type so that the user can select which request body content-type to view in API explorer.
- also unified `operation.getRequestBody` to return the same shape whether or not content-type was passed in. Makes call sites simpler.

## 🧬 QA & Testing

### DIFF
--- a/.changeset/quick-cycles-buy.md
+++ b/.changeset/quick-cycles-buy.md
@@ -1,6 +1,6 @@
 ---
-"@readme/oas-to-har": major
-"oas": major
+'@readme/oas-to-har': major
+'oas': major
 ---
 
 getParameterAsJSONSchema returns based on content-type parameter
--- a/.changeset/quick-cycles-buy.md
+++ b/.changeset/quick-cycles-buy.md
@@ -1,6 +1,6 @@
 ---
-"@readme/oas-to-har": minor
-"oas": minor
+"@readme/oas-to-har": major
+"oas": major
 ---
 
 getParameterAsJSONSchema returns based on content-type parameter
--- a/.changeset/quick-cycles-buy.md
+++ b/.changeset/quick-cycles-buy.md
@@ -1,0 +1,6 @@
+---
+"@readme/oas-to-har": minor
+"oas": minor
+---
+
+getParameterAsJSONSchema returns based on content-type parameter
--- a/packages/oas/src/operation/index.ts
+++ b/packages/oas/src/operation/index.ts
@@ -890,7 +890,9 @@ export class Operation {
    * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#media-type-object}
    * @param mediaType Specific request body media type to retrieve if present.
    */
-  getRequestBody(mediaType?: string): false | { mediaType: string; mediaTypeObject: MediaTypeObject; description?: string } {
+  getRequestBody(
+    mediaType?: string,
+  ): false | { mediaType: string; mediaTypeObject: MediaTypeObject; description?: string } {
     if (!this.hasRequestBody()) {
       return false;
     }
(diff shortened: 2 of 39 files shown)


packages/oas/src/operation/index.ts
```diff
@@ -890,7 +890,7 @@ export class Operation {
    * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#media-type-object}
    * @param mediaType Specific request body media type to retrieve if present.
    */
-  getRequestBody(mediaType?: string): MediaTypeObject | false | [string, MediaTypeObject, ...string[]] {
+  getRequestBody(mediaType?: string): false | [string, MediaTypeObject, ...string[]] {
```
erunion: I have long hated the way I did this method return shape but I'm kind of hesitant to ship another breaking change release so soon after 32.0.0 a month ago. If we're going to rework this, can we go an extra step further and move away from the ugly `[string, MediaTypeObject]` shape instead? Something like this?

```suggestion
  getRequestBody(mediaType?: string): false | { mediaType: string, object: MediaTypeObject, description?: string } {
```

Thoughts?

packages/oas/test/operation/index.test.ts
```diff
@@ -1666,6 +1696,53 @@ describe('#getRequestBody()', () => {
...
+          },
+        },
+      });
+
+      const result = op.getRequestBody();
+      expect(result).toStrictEqual([
+        'application/json',
+        { schema: { type: 'object' } },
+      ]);
+      expect(result).toHaveLength(2);
```
erunion: The `strictEqual` check is good enough here.

```suggestion
```

erunion (changes requested):

erunion:
Can you also run `npm run changeset` and create+commit a changeset for your work here?

packages/oas/src/operation/index.ts
azyl20: yeah I agree that sounds good. Have made this change and ran changeset for a minor version bump

.changeset/quick-cycles-buy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@readme/oas-to-har": minor
+"oas": minor
```
erunion: The work here is going to be a breaking change and should be treated as a major release.

```suggestion
"@readme/oas-to-har": major
"oas": major
```

packages/oas/src/operation/index.ts
```diff
@@ -903,7 +903,11 @@ export class Operation {
         return false;
       }
 
-      return requestBody.content[mediaType];
+      return {
+        mediaType,
+        mediaTypeObject: requestBody.content[mediaType],
+        ...(requestBody.description ? { description: requestBody.description } : {}),
```
erunion: would just let this be undefined if it's not here

```suggestion
        description: requestBody.description,
```

packages/oas/src/operation/index.ts
```diff
@@ -925,11 +929,11 @@ export class Operation {
...
     if (availableMediaType) {
-      return [
-        availableMediaType,
-        requestBody.content[availableMediaType],
-        ...(requestBody.description ? [requestBody.description] : []),
-      ];
+      return {
+        mediaType: availableMediaType,
+        mediaTypeObject: requestBody.content[availableMediaType],
+        ...(requestBody.description ? { description: requestBody.description } : {}),
```
erunion:
```suggestion
        description: requestBody.description,
```

erunion (changes requested):
big fan of this refactor

erunion (dismissed):

erunion (changes requested):
looks like you have a lot of broken tests.

azyl20:
> looks like you have a lot of broken tests.

ah yeah cus of the description change, should be good now

erunion (approved):

PR #1126, opened 2026-04-29
fix(oas): preserve $ref metadata siblings on props inside allOf

maximilianfalco:
| 🚥 Resolves CX-3276 |
| :------------------- |

## 🧰 Changes

When a property under an `allOf` branch had `$ref` with sibling keywords (e.g. `description`, `summary`), the siblings were being dropped during property-ref inlining — so customers writing valid OAS 3.1 schemas like:

```json
"code": {
  "$ref": "#/components/schemas/Pet",
  "description": "The code of the error"
}
```

…would see `code` render without the description in API Explorer.

This PR updates `inlinePropertyRefsForMerge` to spread the original `$ref` siblings on top of the resolved schema, mirroring the existing top-level `$ref` sibling-preservation pattern. The invalid `properties` is still stripped to prevent another occurrence of [CX-3171](https://linear.app/readme-io/issue/CX-3171) (added a regression test as well)

> [!NOTE]
this does deviate from the original OpenAPI 3.0 spec. in version 3.0, `refs` need ALL their sibling props removed including `description` and `summary`, see [here](https://swagger.io/docs/specification/v3_0/using-ref/). but this preservation does align with the OpenAPI 3.1 spec: [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schema-object) and [here](https://json-schema.org/draft/2020-12/json-schema-core#name-direct-references-with-ref)

## 🧬 QA & Testing

Added `CX-3276.json` fixture and a regression test in `get-parameters-as-json-schema.test.ts` covering both shapes below.

<details>
<summary>Testing Details</summary>

### 1. Preserve `description` / `summary` (CX-3276)

**Input**: property `code` with `$ref` plus sibling `description` and `summary`:

```json
"Error": {
  "allOf": [
    { "$ref": "#/components/schemas/Base" },
    {
      "type": "object",
      "properties": {
        "code": {
          "$ref": "#/components/schemas/Pet",
          "description": "The code of the error",
          "summary": "Error code"
        }
      }
    }
  ]
}
```

<table>
<thead>
<tr>
<th>❌ Before: siblings silently dropped</th>
<th>✅ After: <code>description</code> and <code>summary</code> survive</th>
</tr>
</thead>
<tbody>
<tr>
<td>

```json
"code": {
  "type": "object",
  "x-readme-ref-name": "Pet",
  "properties": {
    "id": { "type": "integer", "format": "int64" },
    "name": { "type": "string" }
  }
}
```

</td>
<td>

```json
"code": {
  "type": "object",
  "x-readme-ref-name": "Pet",
  "description": "The code of the error",
  "summary": "Error code",
  "properties": {
    "id": { "type": "integer", "format": "int64" },
    "name": { "type": "string" }
  }
}
```

</td>
</tr>
</tbody>
</table>

### 2. Strip invalid `properties` sibling (CX-3171 regression)

**Input**: property `stripMe` with `$ref` plus an invalid `properties` sibling that contains a deeper `$ref` (this was the original CX-3171 crash shape):

```json
"stripMe": {
  "$ref": "#/components/schemas/Pet",
  "properties": {
    "deepRef": { "$ref": "#/components/schemas/Tag" }
  }
}
```

<table>
<thead>
<tr>
<th>❌ Before: would crash on the stray <code>deepRef</code></th>
<th>✅ After: invalid <code>properties</code> dropped, no leak</th>
</tr>
</thead>
<tbody>
<tr>
<td>

```
(crash in API Explorer — CX-3171)
```

</td>
<td>

```json
"stripMe": {
  "type": "object",
  "x-readme-ref-name": "Pet",
  "properties": {
    "id": { "type": "integer", "format": "int64" },
    "name": { "type": "string" }
  }
}
```

</td>
</tr>
</tbody>
</table>

The regression test asserts both shapes simultaneously, so any future change that removes the `properties` strip (or breaks sibling preservation) will fail.

</details>

### DIFF
--- a/.changeset/pre-ajv-security-scheme-validation.md
+++ b/.changeset/pre-ajv-security-scheme-validation.md
@@ -1,5 +1,5 @@
 ---
-"@readme/openapi-parser": minor
+'@readme/openapi-parser': minor
 ---
 
 Added pre-AJV validation for quirky security schemes. Surfaces clear, targeted errors (missing `type`, cross-type contamination, invalid `apiKey.in`, empty `oauth2.flows`, Swagger 2.0 ↔ OAS 3.x type confusion) instead of AJV's noisy `oneOf` failures, configurable via the new `invalid-security-scheme-properties` rule.
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -32,6 +32,14 @@ const UNSUPPORTED_SCHEMA_PROPS = [
   'xml',
 ] as const;
 
+/**
+ * Annotation-only keywords we preserve when they appear as siblings of `$ref` inside `allOf`
+ * properties. Validation keywords (`type`, `properties`, `items`, etc.) are intentionally
+ * excluded, they would conflict with the resolved schema. `x-` extensions are also preserved
+ * (handled separately at the call site).
+ */
+const METADATA_SIBLING_KEYS = new Set(['description', 'summary', 'title']);
+
 const mergeAllOfSchemasOptions: JSONSchemaMergeAllOfOptions = {
   ignoreAdditionalProperties: true,
   resolvers: {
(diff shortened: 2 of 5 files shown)


.changeset/pre-ajv-security-scheme-validation.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@readme/openapi-parser": minor
+'@readme/openapi-parser': minor
```
maximilianfalco, line 2: surprised the linter didnt catch this one again...

erunion: Just noticed it isn't getting caught because I'm running `npm run format` in CI which rewrites the files if they've got problems. I've changed this to run checks instead.

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -233,8 +233,27 @@ function inlinePropertyRefsForMerge(
...
+        // https://swagger.io/docs/specification/v3_0/using-ref/#:~:text=%24ref%20and%20Sibling%20Elements
+        //
+        // OpenAPI 3.1 permits us! OAS 3.1 Schema Object is "a superset of the JSON Schema Specification
+        // Draft 2020-12", see:
+        // https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schema-object
+        //
+        // …and it explicitly states that "other keywords can appear alongside of `$ref` in the same schema
+        // object", see:
+        // https://json-schema.org/draft/2020-12/json-schema-core#name-direct-references-with-ref
+        const { $ref: _$ref, properties: _propertiesWithRef, ...siblings } = val as Record<string, unknown>;
```
erunion: What about `items`? Is the only sibling we want to preserve here `description`?

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -233,8 +233,27 @@ function inlinePropertyRefsForMerge(
...
       if (resolved !== undefined && !isPendingSchema(resolved)) {
+        // Preserve metadata siblings (e.g. description, summary) alongside the `$ref` pointer so
+        // they survive `allOf` merging.
+        //
+        // NOTE: This deviates a little bit from the 3.0 spec but is necessary
+        // to avoid losing metadata when inlining `$ref` schemas into `allOf`. Also just better
+        // user experience to keep the metadata around.
+        //
+        // OpenAPI 3.0 requires us to strip everything, see:
+        // https://swagger.io/docs/specification/v3_0/using-ref/#:~:text=%24ref%20and%20Sibling%20Elements
```
darrenyong: I didn't realize this before, but one of our past engineers mentioned that the documentation here is a company's interpretation of how OAS should function so it shouldn't be treated as official documentation (it's actually by a company called SmartBear). The official documentation lives [here](https://github.com/OAI/OpenAPI-Specification/tree/main/versions)

darrenyong (approved):
just the comment from Jon!

packages/oas/src/lib/openapi-to-json-schema.ts
erunion: yeah +1 to this. I'd avoid https://swagger.io for OpenAPI docs when possible and instead rely on the official specification

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: i believe the spec only allows metadata to be passed so items _should_ also be stripped much like properties

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: oh what... okay noted, srry didnt know this. will ammend!

packages/oas/src/lib/openapi-to-json-schema.ts
erunion: 👍 looks like `items` needs to be excluded then.

erunion: Or instead of making this into a denylist of fields we _don't_ want, just pick out the metadata ones: `description`, `summary`, `title`, and maybe `x-` extensions. Any others?

packages/oas/src/lib/openapi-to-json-schema.ts
darrenyong: no worries - I thought swagger.io was official documentation too!

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: i think thats it tbh, will have another read at the docs to make sure before merging this in!

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: fixed in https://github.com/readmeio/oas/pull/1126/commits/9f353f78236403d9ff0ef9e0332171a3f1b68c8a

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: fixed in https://github.com/readmeio/oas/pull/1126/commits/9f353f78236403d9ff0ef9e0332171a3f1b68c8a

PR #1127, opened 2026-04-29
chore: flipping on oxlint's `suspicious` group and addressing issues

erunion:
## 🧰 Changes

This flips on Oxlint's `suspicious` group across the library and addresses the handful of issues that cropped up.

### DIFF
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -481,7 +481,7 @@ function searchForValueByPropAndPointer(
     return;
   }
 
-  const locSplit = pointer.split('/').filter(Boolean).reverse();
+  const locSplit = pointer.split('/').filter(Boolean).toReversed();
   const pointers = [];
 
   let point = '';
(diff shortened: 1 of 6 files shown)


packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -481,7 +481,7 @@ function searchForValueByPropAndPointer(
     return;
   }
 
-  const locSplit = pointer.split('/').filter(Boolean).reverse();
+  const locSplit = pointer.split('/').filter(Boolean).toReversed();
```
erunion, line 484: `.toReversed()` and `.toSorted()` should be generally available now.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed

darrenyong (approved):

darrenyong:
for anyone else: the `suspicious` flag warns or throws errors on [code that is likely to be wrong or useless](https://arc.net/l/quote/mbjlrkke). pretty cool!

PR #1128, opened 2026-04-30
fix(oas): correctly resolve deeply nested ref schemas in allof merge

maximilianfalco:
| 🚥 Resolves CX-3312 |
| :------------------- |

## 🧰 Changes

This is a follow-up to CX-3213. That fix made `inlinePropertyRefsForMerge` recurse into nested `properties` objects, but only when the property *value* was already concrete. When the value was itself a `$ref` likely chained `$refs`, it inlined one hop and stopped. 

If two `allOf` branches reach the same property path through chained refs, the surviving inner `$ref` collides with the other branch's content during merge, and downstream `$ref`-sibling cleanup strips everyone's properties.

<details>
<summary>Walkthrough of the bug with a minimal schema</summary>

### Setup

```jsonc
allOf: [
  // branch A — fully inline
  { properties: { email: { properties: { nested: { properties: { fromA } } } } } },

  // branch B — chain of refs
  { properties: { email: { $ref: "EmailRef" } } }
]

EmailRef    = { properties: { nested: { $ref: "EmailNested" } } }   // ← chained
EmailNested = { properties: { fromB } }
```

Both branches reach `email.nested`.

### Step 1 — old single-hop inlining on branch B

```jsonc
B = {
  properties: {
    email: {                            // ← $ref replaced (one hop)
      properties: {
        nested: { $ref: "EmailNested" } // ← survived, NOT inlined
      }
    }
  }
}
```

### Step 2 — what the merge sees at `email.nested`

```jsonc
A.email.nested = { properties: { fromA } }
B.email.nested = { $ref: "EmailNested" }
```

### Step 3 — merge produces

```jsonc
email.nested = {
  $ref: "EmailNested",
  properties: { fromA }   // briefly survives
}
```

### Step 4 — `$ref`-sibling cleanup runs

OpenAPI 3.0 says `$ref` can't have siblings, so cleanup strips them:

```jsonc
email.nested = {
  $ref: "EmailNested"
  // ✗ properties: { fromA }   stripped
}
```

The renderer follows `$ref` to `EmailNested`, gets `{ fromB }`, and shows only that. `fromA` is gone.

That's the bug: a `$ref` survived past one inlining hop, hit a conflict path, and triggered the sibling-strip.

</details>

## 🛠 The fix

Introduces a `collectConflictPaths` helper that walks every `allOf` branch, following `$ref`s through the existing `usedSchemas` cache and returns property paths reached by 2+ branches. 

`inlinePropertyRefsForMerge` then takes that set as input: at conflict paths it inlines AND recurses into the just-inlined value 

> [!NOTE]
the deep `$ref` resolution is only fixing UP UNTIL the conflict paths, once in a non conflicting path it hops ONE last time and then preserves the rest. this is inline with the strategy `toJSONSchema` currently has. we could in theory recurse to resolve it entirely but then it may very likely bloat the schema and produce unnecessarily large schemas

### DIFF
--- a/.changeset/cx-3312-deep-ref-chain-allof-merge.md
+++ b/.changeset/cx-3312-deep-ref-chain-allof-merge.md
@@ -2,4 +2,4 @@
 'oas': patch
 ---
 
-Deep-merge `allOf` branches when nested properties at the same path resolve through a chain of `$ref` schemas. Previously `inlinePropertyRefsForMerge` only inlined one ref level deep; if two branches reached the same property path through chained refs, the inner refs survived into `json-schema-merge-allof`, which silently kept only the first branch. Now allOf branches are scanned for conflicting property paths and refs along those paths are recursively inlined before merging. Non-conflicting refs are still preserved.
+Deep-merge `allOf` branches when nested properties at the same path resolve through a chain of `$ref` schemas. Previously `inlinePropertyRefsForMerge` only inlined one ref level deep; if two branches reached the same property path through chained refs, the inner refs survived into `json-schema-merge-allof`, which silently kept only the first branch. Now `allOf` branches are scanned for conflicting property paths and refs along those paths are recursively inlined before merging. Non-conflicting refs are still preserved.
--- a/.changeset/cx-3312-deep-ref-chain-allof-merge.md
+++ b/.changeset/cx-3312-deep-ref-chain-allof-merge.md
@@ -1,0 +1,5 @@
+---
+'oas': patch
+---
+
+Deep-merge `allOf` branches when nested properties at the same path resolve through a chain of `$ref` schemas. Previously `inlinePropertyRefsForMerge` only inlined one ref level deep; if two branches reached the same property path through chained refs, the inner refs survived into `json-schema-merge-allof`, which silently kept only the first branch. Now allOf branches are scanned for conflicting property paths and refs along those paths are recursively inlined before merging. Non-conflicting refs are still preserved.
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -248,7 +248,13 @@ function inlinePropertyRefsForMerge(
       }
     } else if (val && typeof val === 'object' && !Array.isArray(val)) {
       if ('properties' in val) {
-        out.properties[key] = inlinePropertyRefsForMerge(val as SchemaObject, usedSchemas, refLogger, conflictPaths, path);
+        out.properties[key] = inlinePropertyRefsForMerge(
+          val as SchemaObject,
+          usedSchemas,
+          refLogger,
+          conflictPaths,
+          path,
+        );
       }
 
       // Inline allOf-path refs inside oneOf/anyOf to prevent synthetic allOf scaffolding.
@@ -269,10 +275,7 @@ function inlinePropertyRefsForMerge(
  * however, in theory, usedSchemas should already contain all the possible $refs since that should
  * be resolved and populated by toJSONSchema, the only caller of this function
  */
-function collectConflictPaths(
-  allOfBranches: SchemaObject[],
-  usedSchemas: Map<string, SchemaObject>,
-): Set<string> {
+function collectConflictPaths(allOfBranches: SchemaObject[], usedSchemas: Map<string, SchemaObject>): Set<string> {
   // Maps each property path to the number of branches that reach it.
   const pathBranchCount = new Map<string, number>();
 
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -224,7 +224,7 @@ function inlinePropertyRefsForMerge(
   usedSchemas: Map<string, SchemaObject>,
   refLogger: NonNullable<toJSONSchemaOptions['refLogger']>,
   conflictPaths?: Set<string>,
-  currentPath = '',
+  currentPath?: string,
 ): SchemaObject {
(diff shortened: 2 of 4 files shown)


.changeset/cx-3312-deep-ref-chain-allof-merge.md
```diff
@@ -0,0 +1,5 @@
+---
+'oas': patch
+---
+
+Deep-merge `allOf` branches when nested properties at the same path resolve through a chain of `$ref` schemas. Previously `inlinePropertyRefsForMerge` only inlined one ref level deep; if two branches reached the same property path through chained refs, the inner refs survived into `json-schema-merge-allof`, which silently kept only the first branch. Now allOf branches are scanned for conflicting property paths and refs along those paths are recursively inlined before merging. Non-conflicting refs are still preserved.
```
erunion:
```suggestion
Deep-merge `allOf` branches when nested properties at the same path resolve through a chain of `$ref` schemas. Previously `inlinePropertyRefsForMerge` only inlined one ref level deep; if two branches reached the same property path through chained refs, the inner refs survived into `json-schema-merge-allof`, which silently kept only the first branch. Now `allOf` branches are scanned for conflicting property paths and refs along those paths are recursively inlined before merging. Non-conflicting refs are still preserved.
```

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -291,6 +309,70 @@ function inlinePropertyRefsForMerge(
...
+      const childPath = path ? `${path}.${propertyName}` : propertyName;
+      recordPath(childPath);
+      walk(childSchema, childPath);
+    }
+  }
+
+  for (const branch of allOfBranches) {
+    pathsSeenInThisBranch.clear();
+    refsCurrentlyResolving.clear();
+    walk(branch, '');
```
erunion: Instead of supplying an empty string for the initial `path` can `path` just be marked as optional?

```suggestion
    walk(branch);
```

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -291,6 +309,70 @@ function inlinePropertyRefsForMerge(
...
+  const pathsSeenInThisBranch = new Set<string>();
+  const refsCurrentlyResolving = new Set<string>();
+
+  function recordPath(path: string): void {
+    if (pathsSeenInThisBranch.has(path)) return;
+    pathsSeenInThisBranch.add(path);
+    pathBranchCount.set(path, (pathBranchCount.get(path) ?? 0) + 1);
+  }
+
+  function walk(schema: unknown, path: string): void {
```
erunion:
```suggestion
  function walk(schema: unknown, path?: string): void {
```

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -223,13 +223,16 @@ function inlinePropertyRefsForMerge(
   schema: SchemaObject,
   usedSchemas: Map<string, SchemaObject>,
   refLogger: NonNullable<toJSONSchemaOptions['refLogger']>,
+  conflictPaths?: Set<string>,
+  currentPath = '',
```
erunion:
```suggestion
  currentPath?: string,
```

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -787,9 +869,12 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
           return toJSONSchema(item as SchemaObject, allOfOptions);
         });
 
+        // Find paths reached by 2+ branches, then inline each branch with deep recursion at
+        // those conflict paths.
+        const conflictPaths = collectConflictPaths(allOfSchemas, usedSchemas);
         schema = {
           ...schema,
-          allOf: allOfSchemas.map(s => inlinePropertyRefsForMerge(s, usedSchemas, refLogger)),
+          allOf: allOfSchemas.map(s => inlinePropertyRefsForMerge(s, usedSchemas, refLogger, conflictPaths, '')),
```
erunion:
```suggestion
          allOf: allOfSchemas.map(s => inlinePropertyRefsForMerge(s, usedSchemas, refLogger, conflictPaths)),
```

erunion (changes requested):
some very minor suggestions but otherwise this lgtm

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: fixed in https://github.com/readmeio/oas/pull/1128/commits/c81b7f4a79d9ba7f16083dee1f7bf9cf3433d844

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: fixed in https://github.com/readmeio/oas/pull/1128/commits/c81b7f4a79d9ba7f16083dee1f7bf9cf3433d844

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: fixed in https://github.com/readmeio/oas/pull/1128/commits/c81b7f4a79d9ba7f16083dee1f7bf9cf3433d844

packages/oas/src/lib/openapi-to-json-schema.ts
maximilianfalco: fixed in https://github.com/readmeio/oas/pull/1128/commits/c81b7f4a79d9ba7f16083dee1f7bf9cf3433d844

erunion (approved):

PR #1133, opened 2026-05-04
feat: allow empty arrays in json request body

HugoHSun:
| 🚥 Resolves [CX-3296](https://linear.app/readme-io/issue/CX-3296/feature-request-ability-to-add-and-send-empty-arrays-in-try-it) | 🤖 Prepared with `/prep-pr` |
| :------------------------------------------------------------------------------------------------------------------------------ | :------------------------- |

## 🧰 Changes

Preserves explicit empty arrays in JSON request bodies when `@readme/oas-to-har` removes undefined values before generating HAR payloads. This covers both regular JSON request bodies and ReadMe `RAW_BODY` object payloads, while still omitting undefined properties.

## 👥 Customer Impact

Customers using Try It! with JSON request examples or defaults that intentionally contain `[]` can send those empty arrays instead of having them stripped from the generated request payload.

## 🧬 QA & Testing

- Test file:  [cx-3296-empty-arrays.json](https://github.com/user-attachments/files/27355755/cx-3296-empty-arrays.json)
- Ensure that empty arrays are preserved in both the code snippet and request

Related downstream PR: https://github.com/readmeio/readme/pull/18593

### DIFF
--- a/.changeset/polite-arrays-smile.md
+++ b/.changeset/polite-arrays-smile.md
@@ -1,0 +1,5 @@
+---
+'@readme/oas-to-har': patch
+---
+
+Preserve explicit empty arrays in JSON request bodies when generating HAR payloads.
--- a/.changeset/polite-arrays-smile.md
+++ b/.changeset/polite-arrays-smile.md
@@ -2,4 +2,4 @@
 '@readme/oas-to-har': patch
 ---
 
-Preserve explicit empty arrays in JSON request bodies when generating HAR payloads.
+Preserve explicit empty arrays in JSON request payloads when generating HAR payloads.
--- a/packages/oas-to-har/src/index.ts
+++ b/packages/oas-to-har/src/index.ts
@@ -162,6 +162,7 @@ function stringify(json: Record<string | 'RAW_BODY', unknown>) {
   return JSON.stringify(
     removeUndefinedObjects(typeof json.RAW_BODY !== 'undefined' ? json.RAW_BODY : json, {
       preserveNullishArrays: true,
+      preserveEmptyArray: true,
     }),
   );
 }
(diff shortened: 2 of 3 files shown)


darrenyong (commented):
do we need to add a changeset to this?

HugoHSun:
> do we need to add a changeset to this?

Yep added!

erunion:
@darrenyong @HugoHSun yes, all changes that need to be published to NPM should have a changeset.

erunion (changes requested):
I'm a bit concerned that if we always send empty arrays we may break some API expectations because an empty array could be inferred as “delete everything here” rather than “don’t update anything”. Like if I set content into this form and then remove it `array<string>` is removed from my payload because it’s an empty array. this change makes it seem like we’ll be sending an empty array instead without a way to _not_ do that.

Image

HugoHSun:
> I'm a bit concerned that if we always send empty arrays we may break some API expectations because an empty array could be inferred as “delete everything here” rather than “don’t update anything”. Like if I set content into this form and then remove it `array<string>` is removed from my payload because it’s an empty array. this change makes it seem like we’ll be sending an empty array instead without a way to _not_ do that.
> 
> Image

Yep that's why this PR https://github.com/readmeio/readme/pull/18593 sets an array field to undefined rather than [] when all array items are removed. And if the user wanna explicitly send [], they add an item without filling any values in the array item.

.changeset/polite-arrays-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@readme/oas-to-har': patch
+---
+
+Preserve explicit empty arrays in JSON request bodies when generating HAR payloads.
```
erunion:
```suggestion
Preserve explicit empty arrays in JSON request payloads when generating HAR payloads.
```

erunion (approved):

PR #1135, opened 2026-05-06
fix(oas-to-har): honor additionalProperties schemas

HugoHSun:
| 🚥 Resolves CX-3311 |
| :------------------- |

## 🧰 Changes

Try It was coercing digits-only values inside `additionalProperties` maps from strings into numbers when generating the HAR/cURL request body.

The root cause was in the JSON request body serialization path. We already used named `properties` schemas to decide whether a string should stay a string or be parsed as JSON. But for dynamic object keys, like `custom_fields.employeeID`, we weren't using the schema from `additionalProperties`.

So a schema like this:

```json
{
  "custom_fields": {
    "type": "object",
    "additionalProperties": {
      "type": "string"
    }
  }
}
```

still fell back to loose JSON parsing for `employeeID`. Since `JSON.parse("1009")` returns `1009`, the generated request body sent a number instead of the string required by the OpenAPI schema.

This updates `oas-to-har` to treat schema-object `additionalProperties` as the schema for dynamic keys. Named properties still take precedence, but unknown keys now honor `additionalProperties` when it defines a schema or `$ref`.

## 🧬 QA & Testing

OAS: [cx-3311-custom-fields-string-values.json](https://github.com/user-attachments/files/27439720/cx-3311-custom-fields-string-values.json)

- `additionalProperties: { type: "string" }` preserves digits-only values as strings.
image

- `additionalProperties` with a `$ref` to a string schema preserves digits-only values as strings.
image

- named `properties` still take precedence over `additionalProperties`.
- mixed schemas still coerce named number/boolean properties correctly.
image

- existing `format: json` request body behavior still parses JSON strings where expected.
image

### DIFF
--- a/.changeset/quiet-fields-parse.md
+++ b/.changeset/quiet-fields-parse.md
@@ -1,0 +1,5 @@
+---
+'@readme/oas-to-har': patch
+---
+
+Preserve numeric strings in JSON request body maps when `additionalProperties` defines string values.
--- a/packages/oas-to-har/src/lib/utils.ts
+++ b/packages/oas-to-har/src/lib/utils.ts
@@ -365,6 +365,12 @@ function parseJSONStringsObjectContainersOnly(obj: unknown): unknown {
   return obj;
 }
 
+function getAdditionalPropertiesSchema(schema: SchemaObject): SchemaObject | undefined {
+  return schema.additionalProperties && typeof schema.additionalProperties === 'object'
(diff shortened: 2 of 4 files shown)


erunion (changes requested):
Can you add a test for `{ additionalProperties: { type: "number" } }` to ensure that those are handled accordingly as well?

erunion (approved):

PR #1146, opened 2026-05-28
fix(oas): preserve schema shape when allOf merge throws on a nested conflict

maximilianfalco:
| 🚥 Resolves RM-4285 |
| :------------------- |

## 🧰 Changes

When `toJSONSchema` runs an `allOf` through `json-schema-merge-allof` and the merger throws on an irreconcilable conflict (eg. `properties.documents.type` being `array` in one branch and `object` in another, as in the customer spec behind RM-4285), the previous `catch` block deleted the entire `allOf` and returned the wrapper schema with no `type` or `properties` at all and leaves the plain "json" placeholder

This replaces that lossy fallback in with a best-effort shallow merge.

> The change is additive on the success path, only the throw fallback is touched

## 🧬 QA & Testing

Use this doc: [minimal-repro.json](https://github.com/user-attachments/files/28334976/minimal-repro.json)

https://github.com/user-attachments/assets/4e521de8-eb82-40c9-b772-3bb8f60047f9

### DIFF
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -901,14 +901,40 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
       try {
         schema = mergeJSONSchemaAllOf(schema as JSONSchema, mergeAllOfSchemasOptions) as SchemaObject;
       } catch {
-        // If we can't merge the `allOf` for whatever reason (like if one item is a `string` and
-        // the other is a `object`) then we should completely remove it from the schema and continue
-        // with whatever we've got. Why? If we don't, any tooling that's ingesting this will need
-        // to account for the incompatible `allOf` and it may be subject to more breakages than
-        // just not having it present would be.
-        const { ...schemaWithoutAllOf } = schema;
-        schema = schemaWithoutAllOf as SchemaObject;
-        delete schema.allOf;
+        // The merge can throw on irreconcilable conflicts (eg. `properties.foo.type` being `array`
+        // in one branch and `object` in another). Dropping the entire `allOf` here would leave the
+        // schema with no `type` / `properties` at all, and downstream renderers would have nothing
+        // to show. Instead, do a best-effort shallow merge of each branch's `properties` (later
+        // branches win on conflict) and union `required`, so consumers still see most of the shape.
+        const branches = (schema.allOf as SchemaObject[]) ?? [];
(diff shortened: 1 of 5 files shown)


packages/oas/test/lib/openapi-to-json-schema.test.ts
```diff
@@ -731,25 +736,98 @@ describe('toJSONSchema()', () => {
       expect(toJSONSchema(schema).type).toBeUndefined();
     });
 
+    it('should infer a `type` on an `allOf` schema when the fallback fires with properties', () => {
+      const schema: SchemaObject = {
+        allOf: [
+          {
+            title: 'range_query_specs',
```
erunion, line 743: Should this be retained in the resulting object?

maximilianfalco: oh right, thanks for catching that!

maximilianfalco: fixed in [5b9fc13](https://github.com/readmeio/oas/pull/1146/commits/5b9fc13f07aeb2bc9b5a4ce21b58b657c774e2a2)

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -901,14 +901,46 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
...
+        const requiredSet = new Set<string>(Array.isArray(fallback.required) ? (fallback.required as string[]) : []);
+        for (const branch of branches) {
+          if (!branch || typeof branch !== 'object' || isRef(branch)) continue;
+          if (fallback.type === undefined && 'type' in branch && branch.type !== undefined) {
+            fallback.type = branch.type;
+          }
+          if (fallback.title === undefined && typeof branch.title === 'string' && branch.title.length > 0) {
+            fallback.title = branch.title;
+          }
+          if (typeof branch.description === 'string' && branch.description.length > 0) {
```
erunion: why don't we care about `fallback.description` here like `fallback.title`?

maximilianfalco: oh right sorry about that, hadnt realised it was using the last-win. https://github.com/readmeio/oas/pull/1146/commits/9f455d2866ae01bd0d60793113292d46b777584e made it follow the first-win convention like `title`

erunion (approved):
this logic seems pretty nuts but it looks to do what it should

PR #1150, opened 2026-06-04
fix(oas): prefer schema examples before generating samples

HugoHSun:
| 🚥 Resolves CX-3480 |
| :------------------- |

## 🧰 Changes

Schema-level response examples now take priority before sample generation runs through `allOf` / `oneOf` / `anyOf` branches.

This fixes composed response schemas where a root-level `example` could be skipped during `allOf` merging, causing the generated example to collapse to only the first referenced branch instead of showing the full authored payload.

## 🧬 QA & Testing

Added regression coverage for a nested `allOf` response schema with a root-level schema example.

Verified the response example now returns the full schema-level example instead of generating from only the nested `$ref` branch.


packages/oas/src/samples/index.ts
```diff
@@ -137,6 +137,21 @@ function sampleFromResolvedSchema(
...
 ) {
   let { type } = schema;
+  const { example, additionalProperties, properties, items } = schema;
+  const { includeReadOnly, includeWriteOnly } = opts;
+
+  if (example !== undefined) {
+    // Schema-level examples are explicit authored data, so they should win before we try to merge
+    // polymorphic schemas. Merged allOf schemas can retain a `$ref` alongside sibling metadata,
+    // which would otherwise be dereferenced as a plain ref and drop this example.
+    const cleanedExample = deeplyStripKey(example, '$$ref', (val: unknown): val is string => {
```
erunion: It's been **years** since I ported this code over from Swagger Editor but do you know what this `$$ref` thing is, I'm struggling to remember why and what this is for?  Digging through the library these are the only two areas of `$$ref` are this and tests for it, but the tests don't explain _what_ it is.

erunion: Oh it's something that `swagger-client` adds when resolving API definitions with its `allowMetaPatches` option. 

https://github.com/swagger-api/swagger-js/blob/master/docs/usage/openapi-definition-resolver.md

Can we remove it?

erunion (changes requested):

packages/oas/src/samples/index.ts
darrenyong: removed `deeplyStripKey` and the associated tests that were testing for that

darrenyong (approved):

erunion (approved):

PR #1152, opened 2026-06-12
fix(oas-to-snippet): disable curl globbing for literal query params

HugoHSun:
| 🚥 Resolves RM-10215 |
| :------------------- |

## 🧰 Changes

Fixes generated cURL snippets for query params that render with raw bracket syntax. cURL treats `[]` and `{}` as [URL globbing syntax](https://everything.curl.dev/cmdline/urls/globbing.html), so copied snippets could fail before the request was sent.

Since generated snippets represent one literal request, this adds `--globoff` when the request URL or query params contain cURL glob characters. This keeps the URL readable without URL-encoding or backslash-escaping the query param names.

Note that `--globoff` won't appear when url doesn't have raw brackets even though it's harmless.  

## 🧬 QA & Testing

image

```
curl --request GET \
     --globoff \
     --url 'https://httpbin.org/anything?filter[desiredSecurityProfileStatus]=pending'
     
{
  "args": {
    "filter[desiredSecurityProfileStatus]": "pending"
  }, 
  "data": "", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "*/*", 
    "Host": "httpbin.org", 
    "User-Agent": "curl/8.7.1", 
    "X-Amzn-Trace-Id": "Root=1-6a2bae60-3ace9f8b6861c8cb410e21e9"
  }, 
  "json": null, 
  "method": "GET", 
  "origin": "", 
  "url": "https://httpbin.org/anything?filter[desiredSecurityProfileStatus]=pending"
}
```

### DIFF
--- a/packages/oas-to-snippet/src/index.ts
+++ b/packages/oas-to-snippet/src/index.ts
@@ -10,6 +10,26 @@ import generateHar from '@readme/oas-to-har';
 
 import { getClientInstallationInstructions, getLanguageConfig, getSupportedLanguages } from './languages.js';
 
+type HarForSnippet = HarRequest | ReturnType<typeof generateHar>;
+
+function containsCurlGlobSyntax(value: string) {
+  return value.includes('[') || value.includes(']') || value.includes('{') || value.includes('}');
+}
+
+function hasCurlGlobSyntax(har: HarForSnippet) {
+  const requests = 'log' in har ? har.log.entries.map(entry => entry.request) : [har];
+
+  return requests.some(request => {
+    if (typeof request.url === 'string' && containsCurlGlobSyntax(request.url)) {
+      return true;
+    }
+
+    return request.queryString?.some(
+      ({ name, value }) => containsCurlGlobSyntax(name) || containsCurlGlobSyntax(value),
+    );
+  });
+}
+
 export default function oasToSnippet(
   oas: Oas,
   operation: Operation,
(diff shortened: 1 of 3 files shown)


packages/oas-to-snippet/src/index.ts
```diff
@@ -112,9 +132,16 @@ export default function oasToSnippet(
...
 
-  let targetOpts = config.httpsnippet.targets[target].opts || {};
+  let targetOpts = { ...config.httpsnippet.targets[target].opts };
   const highlightMode = config.highlight;
 
+  if (language === 'shell' && target === 'curl' && hasCurlGlobSyntax(har)) {
+    // Generated snippets represent one literal request, so raw `[]` and `{}` should not be
+    // interpreted by cURL as URL glob syntax.
+    // https://everything.curl.dev/cmdline/urls/globbing.html
+    targetOpts.globOff = true;
```
darrenyong, line 142: is it possible to make it so a `-g` flag instead of `--globOff` — wonder if that'd be cleaner?

HugoHSun: Yea we could use the short form option, the caveat is that it's global tho. So it would also switch flags like `--request` to `-X`, `--header` to `-H`, etc. Which I feel like the other parts are less readable

erunion (approved):

PR #1153, opened 2026-06-15
feat(oas): support x-default for basic auth

HugoHSun:
| 🚥 Resolves RM-42 |
| :------------------- |

## 🧰 Changes

Adds support for using `x-default` with HTTP Basic auth schemes. `oas.getAuth()` can now use an object-shaped default like `{ "user": "...", "pass": "..." }` to prefill Basic auth credentials when no user-provided auth is available. Cookie auth already flows through the existing `apiKey` `x-default` path, so this only adds the missing Basic auth object handling.

User-provided credentials still take precedence over the spec default, so existing custom-login/user-data behavior should keep winning when present.

## 🧬 QA & Testing

- Confirmed that the extension is supported for the new shape and user input takes precedence
[rm-42-x-default-auth-httpbin.json](https://github.com/user-attachments/files/28955981/rm-42-x-default-auth-httpbin.json)

https://github.com/user-attachments/assets/49c19f7a-e758-490b-92da-ae2d0bd584be

### DIFF
--- a/packages/oas/src/types.ts
+++ b/packages/oas/src/types.ts
@@ -288,7 +288,7 @@ export type KeyedSecuritySchemeObject = SecuritySchemeObject & {
 
   // `x-default` is our custom extension for specifying auth defaults.
   // https://docs.readme.com/docs/openapi-extensions#authentication-defaults
-  'x-default'?: number | string;
+  'x-default'?: number | string | { pass?: number | string; user?: number | string };
 };
 
 /**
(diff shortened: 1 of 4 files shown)


packages/oas/src/types.ts
```diff
@@ -288,7 +288,7 @@ export type KeyedSecuritySchemeObject = SecuritySchemeObject & {
 
   // `x-default` is our custom extension for specifying auth defaults.
   // https://docs.readme.com/docs/openapi-extensions#authentication-defaults
```
erunion, line 290: Please make sure to update these docs after this is live.

erunion (approved):

PR #1156, opened 2026-06-17
fix: support `example` keyword with `$ref` pointers

darrenyong:
| 🚥 Resolves CX-3542 |
| :------------------- |

## 🧰 Changes
If a parameter uses a `$ref` pointer and has an `example` keyword, we currently don't do anything with it. The changes here converts it to the `examples` keyword since the singular version is deprecated and use of it is discouraged.

## 🧬 QA & Testing

### DIFF
--- a/.changeset/fifty-falcons-sink.md
+++ b/.changeset/fifty-falcons-sink.md
@@ -1,5 +1,5 @@
 ---
-"oas": major
+"oas": minor
 ---
 
 Add support for `example` keyword with `$ref` pointers
--- a/.changeset/fifty-falcons-sink.md
+++ b/.changeset/fifty-falcons-sink.md
@@ -1,0 +1,5 @@
+---
+"oas": major
+---
+
+Add support for `example` keyword with `$ref` pointers
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -762,6 +762,15 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
       // level as `$ref` however is invalid in JSON Schema and should be ignored.
       const { $ref: _$ref, properties: _propertiesWithRef, ...siblings } = schema as Record<string, unknown>;
       if (Object.keys(siblings).length > 0) {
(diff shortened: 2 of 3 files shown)


packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -762,6 +762,15 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
       // level as `$ref` however is invalid in JSON Schema and should be ignored.
       const { $ref: _$ref, properties: _propertiesWithRef, ...siblings } = schema as Record<string, unknown>;
       if (Object.keys(siblings).length > 0) {
+        // An `example` sibling (common when a parameter-level `example` is merged onto a `$ref`
+        // schema) never reaches the `isSchema` normalization branch below, so normalize it here:
+        // promote primitives to the JSON Schema `examples` array and drop non-primitives.
+        if ('example' in siblings) {
+          if (isPrimitive(siblings.example)) {
+            siblings.examples = [siblings.example];
+          }
```
HugoHSun, line 771: Maybe out of scope for this ticket but looks like the [non-ref schema path](https://github.com/readmeio/oas/blob/79bdac2310f883ef20a6637ac89bc8330de9482d/packages/oas/src/lib/openapi-to-json-schema.ts#L1235-L1240) also handles `example` as an array, e.g. `example: ['active', 'inactive']`. Could be nice to mirror that here so `$ref` sibling examples behave the same way

.changeset/fifty-falcons-sink.md
```diff
@@ -0,0 +1,5 @@
+---
+"oas": major
```
erunion: Is this a breaking change or just a minor bug fix?

darrenyong: oh no this should be a minor one. not sure why `npm run changeset` did a major one 🤔

erunion:
```suggestion
"oas": minor
```

erunion (approved):

packages/oas/src/lib/openapi-to-json-schema.ts
darrenyong: totally missed this, but your suggestions have been added [here](https://github.com/readmeio/oas/pull/1166)!

PR #1157, opened 2026-06-17
feat(oas): support OAuth client credential defaults in x-default

HugoHSun:
| 🚥 Resolves CX-2723 | 
| :------------------- | 

## 🧰 Changes

Adds OAS support for OAuth client credential defaults in `x-default`.

This PR widens the `x-default` security scheme type so OAuth schemes can define client credential defaults:

```yaml
x-default:
  client_id: default-client-id
  client_secret: default-client-secret
```

It also keeps OAuth token defaults and OAuth client credential defaults separate. Primitive `x-default` values still behave as auth token defaults, while object-shaped OAuth defaults are treated as API Explorer client credential metadata and are not returned by `getAuth()` as an auth token.

The paired ReadMe app change is in [readmeio/readme#19277](https://github.com/readmeio/readme/pull/19277), which consumes these defaults to prefill the OAuth client ID and client secret fields.

## 🧬 QA & Testing

Screen recording:

https://github.com/user-attachments/assets/46c494bb-2b75-4761-b9db-98a9737e8f1e

Expected behavior:

- An OAuth scheme with a primitive `x-default` still returns that value from `getAuth()` as the OAuth auth/token default.
- An OAuth scheme with object-shaped `x-default` client credentials does not return that object from `getAuth()` as an auth token.
- The object-shaped `x-default` remains available on the security scheme for the ReadMe app to read.
- When paired with [readmeio/readme#19277](https://github.com/readmeio/readme/pull/19277), `client_id` and `client_secret` prefill the OAuth client ID and client secret fields in API Explorer.
- Existing user-provided auth still takes precedence over `x-default`.

### DIFF
--- a/packages/oas/src/types.ts
+++ b/packages/oas/src/types.ts
@@ -55,6 +55,25 @@ export interface DataForHAR {
 
 export type AuthForHAR = Record<string, number | string | { pass?: string; user?: string }>;
 
+export type SecuritySchemeDefaultAuthPrimitive = number | string;
+
+export interface BasicAuthDefault {
+  pass?: SecuritySchemeDefaultAuthPrimitive;
+  user?: SecuritySchemeDefaultAuthPrimitive;
+}
+
+export interface OAuthClientCredentialsDefault {
+  client_id?: SecuritySchemeDefaultAuthPrimitive;
+  client_secret?: SecuritySchemeDefaultAuthPrimitive;
+  clientId?: SecuritySchemeDefaultAuthPrimitive;
+  clientSecret?: SecuritySchemeDefaultAuthPrimitive;
+}
+
+export type SecuritySchemeDefaultAuth =
+  | SecuritySchemeDefaultAuthPrimitive
+  | BasicAuthDefault
+  | OAuthClientCredentialsDefault;
+
 export interface User {
   [key: string]: unknown;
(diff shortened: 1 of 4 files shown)


packages/oas/src/types.ts
```diff
@@ -55,6 +55,25 @@ export interface DataForHAR {
...
+export interface BasicAuthDefault {
+  pass?: SecuritySchemeDefaultAuthPrimitive;
+  user?: SecuritySchemeDefaultAuthPrimitive;
+}
+
+export interface OAuthClientCredentialsDefault {
+  client_id?: SecuritySchemeDefaultAuthPrimitive;
+  client_secret?: SecuritySchemeDefaultAuthPrimitive;
+  clientId?: SecuritySchemeDefaultAuthPrimitive;
+  clientSecret?: SecuritySchemeDefaultAuthPrimitive;
```
erunion, line 69: Why do we need to support snake_case and camelCase?

HugoHSun: snake_case is the OAuth protocol parameter names and camelCase is the convention from other extensions - but ultimately just depends on preferences i guess

darrenyong (approved):

erunion (approved):

PR #1158, opened 2026-06-18
fix(oas): ignore reserved header parameters in schema output

HugoHSun:
| 🚥 Resolves CX-3544 |
| :------------------- |

## 🧰 Changes

API Explorer was rendering reserved header parameters like `Accept`, `Content-Type`, and `Authorization` when they were defined as custom `header` params in an OpenAPI file. Those names are reserved by OAS and should be ignored as custom header parameters.

This filters those reserved header params out of `getParametersAsJSONSchema()` so they don't show up as editable request fields or interfere with request body rendering.

Normal custom headers still render, and the reserved names are only ignored when they're defined as header parameters.

## 🧬 QA & Testing

Added coverage for the CX-3544 cases:

- reserved header params are ignored case-insensitively
- normal headers like `X-Request-ID` still render
- reserved names in non-header locations still render
- request body schemas are preserved when a reserved `Content-Type` header param is ignored

### DIFF
--- a/packages/oas/src/operation/transformers/get-parameters-as-json-schema.ts
+++ b/packages/oas/src/operation/transformers/get-parameters-as-json-schema.ts
@@ -17,6 +17,12 @@ import {
 } from '../../lib/refs.js';
 import { isRef } from '../../types.js';
 
+const RESERVED_HEADER_PARAMETERS = new Set(['accept', 'authorization', 'content-type']);
+
+function isReservedHeaderParameter(param: ParameterObject) {
+  return param.in === 'header' && RESERVED_HEADER_PARAMETERS.has(param.name.toLowerCase());
+}
+
 /**
  * The order of this object determines how they will be sorted in the compiled JSON Schema
  * representation.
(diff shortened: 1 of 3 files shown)


packages/oas/src/operation/transformers/get-parameters-as-json-schema.ts
```diff
@@ -177,7 +183,11 @@ export function getParametersAsJSONSchema(
 
         // This `as` actually *could* be a ref, but we don't want refs to pass through here, so
         // `.in` will never match `type`
-        const parameters = operationParams.filter(param => (param as ParameterObject).in === type);
+        const parameters = operationParams.filter(param => {
+          const parameter = param as ParameterObject;
```
erunion: Can you avoid force casting here and rely on type narrowing instead? Im assuming you're doing this because this can be a parameter or reference object?

erunion (changes requested):

packages/oas/src/operation/transformers/get-parameters-as-json-schema.ts
HugoHSun: Yep looks like it can only be parameter object now so removed the cast

erunion (approved):

PR #1159, opened 2026-06-19
feat(oas): add operation-aware server helpers

HugoHSun:
| 🚥 Resolves RM-1745 |
| :------------------- |

## 🧰 Changes

Adds operation-aware server helpers to `oas` so consumers can resolve the correct server list for an operation using OpenAPI precedence:

1. operation-level `servers`
2. path-item `servers`
3. root-level `servers`
4. `https://example.com` fallback

This adds `getServers()` to `Oas` for root-level server access, and adds operation-aware helpers on `Operation` for server lists, URLs, variables, default variables, and split URL parts.

This keeps the existing root-level `Oas#url()` behavior intact, while giving API Explorer and related packages a cleaner API to use when an endpoint overrides its server at the path or operation level.

This also updates `@readme/oas-to-har` to build request URLs from the operation-aware server helpers, so generated HAR requests use path-item and operation-level servers instead of always using the root server.

## 🧬 QA & Testing

Added coverage for:

- root, path-item, and operation-level server precedence
- multiple operation-level servers
- server variables and user variable overrides
- relative path-item and operation server URLs
- path-item `$ref` server resolution
- empty operation/path server arrays falling through to the next level
- missing server URLs falling back to `https://example.com`
- HAR request URLs using path-item and operation-level servers
- HAR request URLs respecting selected operation servers
- HAR request URLs applying supplied operation server variables
- plain-object operation schemas preserving path-item server resolution


erunion (commented):
review incoming

packages/oas-to-har/src/index.ts
```diff
@@ -274,7 +274,9 @@ export default function oasToHar(
      *
      * It's weird. This is easier.
      */
-    const currentOas = Oas.init(oas as unknown as OASDocument);
+    // Preserve an existing Oas instance so operation-level URL helpers can resolve servers from
+    // its underlying API definition instead of wrapping the instance itself as raw OAS JSON.
+    const currentOas = typeof oas.getDefinition === 'function' ? oas : Oas.init(oas as unknown as OASDocument);
```
erunion: I've read this comment a dozen times and still don't totally understand this.

packages/oas/test/index.test.ts
```diff
@@ -615,6 +628,144 @@ describe('Oas', () => {
     });
   });
 
+  describe('Operation server helpers', () => {
```
erunion: These tests need to be moved to `operation/index.test.ts`.

packages/oas/test/__datasets__/operation-server-precedence.json
erunion, line 1: So we aren't duplicating this multiple times cfan you move this example spec into `oas-examples` to be alongside the `server-variables` example we have there?

erunion (changes requested):

packages/oas/src/lib/urls.ts
```diff
@@ -27,6 +70,14 @@ export function stripTrailingSlash(url: string): string {
   return url;
 }
 
+/**
+ * Retrieve a usable list of servers, falling back to the same default URL that `normalizedURL()`
+ * uses when an OpenAPI definition has no usable server data.
+ */
+export function getServers(servers?: ServerObject[]): ServerObject[] {
+  return servers?.length ? servers : [{ url: 'https://example.com' }];
```
darrenyong: is there a reason we ever need to return `https://example.com` here? if there aren't any `servers`, this line in `variablesFromServers` will always return an empty object right?

```
getServers(servers)[selected]?.variables || {}
```

packages/oas/src/index.ts
```diff
@@ -136,89 +140,25 @@ export default class Oas {
     return this.api;
   }
 
+  getServers(): ServerObject[] {
+    return getServerList(this.api.servers);
+  }
+
   url(selected = 0, variables?: ServerVariable): string {
     const url = normalizedURL(this.api, selected);
```
darrenyong: @erunion thoughts on changing this to `normalizedURLFromServers`? looks like this is the only place that's using `normalizedURL` and that function is just a wrapper for `normalizedURLFromServers`. that way we can just remove the wrapper in [packages/oas/src/lib/urls.ts](https://github.com/readmeio/oas/pull/1159/files#diff-732cb3014650d403001235cb71de01baf45d8f678b3392f9ba8a7d436161f6f0R131)

erunion: good idea

packages/oas-to-har/src/index.ts
HugoHSun: If `oas` is already an `Oas` instance and we pass it through `Oas.init()` again, the real definition ends up at `oas.api.api` instead of `oas.api`.

That breaks any `Operation` methods that rely on the linked `Oas` instance for the underlying API definition, like the new server helpers.

Looking at the signature here though, `oas` should only be an `Oas` instance, so I removed the extra guard and just pass the existing instance through.

packages/oas/src/lib/urls.ts
HugoHSun: removed this function since we just do `servers || []`

packages/oas/src/index.ts
HugoHSun: removed the wrapper and calls `normalizedURLFromServers` straight away

packages/oas/test/__datasets__/operation-server-precedence.json
HugoHSun: have deduped the fixtures for both `oas` and `oas-to-har`

packages/oas-examples/3.0/json/server-variables.json
```diff
@@ -168,6 +168,70 @@
           }
         }
       }
+    },
+    "/relative-path-server": {
```
erunion: Sorry, no I meant to move these into a new `server.json` example in `packages/oas-examples/3.0/json`.

erunion (changes requested):

packages/oas-examples/3.0/json/server-variables.json
darrenyong: added them to a `server-path-level.json` and `server-path-level.yaml` file

darrenyong (dismissed):

HugoHSun:
alright the test fixtures should be sorted now @erunion

erunion (approved):

PR #1163, opened 2026-07-01
fix(oas-normalize): dont allow unsafe URLs to be accessed

erunion:
## 🧰 Changes

This resolves an oversight in `oas-normalize` where it did not use `@readme/openapi-parser`'s HTTP fetching mechanisms, had its own, and could load private URLs.

Instead of using `@readme/openapi-parser`'s HTTP fetching mechanism, which would have required a pretty large overhaul to `oas-normalize` I'm instead exposing the unsafe URL method that we have access to via [@apidevtools/json-schema-ref-parser](https://npm.im/@apidevtools/json-schema-ref-parser) (and writing our own tests for it).

### DIFF
--- a/packages/parser/src/lib/urls.ts
+++ b/packages/parser/src/lib/urls.ts
@@ -1,0 +1,11 @@
+import { isUnsafeUrl } from '@apidevtools/json-schema-ref-parser';
+
+/**
+ * Determine if a given URL is unsafe.
+ *
+ * An unsafe URL is one that is not publicly accessible (no private IPs, private ports, etc.)
+ *
+ */
+export function isUnsafeURL(url: string): boolean {
+  return isUnsafeUrl(url);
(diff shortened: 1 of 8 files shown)


packages/parser/src/lib/urls.ts
domharrington, line 1: Is this re-export preferable vs just importing '@apidevtools/json-schema-ref-parser' directly from packages/oas-normalize/src/index.ts? Is it to add types or something?

domharrington (approved):

HugoHSun (approved):

packages/parser/src/lib/urls.ts
erunion: Doing this so I don't have to manage the `@apidevtools/json-schema-ref-parser` dependency in two packages.

PR #1164, opened 2026-07-01
feat(oas): add `status-url` ReadMe extension

maximilianfalco:
| 🚥 Resolves RM-17218 |
| :------------------- |

## 🧰 Changes

In efforts to implement [RFC-9727](https://www.rfc-editor.org/info/rfc9727/), we want to define a new custom `status-url` extension under the `x-readme` convention so we can keep track of the health URLs of OAS files.

```jsonc
{
  "x-readme": {
    "status-url": "https://status.example.com"
  }
}
```

The goal is to eventually surface these as the RFC 9727 `status` link relation in a project's `/.well-known/api-catalog`, so agents have a way to _'health-check'_ APIs without us needing to do it for them. This PR only defines the extension and its default (`undefined`); surfacing it is separate follow-up work.

## 🧬 QA & Testing

Covered by unit tests in `packages/oas/test/extensions.test.ts`: asserts the extension has a default value and that `getExtension` locates it under `x-readme`.

### DIFF
--- a/.changeset/status-url-extension.md
+++ b/.changeset/status-url-extension.md
@@ -1,0 +1,5 @@
+---
+"oas": minor
+---
+
+Add support for the `status-url` ReadMe extension (`x-readme.status-url`). This customer-declared health/status URL is surfaced to AI agents as the RFC 9727 `status` link relation in the project's `/.well-known/api-catalog`.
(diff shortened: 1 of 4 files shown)


.changeset/status-url-extension.md
erunion, line 1: Can you remove our ticket number from this? Its not publicly accessible and anyone referencing this later can use the PR instead.

erunion (dismissed):
Minor request but lgtm

maximilianfalco:
> Minor request but lgtm

oh whoops i think it made your review stale

erunion (approved):

PR #1166, opened 2026-07-01
feat: support examples in `ref` pointers

darrenyong:
| 🚥 Resolves CX-3637 |
| :------------------- |

## 🧰 Changes
In my previous [PR](https://github.com/readmeio/oas/pull/1156) to add support for  the `example` keyword in `$ref` pointers, I didn't add any for the `examples` case like @HugoHSun suggested. This here is a follow up to that and adds support for the `examples` keyword.

Additionally, since the logic is being duplicated in the `$ref` pointer case and the non-ref pointer case, I've DRY'd up the code a bit into a function called `convertExamples()`. 

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/src/lib/openapi-to-json-schema.ts
+++ b/packages/oas/src/lib/openapi-to-json-schema.ts
@@ -742,7 +742,7 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
     refsEmittedAsStub,
   };
 
-  function convertExamples(obj: Record<string, unknown>): void {
+  function convertExamples(obj: SchemaObject): void {
     if ('example' in obj) {
       if (isPrimitive(obj.example)) {
         obj.examples = [obj.example];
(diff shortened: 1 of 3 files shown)


packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -742,6 +742,61 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
     refsEmittedAsStub,
   };
 
+  function convertExamples(obj: Record<string, unknown>): void {
```
erunion: Can `obj` be typed as a `SchemaObject`?

erunion (commented):
can you add a changeset for this? the release flow won't create a package bump without it

darrenyong:
🤦 I keep forgetting the changeset. adding that now!

packages/oas/src/lib/openapi-to-json-schema.ts
darrenyong: updated in https://github.com/readmeio/oas/pull/1166/commits/5dbe0471e4dce0a19d2114f2fc8356d7d7cb12a3

erunion (approved):

PR #1169, opened 2026-07-02
test(oas-to-har): cover optional auth alternatives

kriptoburak:
## Summary

- Add an OAS-to-HAR fixture route whose operation allows anonymous access or API key auth.
- Assert supplied header auth is still serialized when one security alternative is empty.

## Validation

- Git diff whitespace check passed.
- Targeted Vitest run could not start because this temporary checkout has no installed Vitest dependency.


erunion (approved):
thank you!

don't worry about a changeset for this since it's just a test change.

PR #1170, opened 2026-07-06
perf(oas): analyze a single operation without reducing the definition first

erunion:
> [!IMPORTANT]
> This is contains breaking changes for `oas` and will be published to a new major release.

## 🧰 Changes

Analyzing a large OpenAPI definition operation-by-operation used to require cloning and reducing the full document with `oas/reducer` for each operation, then running `analyzer()` on that reduced copy — re-dereferencing and re-scanning from scratch every time. On large specs with hundreds of operations, that added up to minutes in production.

This refactor adds scoped analysis against the full, unreduced definition, with caching so the expensive work is paid once per definition reference.

### New entry points:
* `analyzeOperation(definition, { path, method, query? })` — analyze a single path operation and anything it references
* `analyzeWebhookOperation(definition, { webhookName, method, query? })` — same for OpenAPI 3.1 webhooks
Each call computes an OperationScope: the operation (or webhook operation), path-level common parameters, and every $ref transitively reachable from those — mirroring the ref-walk OpenAPIReducer already does, but read-only and without cloning the document.

Each call computes an `OperationScope`: the operation (or webhook operation), path-level common parameters, and every `$ref` transitively reachable from those — mirroring the ref-walk `OpenAPIReducer` already does, but read-only and without cloning the document.

### Performance
Two caches keyed by the original `definition` object reference:

* `dereferenceOasShared()` — clones and dereferences once; reuse across calls
* `queryCached()` — runs each JSONPath scan once per definition; filter results per operation
 
**Important:** pass the same `definition` reference across calls. Fresh clones defeat the cache.

Benchmark (`test/benchmarks/analyzer.bench.ts`), 50 operations from a ~2.8MB DocuSign spec:

| Approach | Time (3 iterations) |
| :--- | :--- |
| OpenAPIReducer.reduce() + analyzer() per operation | ~2564ms |
| analyzeOperation() against the full definition | ~210ms |

**~12× faster**, with the gap widening as operation count grows.

### Selective queries
`analyzer()` and the new scoped helpers accept an optional query array to run only the metrics you need:

```ts
await analyzer(petstore, ['polymorphism', 'references']);
await analyzeOperation(petstore, { path: '/pet', method: 'post', query: ['references'] });
```

### Query improvements
* `mediaTypes` now scans `components.requestBodies`, `components.responses`, and `webhooks` in addition to `paths` (fixes missed media types behind `$ref`)
* `additionalProperties` correctly finds usage inside $ref pointers

## 💥 Breaking changes
| Before | After |
| :--- | :--- |
| `analysis.general.*` / `analysis.openapi.*` | Flat top-level keys (analysis.mediaTypes, analysis.references, etc.) |
| `rawFileSize`, `dereferencedFileSize` | Removed |
| Individual query functions exported from `oas/analyzer` |No longer exported; use `analyzer(def, ['queryName'])` instead |

### DIFF
--- a/packages/oas/src/analyzer/queries/openapi.ts
+++ b/packages/oas/src/analyzer/queries/openapi.ts
@@ -2,9 +2,10 @@ import type { OASDocument } from '../../types.js';
 import type { OperationScope } from '../scope.js';
 import type { JSONPathResult } from '../util.js';
 
+import { toPointer } from '../../lib/refs.js';
 import { dereferenceOas } from '../dereference.js';
 import { queryCached } from '../query-cache.js';
-import { isAncestorOfScope, isPointerInScope, toPointer } from '../scope.js';
+import { isAncestorOfScope, isPointerInScope } from '../scope.js';
 import { refizePointer } from '../util.js';
 
 /**
--- a/packages/oas/src/analyzer/queries/openapi.ts
+++ b/packages/oas/src/analyzer/queries/openapi.ts
@@ -85,37 +85,6 @@ export function discriminators(definition: OASDocument, scope?: OperationScope):
 }
 
 /**
- * Calculate the size of the raw and dereferenced OpenAPI file in MB.
- *
- * If a dereferenced API definition is too large to be stringified the file size will be returned
- * as NaN.
- *
- */
-export function fileSize(
-  definition: OASDocument,
-  definitionDereferenced: OASDocument,
-): { raw: number; dereferenced: number | typeof NaN } {
-  const originalSizeInBytes = Buffer.from(JSON.stringify(definition)).length;
-  const raw = Number((originalSizeInBytes / (1024 * 1024)).toFixed(2));
-
-  let dereferenced: number;
-  try {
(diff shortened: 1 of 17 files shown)


packages/oas/src/analyzer/queries/openapi.ts
```diff
@@ -109,10 +104,21 @@ export function links(definition: OASDocument): string[] {
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#request-body-object}
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#request-body-object}
  */
-export function mediaTypes(definition: OASDocument): string[] {
+export function mediaTypes(definition: OASDocument, scope?: OperationScope): string[] {
   const results = Array.from(
     new Set(
-      query(['$..paths..content'], definition).flatMap(res => {
```
erunion, line 115: This query changed because if an OpenAPI definition, now that we aren't reducing them anymore, had a `requestBody` that was a `$ref` we wouldn't query into that to pick up whatever the media type it used was.

jboyens (approved):
Only skimmed the tests, but otherwise it all looks good to me.

PR #1172, opened 2026-07-07
feat: support OpenAPI 3.2 typings by migrating to `@scalar/openapi-types`

erunion:
## 🧰 Changes

This replaces the unmaintained `openapi-types` package with `@scalar/openapi-types` in order to allow us to begin working to support OpenAPI 3.2.

We do **not** yet support OpenAPI 3.2 with this work, it just allows us to begin typing data against OpenAPI 3.2 schemas.


packages/oas-normalize/src/index.ts
```diff
@@ -212,9 +212,7 @@ export default class OASNormalize {
           throw new Error('Swagger v1.2 is unsupported.');
         }
 
-        return converter
-          .convertObj(schema, { anchors: true })
-          .then((options: { openapi: OpenAPI.Document }) => options.openapi);
```
erunion, line 217: This had to change because `swagger2openapi`, a package we haven't forked, still uses `openapi-types` and that library's types aren't compatible with `@scalar/openapi-types`.

packages/oas-normalize/src/index.ts
```diff
@@ -326,7 +324,7 @@ export default class OASNormalize {
         case 'openapi':
           return {
             specification: 'openapi',
-            version: (schema as unknown as OpenAPIV3.Document).openapi,
+            version: (schema as OpenAPIV3.Document).openapi ?? 'unknown',
```
erunion, line 327: This is a little funky but `openapi` is typed as optional in this library. I _guess_ that's right because by this point we haven't fully validated the API definition so _technically_ it can be "unknown" if we have garbage data because the `getAPIDefinitionType` check above this that determines we have an `openapi` file just looks if `openapi` is present in the file.

packages/oas-to-har/src/lib/style-formatting/index.ts
```diff
@@ -14,7 +14,10 @@ function shouldNotStyleEmptyValues(parameter: ParameterObject) {
 }
 
 function shouldNotStyleReservedHeader(parameter: ParameterObject) {
-  return ['accept', 'authorization', 'content-type'].includes(parameter.name.toLowerCase());
+  return ['accept', 'authorization', 'content-type'].includes(
+    // @ts-expect-error -- optionally typed but the spec requires it be present. https://github.com/scalar/scalar/issues/9669
+    parameter.name,
+  );
```
erunion, line 20: I filed a ticket for this because it doesn't make sense.

packages/oas-to-har/src/lib/configure-security.ts
```diff
@@ -21,10 +21,8 @@ export default function configureSecurity(
   if (Object.keys(values || {}).length === 0) return undefined;
 
   if (!apiDefinition.components?.securitySchemes?.[scheme]) return undefined;
-  const security = apiDefinition.components.securitySchemes[scheme] as SecuritySchemeObject & {
-    'x-bearer-format'?: string;
-  };
```
erunion, line 26: We don't need to cast this anymore to support `x-bearer-format` because `@scalar/openapi-types` has support for `additionalProperties`on `apiKey` schemes.

packages/oas-to-har/src/lib/configure-security.ts
```diff
@@ -63,26 +61,26 @@ export default function configureSecurity(
   if (security.type === 'apiKey') {
     if (security.in === 'query') {
       return harValue('queryString', {
-        name: security.name,
+        name: security.name!,
```
erunion, line 64: More quirks with this library where things that are required by the spec are typed as optional.

packages/oas-to-har/src/lib/utils.ts
```diff
@@ -505,7 +509,13 @@ export function getSafeRequestBody(schema: SchemaObject, payload: unknown, api:
     resolved = pickPolymorphicBranch(alternatives, keys, api);
   }
 
-  if ('allOf' in resolved && Array.isArray(resolved.allOf) && resolved.allOf.length) {
+  if (
+    typeof resolved === 'object' &&
+    resolved !== null &&
```
erunion, line 514: Had to add this because `resolved` isn't guaranteed to be an object by way of it being a better typed `SchemaObject` now -- it could be a boolean or `null`.

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -764,8 +764,12 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
           let currentExample = example as ExampleObject | ReferenceObject;
           if (name === '$ref') {
             currentExample = dereferenceRef({ $ref: currentExample } as ReferenceObject, definition, seenRefs);
-            if (!currentExample || isRef(currentExample)) {
-              refLogger(currentExample.$ref, 'ref');
+            if (!currentExample) {
+              return;
```
erunion, line 768: This fixes a crash where we could have been accessing `$ref` on `undefined`.

packages/oas/src/lib/openapi-to-json-schema.ts
```diff
@@ -830,6 +834,12 @@ export function toJSONSchema(data: SchemaObject | boolean, opts?: toJSONSchemaOp
     return schema;
   }
 
+  // `SchemaObject` is a union that includes plain JSON Schema (which permits fully-optional,
+  // boolean-schema shapes). TS can't cleanly exclude `ReferenceObject` from that union via control
+  // flow analysis alone, so we reassert the type here now that the `isRef` branch above has
+  // returned.
+  schema = schema as SchemaObject;
```
erunion, line 841: Annoying but the types on `SchemaObject` are **nuts**.

packages/oas/src/operation/lib/dedupe-common-parameters.ts
```diff
@@ -17,8 +15,6 @@ export function dedupeCommonParameters(
     return !parameters.find((param2: ParameterObject) => {
       if (param.name && param2.name) {
         return param.name === param2.name && param.in === param2.in;
-      } else if (isRef(param) && isRef(param2)) {
-        return param.$ref === param2.$ref;
```
erunion, line 21: This was removed because this method only receives arrays of `ParameterObject` objects, never any `ReferenceObject`s.

packages/oas/src/types.ts
```diff
@@ -99,6 +112,7 @@ export type HttpMethods =
   | 'patch'
   | 'post'
   | 'put'
+  | 'query'
```
erunion, line 115: OpenAPI 3.2 added support for the `QUERY` HTTP method.

packages/oas/src/types.ts
```diff
@@ -177,50 +195,53 @@ export type OperationObject = (OpenAPIV3_1.OperationObject | OpenAPIV3.Operation
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameter-object}
  */
 export type ParameterObject = {
-  in: 'cookie' | 'header' | 'path' | 'query';
-} & (OpenAPIV3_1.ParameterObject | OpenAPIV3.ParameterObject);
+  in: 'cookie' | 'header' | 'path' | 'query' | 'querystring';
```
erunion, line 198: OpenAPI 3.2 added support for a new `in: querystring` parameter type. We don't support this yet.

PR #1173, opened 2026-07-08
fix(oas): validate x-readme.status-url as a string

maximilianfalco:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

`x-readme.status-url` was getting rejected on spec upload with `"x-readme.status-url" must be of type "Boolean"`, even though its a URL string.

`status-url` is the customer-declared health url we surface as the RFC 9727 `status` link relation in a project's `/.well-known/api-catalog`. it got added to `extensionDefaults` but `validateExtension` never got a branch for it, so `validateExtensions()` looped over it and fell through to the catch-all boolean check. this adds a `STATUS_URL` branch (both nested under `x-readme` and at the root `x-status-url`) that validates its a string, so valid urls pass and anything non-string gets a `must be of type "String"` error instead.

## 🧬 QA & Testing

covered by a new row in the `#validateExtension` test table in `packages/oas/test/extensions.test.ts`, a valid url string passes and a boolean now fails with `must be of type "String"`.

verified end to end by symlinking this build into the readme monorepo and uploading a spec with `"x-readme": { "status-url": "https://status.example.com" }`, which used to throw the boolean error and now imports fine.

### DIFF
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -23,6 +23,7 @@ import {
   OAUTH_OPTIONS,
   PARAMETER_ORDERING,
   SAMPLES_LANGUAGES,
+  STATUS_URL,
   validateParameterOrdering,
 } from './extensions.js';
 import { getAuth } from './lib/get-auth.js';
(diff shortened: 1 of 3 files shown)


erunion (approved):

packages/oas/src/index.ts
```diff
@@ -689,6 +690,10 @@ export default class Oas {
           if (typeof data[extension] !== 'object') {
             throw new TypeError(`"x-readme.${extension}" must be of type "Object"`);
           }
+        } else if (extension === STATUS_URL) {
```
maximilianfalco, line 693: went with a plain `typeof === string` check here to stay consistent with the other branches (boolean/object all just do a typeof), rather than validating its a real url via `new URL()`. the upload form already does url-format validation on the frontend so this is really just the type gate, we can tighten it to reject non-url strings later if we ever want to.

PR #1174, opened 2026-07-11
fix(oas-normalize): resolve $ref paths relative to spec file

### LINKED ISSUE #921: BUG: $ref relative paths are relative to CWD, should be relative to current document
```
.
|- tools
|    |- bundler.js // contains use of oasNormalize
|- api
     |- my-api.yaml
     |- paths
          |- path.yaml
```

If `my-api.yaml` contains a `$ref: "./paths/path.yaml"` it will fail with
```
node ./tools/bundler.js ./api/myapi.yaml
```

But will succeed with
```
cd ./api
node ../tools/bundler.js ./myapi.yaml
```

A `$ref` should be evaluated relative to the currently open file.

brain-marchine:
Fixes #921

When `OASNormalize` is initialized with a filesystem path (`enablePaths: true`), `bundle()`, `dereference()`, and `validate()` now pass that path through to `@readme/openapi-parser` instead of the in-memory schema object.

This lets relative `$ref` pointers resolve against the spec file location rather than the process working directory.

## Test plan

- [x] Added regression test that bundles `./pet.json` refs while `process.cwd()` is outside the fixture directory
- [ ] CI (vitest) on Linux

### DIFF
--- a/packages/oas-normalize/src/index.ts
+++ b/packages/oas-normalize/src/index.ts
@@ -3,6 +3,7 @@ import type { ParserOptions, ValidationResult } from '@readme/openapi-parser';
 import type { OpenAPI, OpenAPIV2, OpenAPIV3 } from 'openapi-types';
 
 import fs from 'node:fs';
+import path from 'node:path';
 
 import { bundle, compileErrors, dereference, validate } from '@readme/openapi-parser';
 import { isUnsafeURL } from '@readme/openapi-parser/lib/urls';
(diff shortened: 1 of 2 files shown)


erunion (changes requested):
There's something up with the diff here, did it rewrite every indent to a tab? Can you run `npm run lint` to clean this up?

Also can you add a changeset for this fix? You can do that through `npm run changeset` and give it a short description of what you've done -- without that our publishing flow won't generate a new release of `oas-normalize`.

PR #1176, opened 2026-07-14
feat(oas): match URLs served by path-item and operation-level servers

HugoHSun:
| 🚥 Resolves RM-1745 |
| :------------------- |

## 🧰 Changes

`findOperation()` and friends only matched URLs against root-level `servers`, so operations that override them at the path-item or operation level could never resolve. #1159 taught the display side about those servers, this teaches lookups.

- Operations are grouped by the server list that governs them (operation > path-item > root, empty arrays fall through), then matched against those servers in declared order using the same two strategies as root matching.
- When a URL resolves to both a root match and an override match at equal specificity, the override wins.
- Root matching behavior is unchanged. The split strategies it shares with the new code are now common helpers.
- Paths without overrides are skipped via a cheap guard, so root-only definitions barely pay for this.

## 🧬 QA & Testing

New tests cover server precedence (including two operations on one path governed by different servers), relative server URLs, variables with non-default values, path-item `$ref`s, empty server arrays, and the `findOperationWithoutMethod()`/`getOperation()` wrappers. The rest of the suite passes untouched, which pins root-only behavior.

### DIFF
--- a/packages/oas/src/index.ts
+++ b/packages/oas/src/index.ts
@@ -5,6 +5,7 @@ import type {
   HttpMethods,
   OASDocument,
   OperationObject,
+  PathsObject,
   SecuritySchemeObject,
   ServerObject,
   Servers,
@@ -43,6 +44,14 @@ import {
 import { Operation, Webhook } from './operation/index.js';
 import { isOpenAPI31, isRef } from './types.js';
 import { SERVER_VARIABLE_REGEX, supportedMethods } from './utils.js';
+
+/**
+ * A URL decomposed into the server URL that serves it and the path that's left over.
+ */
+interface ServerURLSplit {
+  origin: string;
+  pathName: string;
+}
 
 export default class Oas {
   /**
@@ -256,12 +265,26 @@ export default class Oas {
   }
 
   findOperationMatches(url: string): PathMatches | undefined {
+    // Root server matches are collected first so that when `findTargetPath()` breaks a tie
+    // between equally-specific matches, the last one -- the server defined closest to the
+    // operation -- wins.
+    const annotatedPaths = [
+      ...(this.findRootServerOperationMatches(url) || []),
+      ...this.findOverriddenServerOperationMatches(url),
+    ];
+    if (!annotatedPaths.length) return undefined;
+
+    return annotatedPaths;
+  }
+
+  /**
+   * Find matches for a URL against paths as they are served by the root-level `servers`.
+   */
+  private findRootServerOperationMatches(url: string): PathMatches | undefined {
     const { origin, hostname } = new URL(url);
     const originRegExp = new RegExp(origin, 'i');
     const { servers, paths } = this.api;
 
-    let pathName: string | undefined;
-    let targetServer: ServerObject | undefined;
     let matchedServer: ServerObject | undefined;
 
     if (!servers || !servers.length) {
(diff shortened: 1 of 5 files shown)


packages/oas/src/index.ts
```diff
@@ -280,64 +303,151 @@ export default class Oas {
...
-      // Instead of setting `url` directly against `matchedServer` we need to set it to an
-      // intermediary object as directly modifying `matchedServer.url` will in turn update
-      // `this.servers[idx].url` which we absolutely do not want to happen.
-      targetServer = {
-        ...matchedServer,
-        url: this.replaceUrl(matchedServer.url, matchedServer.variables || {}),
-      };
-
-      [, pathName] = url.split(new RegExp(targetServer.url, 'i'));
+    const substituted = matchedServer ? this.splitURLOnSubstitutedServer(url, matchedServer) : undefined;
```
HugoHSun, line 306: The `targetServer = { ...matchedServer, url: ... }` intermediary is gone: nothing writes to server objects anymore, the substituted URL flows through `ServerURLSplit.origin` instead. Reported origins are unchanged.

packages/oas/src/index.ts
```diff
@@ -280,64 +303,151 @@ export default class Oas {
...
+    if (!paths) return matches;
+
+    Object.keys(paths).forEach(path => {
+      const rawPathItem = paths[path];
+      if (!rawPathItem) return;
+
+      // Most paths don't override the root servers, so before doing any allocation or
+      // dereferencing work make sure this one might: it declares path-item or operation servers,
+      // or hides them behind a `$ref` that we'd have to resolve to know.
+      let mayHaveOverrides = isRef(rawPathItem) || !!rawPathItem.servers?.length;
```
HugoHSun, line 348: Guard numbers, benchmarked on root-only specs (worst case, the scan finds nothing):

| spec | oas 37 | no guard | with guard |
| --- | --- | --- | --- |
| petstore (14 paths) | ~14 µs/lookup | +26% | +15% |
| synthetic 500 paths | ~568 µs/lookup | +28% | +15% |

Most of the baseline is per-lookup path-to-regexp compilation in `generatePathMatches`, which predates this change. Matcher caching follow-up coming separately.

packages/oas/src/index.ts
```diff
@@ -280,64 +303,151 @@ export default class Oas {
...
+        let group = groups.get(key);
+        if (!group) {
+          group = { operations: {}, servers };
+          groups.set(key, group);
+        }
+        group.operations[method] = operation;
+      });
+
+      groups.forEach(({ operations, servers }) => {
+        for (const server of servers) {
```
HugoHSun, line 390: Ordering note: this tries exact then variable match on one server before moving to the next (the `servers` array is a preference order), while root matching tries exact across all servers, then regex across all servers.

packages/oas/src/index.ts
```diff
@@ -280,64 +303,151 @@ export default class Oas {
...
+          const target = this.splitURLIntoOriginAndPathName(url, server);
+          if (!target) continue;
+
+          const annotatedPaths = generatePathMatches(
+            { [path]: operations } as PathsObject,
+            target.pathName || '/',
+            target.origin,
+          );
+
+          if (annotatedPaths.length) {
```
HugoHSun, line 400: First server that splits *and* template-matches wins, not just splits. Matters when one server is a prefix of another (`https://x.com` vs `https://x.com/base`).

packages/oas/src/index.ts
```diff
@@ -280,64 +303,151 @@ export default class Oas {
...
+   */
+  private splitURLOnSubstitutedServer(url: string, server: ServerObject): ServerURLSplit | undefined {
+    try {
+      const substitutedUrl = this.replaceUrl(server.url, server.variables || {});
+      const [, pathName] = url.split(new RegExp(substitutedUrl, 'i'));
+      if (pathName !== undefined) {
+        return { origin: substitutedUrl, pathName };
+      }
+    } catch {
+      // A server URL that can't be expressed as regex shouldn't fail the whole lookup.
```
HugoHSun, line 430: Behavior delta: root's exact split had no try/catch, so a server URL that forms invalid regex used to throw out of `findOperationMatches()`. Now it falls through to the regex rung instead.

HugoHSun:
@coderabbitai review

packages/oas/test/index.test.ts
```diff
@@ -1205,6 +1207,156 @@ describe('Oas', () => {
...
+          origin: 'https://demo.example.com:443/v2',
+          path: '/global',
+          nonNormalizedPath: '/global',
+          slugs: {},
+          method: 'POST',
+        });
+      });
+
+      it('should support relative path-item and operation-level servers', () => {
+        expect(pathLevelServers.findOperation('https://example.com/v2/relative-path-server', 'get')?.url).toMatchObject(
```
erunion: Can you change these `.toMatchObject()` assertions to be `.toStrictEqual()`? It's not clear from these assertions, or the fixtures, that these are being matched to the right HTTP method we're expecting (though I assume that these fixtures only have one method per endpoint).

packages/oas/src/index.ts
```diff
@@ -280,64 +303,151 @@ export default class Oas {
...
-      targetServer = {
-        ...matchedServerAndPath[0].matchedServer,
-      };
+  /**
+   * Try to split a URL on a server URL with its variables filled in with their defaults.
+   */
+  private splitURLOnSubstitutedServer(url: string, server: ServerObject): ServerURLSplit | undefined {
+    try {
+      const substitutedUrl = this.replaceUrl(server.url, server.variables || {});
+      const [, pathName] = url.split(new RegExp(substitutedUrl, 'i'));
```
erunion, line 425: How com this isn't calling `transformURLIntoRegex()`?

erunion (changes requested):

packages/oas/src/index.ts
HugoHSun: Here the variables are already interpolated to server default/user values; variable to regex happens later if this one doesn't work

erunion (approved):

darrenyong (approved):

PR #1177, opened 2026-07-14
perf(oas): cache compiled path matchers across lookups

HugoHSun:
| 📈 Follow-up to #1176 |
| :------------------- |

## 🧰 Changes

Every URL lookup (`findOperation()` and friends) compiled a `path-to-regexp` matcher for every path template in the definition, executed it once and threw it away. Compiling costs an order of magnitude more than executing, so recompilation was ~90% of a lookup.

A compiled matcher depends only on the template string, so they're now cached process-wide, same pattern as the analyzer's query cache from #1170. String keys can't live in a `WeakMap`, so the cache is size-bounded instead.

Warm lookups against `main`:

| spec | before | after |
| :--- | :--- | :--- |
| petstore (14 paths) | ~14 µs | ~3 µs |
| synthetic 500 paths | ~560 µs | ~60 µs |

## 🧬 QA & Testing

- Templates that fail to compile are cached as never-matching, the same outcome as the old per-lookup try/catch. Execution errors (like a bad percent-encoding hitting `decodeURIComponent`) depend on the incoming URL, not the template, so they stay per-lookup.
- Full suite passes untouched.

### DIFF
--- a/packages/oas/src/lib/urls.ts
+++ b/packages/oas/src/lib/urls.ts
@@ -216,6 +216,41 @@ function normalizePath(path: string) {
   );
 }
 
+// Bounds the matcher cache in long-lived multi-tenant processes.
+const MATCHER_CACHE_MAX_SIZE = 10_000;
+const matcherCache = new Map<string, { cleanedPath: string; matcher: ((path: string) => Match<ParamData>) | null }>();
+
+/**
+ * Compile a path template into its normalized form and `path-to-regexp` matcher. Compiling costs
+ * an order of magnitude more than executing and depends only on the template string, so compiled
+ * matchers are cached process-wide. A `null` matcher marks a template that can't be compiled.
+ */
+function compilePathMatcher(path: string): {
+  cleanedPath: string;
+  matcher: ((path: string) => Match<ParamData>) | null;
+} {
+  let compiled = matcherCache.get(path);
+  if (compiled === undefined) {
+    const cleanedPath = normalizePath(path);
+    let matcher = null;
+    try {
+      matcher = match(cleanedPath, { decode: decodeURIComponent });
+    } catch {
+      // A template that can't be compiled (maybe it has a malformed path parameter) can never
+      // match; record that instead of failing.
+    }
+
+    if (matcherCache.size >= MATCHER_CACHE_MAX_SIZE) {
+      matcherCache.delete(matcherCache.keys().next().value as string);
+    }
+
+    compiled = { cleanedPath, matcher };
+    matcherCache.set(path, compiled);
+  }
+
+  return compiled;
+}
+
 /**
  * Generate path matches for a given path and origin on a set of OpenAPI path objects.
  *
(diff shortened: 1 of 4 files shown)


packages/oas/src/lib/urls.ts
```diff
@@ -216,6 +216,41 @@ function normalizePath(path: string) {
   );
 }
 
+// Bounds the matcher cache in long-lived multi-tenant processes.
+const MATCHER_CACHE_MAX_SIZE = 10_000;
```
HugoHSun, line 220: 10k is a guardrail, roughly 20x a large definition, sized so eviction basically never happens. FIFO instead of LRU because LRU pays two Map ops on every cache hit (times every template, times every lookup) to improve eviction quality in a regime the cap is meant to avoid. If churn past the cap ever shows up in practice, the fix is a bigger cap or clock eviction.

erunion (changes requested):
Can you add a benchmark for this to `test/benchmarks`?

HugoHSun:
> Can you add a benchmark for this to `test/benchmarks`?

Added a benchmark that builds a resolvable URL for every operation in petstore and docusign and looks them all up with `findOperation()` / `findOperationWithoutMethod()`. Same file run on both commits:

| workload | `main` | this branch |
| --- | --- | --- |
| petstore (20 lookups/batch) | 0.251 ms/batch | 0.041 ms/batch (**6.1x**) |
| docusign (404 lookups/batch) | 174.6 ms/batch | 8.19 ms/batch (**21.3x**) |
| docusign, `findOperationWithoutMethod()` | 173.5 ms/batch | 7.95 ms/batch (**21.8x**) |

`npx vitest bench url-matching`

packages/oas/src/lib/urls.ts
```diff
@@ -216,6 +216,41 @@ function normalizePath(path: string) {
...
+const matcherCache = new Map<string, { cleanedPath: string; matcher: ((path: string) => Match<ParamData>) | null }>();
```
erunion, line 221: Should this be a `WeakMap` instead?

packages/oas/src/lib/urls.ts
```diff
@@ -216,6 +216,41 @@ function normalizePath(path: string) {
...
+    try {
+      matcher = match(cleanedPath, { decode: decodeURIComponent });
+    } catch {
+      // A template that can't be compiled (maybe it has a malformed path parameter) can never
+      // match; record that instead of failing.
+    }
+
+    if (matcherCache.size >= MATCHER_CACHE_MAX_SIZE) {
+      matcherCache.delete(matcherCache.keys().next().value as string);
+    }
```
darrenyong, line 245: don't know if we will ever hit the 10k limit, but if we did, this would technically delete entries that are used often, right? whereas entries that not be hit that often would persist?

darrenyong (approved):

packages/oas/src/lib/urls.ts
HugoHSun: `WeakMap` keys would have to be objects, but this is keyed by the path-template string so it can be shared across specs and across the short-lived `Oas` instances metrics ingestion creates. Hence we have a 10k cap here instead of relying on GC

packages/oas/src/lib/urls.ts
HugoHSun: Yep that's correct. FIFO evicts by insertion order, not by use, so in principle a hot template can be dropped while a cold one survives. Though I think the tradeoffs are acceptable here

erunion (approved):

PR #1181, opened 2026-07-27
fix(oas): normalize server URLs after variable substitution

HugoHSun:
Fixes [CX-3748](https://linear.app/readme-io/issue/CX-3748/api-explorer-automatically-prepends-a-protocol-to-the-base-url-causing).

## Changes

- Normalize server URLs after substituting their selected variable values.
- Keep `splitUrl()` templates unresolved while using variable defaults to determine whether they supply the protocol.
- Cover root and operation server variables with and without protocols.

## QA & Testing

- A `{server}/v1` template with an `https://api.example.com` value resolves without duplicating the protocol.
- A variable value without a protocol defaults to `https://`.
- Existing protocol-relative, relative-path, and host-only server URLs keep their current normalization behavior.

### DIFF
--- a/packages/oas/src/lib/urls.ts
+++ b/packages/oas/src/lib/urls.ts
@@ -7,6 +7,8 @@ import { SERVER_VARIABLE_REGEX } from '../utils';
 
 import getUserVariable from './get-user-variable';
 
+const PROTOCOL_REGEX = /^[a-z][a-z\d+.-]*:\/\//i;
+
 export interface PathMatch {
   match?: Match<ParamData>;
   operation: PathsObject;
@@ -76,6 +78,7 @@ export function normalizeURL(url: string | undefined, urlForProtocolCheck: strin
   if (!url) return exampleDotCom;
 
   url = stripTrailingSlash(url);
+  if (!url) return exampleDotCom;
 
   // Check if the URL is just a path and missing an origin, for example `/api/v3`. If so, then make
   // `example.com` the origin to avoid it becoming something invalid like `https:///api/v3`.
@@ -93,7 +96,7 @@ export function normalizeURL(url: string | undefined, urlForProtocolCheck: strin
     return `https:${url}`;
   }
 
-  if (!urlForProtocolCheck?.includes('//')) {
+  if (!PROTOCOL_REGEX.test(urlForProtocolCheck?.trim() ?? '')) {
     return `https://${url}`;
   }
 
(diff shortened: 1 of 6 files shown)


packages/oas/src/lib/urls.ts
```diff
@@ -62,54 +62,58 @@ export function stripTrailingSlash(url: string): string {
...
-      url = urlWithOrigin.href;
-    }
-  } catch {
-    url = exampleDotCom;
-  }
+function normalizedURLFromServers(servers: ServerObject[] | undefined, selected: number): string {
+  const server = servers?.[selected];
+  const resolvedUrl = server?.url.replace(SERVER_VARIABLE_REGEX, (original, key) => {
+    return key in (server.variables || {}) ? String(server.variables?.[key].default) : original;
+  });
```
darrenyong: will there ever be a case where `url` doesn't exist? would this throw an error here if `server.url` is `undefined` or something?

erunion: If we have a `ServerObject` object here `url` should be always be present.

https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-3

darrenyong (approved):

erunion (changes requested):
please address or resolve the coderabbit issues

packages/oas/src/lib/urls.ts
HugoHSun: updated to `server?.url?.` since we are quite lenient here

packages/oas/src/lib/urls.ts
```diff
@@ -62,54 +62,58 @@ export function stripTrailingSlash(url: string): string {
...
+  if (urlForProtocolCheck?.startsWith('//')) {
     return `https:${url}`;
   }
 
-  // Add protocol to urls with no // within them
-  // This is because httpsnippet throws a HARError when it doesnt have a protocol
-  if (!url.match(/\/\//)) {
+  if (!urlForProtocolCheck?.includes('//')) {
     return `https://${url}`;
   }
```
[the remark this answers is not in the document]
HugoHSun: Have updated it to use `const PROTOCOL_REGEX = /^[a-z][a-z\d+.-]*:\/\//i;`

HugoHSun:
> please address or resolve the coderabbit issues

Address them except the query parameter encoding since it shouldn't be used here https://spec.openapis.org/oas/v3.1.2.html#server-object

erunion (approved):

PR #1183, opened 2026-07-30
fix(parser): upgrade json-schema-ref-parser to v15

erunion:
## Summary

- Upgrades `@apidevtools/json-schema-ref-parser` from `^14.1.1` to `^15.5.1` in `packages/parser`.
- v15 is ESM-only, so it's added to `noExternal` in `packages/parser/tsup.config.ts` so it still gets bundled into our CJS build.
- Fixes two upstream `isUnsafeUrl()` regressions in `packages/parser/src/lib/urls.ts`:
  - Relative paths (`./foo.json`, `../foo.json`, or any path with a lone `.`/`..` segment) were being misclassified as unsafe.
  - Bare (unbracketed) private IPv6 literals (e.g. `http://fe80::1`) were being misclassified as safe.
- Updates fixtures in `packages/parser` and `packages/oas` for two intentional (and, in the `oas` circular-ref analyzer's case, more *correct*) v15 behavior changes:
  - Bundling picks a different (but structurally equivalent) `$ref` pointer for self-referencing circular refs.
  - `onCircular` now fires for every occurrence of a circular ref instead of missing some, so `oas`'s `analyzeCircularRefs()` now reports circular `$ref`s more completely (e.g. the DocuSign test fixture goes from 4 reported locations to the true 96).

## Test plan

- [x] `npm run build` succeeds across the monorepo
- [x] `npm run lint:types` passes across the monorepo
- [x] Full `vitest` suite passes (2152 tests)
- [x] Verified the CJS (`require()`) and ESM (`import`) builds of `@readme/openapi-parser` both load and run correctly
- [x] `attw --pack` shows clean CJS/ESM resolution for `@readme/openapi-parser`
- [x] Added regression tests for the `isUnsafeUrl()` fixes (relative paths, bare vs. bracketed private/public IPv6)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/parser/test/specs/circular/validated.ts
```diff
@@ -143,7 +143,7 @@ const validatedAPI = {
             200: {
               description: 'Returns a thing',
               schema: {
-                $ref: '#/definitions/thing',
+                $ref: '#/paths/~1thing/get/responses/200/schema',
```
erunion, line 146: #### ⚠️ Why this changed

The circular deref fixture failed because **v15 rewrites direct self-`$ref`s to a different JSON Pointer**.

In this spec, `definitions.thing` is a direct self-reference:

```yaml
thing:
  $ref: 'circular.yaml#/definitions/thing' # <--- circular reference to self
```

…and `/thing` points at that definition. With circular dereferencing on, the library can’t fully expand it, so it leaves a `$ref` and rewrites it to `pathFromRoot` — the path of the `$ref` being processed when the cycle was detected:

```js
if (directCircular && dereferencedValue !== $ref) {
  // The pointer is a DIRECT circular reference (i.e. it references itself).
  // So replace the $ref path with the absolute path from the JSON Schema root
  dereferencedValue.$ref = pathFromRoot;
}
```

Traversal hits `paths` before `definitions`, so the first encounter is the response schema. v15 therefore emits:

```js
{ $ref: '#/paths/~1thing/get/responses/200/schema' }
```

instead of the old:

```js
{ $ref: '#/definitions/thing' }
```

That’s why `dereferenced.ts` / `validated.ts` had to change in the upgrade commit. Semantically it’s the same cycle (and `paths['/thing']…schema === definitions.thing` still holds); only the preserved pointer string changed.

packages/parser/test/specs/circular-slowdowns/circular-slowdowns.test.ts
```diff
@@ -39,7 +39,7 @@ describe('API with extensive circular $refs that cause slowdowns', () => {
     });
 
     // Ensure that a circular $ref **was** dereferenced.
-    expect(circularRefs).toHaveLength(23);
+    expect(circularRefs).toHaveLength(118);
```
erunion, line 42: These counts blew up because the `onCircular` event in `@apidevtools/json-schema-ref-parser` fires for **every** circular occurrence now, not once per unique target -- meaning that we can better detect when a schema is circular now.

darrenyong (approved):

PR #1191, opened 2026-08-10
fix: honor a schema's own `description` alongside an `allOf` `$ref`

azyl20:
| 🚥 Resolves CX-3769 |
| :------------------- |

## 🧰 Changes

When a schema attaches its own `description` as a sibling of an `allOf`, that
description now takes precedence over any `description` merged in from a branch
(e.g. a `$ref` target).

## Why

OpenAPI 3.0 forbids a `$ref` from having sibling properties, so to give a shared
schema a context-specific description authors wrap the `$ref` in an `allOf` and
hang the new `description` off it:

  ```json
  {
    "Pet": {
      "allOf": [{ "$ref": "#/components/schemas/Error" }],
      "description": "A pet"
    }
  }
```

- This allOf + description pattern is a de facto standard (Swagger and others resolve it this way)

## 🧬 QA & Testing

### DIFF
--- a/packages/oas/test/lib/__snapshots__/openapi-to-json-schema.test.ts.snap
+++ b/packages/oas/test/lib/__snapshots__/openapi-to-json-schema.test.ts.snap
@@ -1037,7 +1037,7 @@ exports[`toJSONSchema() > \`description\` support > should support description i
 exports[`toJSONSchema() > polymorphism / inheritance > quirks > should perform default resolution for unrecognized properties 1`] = `
 {
   "additionalProperties": false,
-  "description": "Represents a namespace (one).",
+  "description": "Represents a namespace (two).",
   "properties": {
     "dataAccessPolicy": {
       "additionalProperties": false,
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/oas/test/lib/__snapshots__/openapi-to-json-schema.test.ts.snap [hunk 1]
```diff
@@ -1037,7 +1037,7 @@ exports[`toJSONSchema() > \`description\` support > should support description i
 exports[`toJSONSchema() > polymorphism / inheritance > quirks > should perform default resolution for unrecognized properties 1`] = `
 {
   "additionalProperties": false,
-  "description": "Represents a namespace (one).",
+  "description": "Represents a namespace (two).",
```
azyl20, line 1040: snapshot change comes directly from the "sibling always wins" fix in this PR

packages/oas/test/operation/transformers/__snapshots__/get-parameters-as-json-schema.test.ts.snap
```diff
@@ -180,7 +180,7 @@ exports[`.getParametersAsJSONSchema() > $ref quirks > should retain component sc
       },
       "properties": {
         "filterBinder": {
-          "description": "
+          "description": "Tells the query engine of the Filters should be applied as an And or an Or (the default is And)
```
azyl20, line 183: snapshot change comes directly from the "sibling always wins" fix in this PR

erunion (approved):

PR #1193, opened 2026-08-19
fix(parser): drop orphaned `$id` keywords when bundling, validating, dereferencing

azyl20:
| 🚥 Resolves CX-3821 |
| :------------------- |

## Summary

`@readme/openapi-parser` now discards **orphaned** `$id` keywords before resolving `$ref`s. This
fixes `Missing $ref pointer "#/…". Token "…" does not exist.` errors that broke uploads of
previously-valid API definitions.

The strip runs in `bundle()`, `dereference()`, and `validate()` — each resolves `$ref`s
independently, so all three hit the same failure and all three need the fix.

## Root cause

Upgrading `json-schema-ref-parser` 14 → 15 (via #20149's SSRF-hardening bump) started honoring
JSON Schema 2020-12's `$id` as a **base URI** for `$ref` resolution. That surfaced a latent issue in
bundling output:

When a definition references the same external schema more than once, the bundler inlines the first
occurrence and rewrites the rest into internal `#/…` pointers — but it also carries the external
schema's `$id` onto the inlined copy. That leftover `$id` re-scopes the sibling `#/…` pointers so
they resolve against the (non-existent) `$id` document instead of the definition root:

```
// what the bundler emits — and then chokes on when re-processed under 15.x
"Pets": { 
  "$id": "requests/search.schema.json",              // ← carried in from the inlined file
  "type": "array",
  "items": { "$ref": "#/paths/~1petsl/…/schema" }    // ← now mis-scoped by the $id above
}
```

The $id is orphaned (nothing references it) and does no work — it only mis-scopes the pointer.
Older resolvers ignored $id entirely, which is why this never surfaced before.

## The fix   

stripOrphanedIds() removes a nested $id only when nothing depends on it:
- it isn't a $ref target,
- it isn't a base URI for a relative $ref (or nested relative $id) beneath it,
- it isn't the document's own root $id.

These are deliberately conservative string-level checks, not full RFC 3986 base-URI resolution.
They always err toward keeping a $id, so the repair never breaks a working definition. In
particular, an $id used as an in-document reference anchor is preserved (verified it still resolves
internally, no external fetch / SSRF). The only cost is that a rare construct a string compare can't
disambiguate (e.g. the same relative id reused across scopes) is left as-is — unimproved, never broken.

Alternatives considered

- Roll back to a pre-$id json-schema-ref-parser
   - not possible: the versions that ignore $id (≤15.3.1) predate the DNS-based SSRF hardening #20149 required.
- A resolver option to ignore $id
   - none exists; $id scoping is hardcoded at parse time.
- Strip all $ids
   - rejected: it breaks in-document $id anchors, converting internal refs into external fetches (wrong content + reintroduces the SSRF surface #20149 closed).

This is also a known upstream issue (APIDevTools/json-schema-ref-parser#355); the maintainer's
recommended workaround is likewise "remove the $id."

## 🧬 QA & Testing

- New regression tests under .bundle() → orphaned $id keywords: two reproductions of the failing artifact + a guard that a referenced $id anchor survives.


packages/oas-normalize/src/index.ts
```diff
@@ -185,7 +186,10 @@ export default class OASNormalize {
 
         return schema;
       })
-      .then(schema => bundle(schema, parserOptions))
+      .then(schema => {
+        stripOrphanedIds(schema);
```
erunion: Why do this work here instead of in the `bundle` method of the parser?

azyl20: @erunion Thanks for pushing on this. You're right it should be there. I've moved everything to the parser, and also extended it to cover `bundle`, `dereference`, and `validate`.

packages/parser/src/repair.ts
```diff
@@ -3,6 +3,77 @@ import type { OpenAPI, OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
...
+ */
+export function stripOrphanedIds(schema: unknown): void {
+  const refs = new Set<string>();
+  collectRefs(schema, refs);
+
+  const isReferenced = (id: string): boolean => {
+    for (const ref of refs) {
+      if (ref.includes(id)) return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
erunion: please address this

erunion (changes requested):
i think this looks fine but it's frustrating that the library is even parsing $id properties like this because those aren't really refs, and scanning for these refs is going to add some additional latency for large files that we didn't have before

does json-schema-ref-parser have no way to disable this functionality? or are there any onRef hooks we can tap into to hack in a solution that doesn't require an additional full schema scan?

packages/parser/src/repair.ts
```diff
@@ -3,6 +3,116 @@ import type { OpenAPI, OpenAPIV3, OpenAPIV3_1 } from 'openapi-types';
...
+export function stripOrphanedIds(schema: unknown): void {
+  const refs = new Set<string>();
+  collectRefs(schema, refs);
+
+  const isReferenced = (id: string): boolean => {
+    for (const ref of refs) {
+      // A `$ref` targets this `$id` when the ref, stripped of any fragment, exactly equals the `$id`.
+      // Matching the whole target (rather than a substring) avoids a short `$id` value spuriously
+      // matching an unrelated ref (e.g. `$id: "pet"` against `#/components/schemas/carpet`).
+      if (ref.split('#')[0] === id) return true;
```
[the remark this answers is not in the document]
azyl20: we're never (and it shouldn't be our job) to maintain parity with RFC 3986. With our string approach, we're not gonna be able to cover all the edge cases that base-URI resolution would. The scope of this ticket should just be to unblock the customer than to cover every edge case.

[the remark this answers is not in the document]
azyl20: no don't remember it as a rule, just rereview the damn pr with it in mind

azyl20:
@erunion, yeah super frustrating. I've looked at different versions and there is no way to disable it, or provide custom logic to go around it. I don't think this code even belongs in our codebase, but what else can we do?

In fact there was an [identical issue opened](https://github.com/APIDevTools/json-schema-ref-parser/issues/355) by another user on that package. It was supposedly fixed but claude determined the fix to be incomplete as it only touched `bundle()` and keeps the $id. I had it verify that `15.3.2`'s own bundle output fails its own `dereference()`. Other versions `15.5.1` / `16.0.0` 

The approach I settled on is this PR. The idea is that it doesn't have 100% parity with RFC 3986 base resolution since we are just using string comparison so that it works for most cases like the customer's, but doesnt add too much complexity that IMO shouldn't be our responsibility in the first place.

let me know what you think. Thanks

.changeset/fine-pants-go.md
```diff
@@ -0,0 +1,5 @@
+---
+"@readme/openapi-parser": patch
```
erunion: Can you bump this up to a `major` and also toss `oas-normalize` in here so it gets a major bump too? The last time I pushed a patch change for something innoculous like this to `oas-normalize` it ended up breaking existing `rdme` installs.

erunion (changes requested):

.changeset/fine-pants-go.md
azyl20: done

erunion (approved):

PR #1194, opened 2026-08-24
feat(oas): add OpenAPI pruner

HugoHSun:
| 🚥 Resolves [CX-3766](https://linear.app/readme-io/issue/CX-3766/hosted-downloadable-oas-should-omit-x-internal-hidden-endpoints) | 🧱 Stacked on [#1195](https://github.com/readmeio/oas/pull/1195) |
| :--- | :--- |

## 🧰 Changes

Adds an `OpenAPIPruner` for removing selected content from an OpenAPI definition while cleaning up components and tags that are no longer reachable.

[readmeio/readme#20280](https://github.com/readmeio/readme/pull/20280) needs the inverse of `OpenAPIReducer` for public OpenAPI downloads: start with the complete definition, remove operations that are not publicly visible, and return a valid definition without their unused dependencies. Keeping this in `oas` makes the reference and component cleanup reusable outside that API route.

> [!NOTE]
> This PR is stacked on [#1195](https://github.com/readmeio/oas/pull/1195), which contains the behavior-preserving transformer extraction. This diff contains the new pruner API and the reference-handling fixes added alongside it.

> [!IMPORTANT]
> The shared transformer has two existing reference-handling limitations that predate this PR:
>
> - Referenced Path Items are preserved intact rather than resolved, so their operations cannot be transformed individually.
> - The complete dependency set of operations and Path Items retained transitively through cross-operation references is not expanded before paths and webhooks are transformed.

### Pruner behavior

| Aspect | Behavior |
| --- | --- |
| Selection | Adds `removeTag()`, `removePath()`, `removeOperation()`, and `removeWebhook()` for case-insensitive removal. Operations without tags are not removed by tag filters. |
| Composition | Pruner filters are additive. An operation is removed when it matches a tag or location filter, and selecting a whole path or webhook is not narrowed by a later operation selection. |
| Cleanup | Removes components and tags that are no longer reachable while leaving the input definition unchanged. |
| Empty results | Allows a valid empty Paths Object when pruning removes every operation. |

### Reference and reducer behavior

| Aspect | Behavior |
| --- | --- |
| Referenced Path Items | Keeps referenced paths and webhooks intact during operation-level filtering. A whole referenced path or webhook can still be removed when it is not used elsewhere. Local sibling dependencies are retained while the Path Item remains. |
| Cross-operation references | Retains operations required by surviving reducer selections. The pruner rejects a removal when it would break a reference from a surviving operation. |
| Common parameters | Retains parameter components used by operationless or surviving Path Items, and removes container metadata when filtering removes all of its operations. |
| Reducer selection | Preserves the existing intersection between tag and path, operation, or webhook filters, now made explicit in the documentation and tests. |

## 🧬 QA & Testing

| Area | Scenarios covered |
| --- | --- |
| Selection | Tag, path, operation, and webhook removals, including combined additive filters, case-insensitive matching, and untagged operations. |
| References | Referenced Path Items, local sibling dependencies, cross-operation references, and rejected removals that would leave dangling references. |
| Cleanup | Unreachable schemas, parameters, security schemes, tags, paths, and webhooks are removed while reachable dependencies remain. |
| Reducer compatibility | Tag and location intersections, referenced operations, common parameters, and filtered container cleanup. |
| Validity | OpenAPI 3.0 and 3.1 output remains valid, including operationless Path Items and an empty Paths Object. |


packages/oas/src/lib/openapi-transformer/transform.ts
```diff
@@ -0,0 +1,845 @@
...
+        // and any components it references.
+        this.$refs.add(webhook.$ref);
+        this.accumulateUsedRefs(definition, this.$refs, webhook.$ref);
+      }
+
+      // If this webhook has no operations then it will remain in our resulting API definition, so
+      // we need to retain any components referenced by its common parameters.
+      if (!isRef(webhook) && !this.hasOperations(webhook)) {
+        this.accumulateParameterRefs(webhook.parameters);
+      }
```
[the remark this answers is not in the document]
HugoHSun: This looks like a false positive. For a retained webhook with operations, common parameters are collected inside the surviving-operation loop:

```ts
this.accumulateParameterRefs(webhook.parameters);
```

`accumulateParameterRefs()` passes `#/paths/~1target/get/parameters/0` to `accumulateUsedRefs()`, which records `/target|get` in `retainPathMethods`. This happens before `transformPaths()`, so the target operation is retained.

I also reproduced this exact case locally. Reducing to a webhook whose common parameter references `#/paths/~1target/get/parameters/0` retained `GET /target`, removed an unrelated path, and produced a valid reference.

The earlier `!this.hasOperations(webhook)` branch only handles operationless webhooks. It isn't the only place webhook common parameters are collected.

packages/oas/src/lib/transformer/index.ts
```diff
@@ -587,35 +679,27 @@ export class OpenAPITransformer {
...
+          if (!retainedByRef) {
+            delete this.definition.paths?.[path]?.[method as HttpMethods];
+            removedOperation = true;
             return;
           }
         }
 
+        // This operation remains in the transformed definition, so retain any components
+        // referenced by its common Path Item parameters.
+        this.accumulateParameterRefs(pathItem?.parameters);
```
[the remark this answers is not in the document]
HugoHSun: Yep, this is valid. I traced it back to the legacy reducer behavior: operations retained through cross-operation references were already kept during the destructive reduction pass without first expanding their complete dependency set. The extracted transformer preserves that existing limitation rather than introducing it with the new pruner API.

A complete fix needs dependency discovery to reach a fixed point before paths or webhooks are mutated, and it overlaps with resolving referenced Path Items. I have added a FIXME and documented this as an existing limitation in the code, package docs, and PR description. I am keeping the broader transformer change as a separate follow-up.

packages/oas/src/lib/transformer/index.ts
```diff
@@ -587,35 +684,27 @@ export class OpenAPITransformer {
...
```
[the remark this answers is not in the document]
HugoHSun: Same underlying pre-existing limitation as the thread above. The complete dependency set of transitively retained operations and Path Items is not expanded before the destructive transform pass.

I have documented the limitation and added a FIXME. The fixed-point dependency pass and referenced Path Item resolution should be handled together as a separate follow-up rather than expanding the scope of this pruner PR.

packages/oas/README.md
```diff
@@ -334,6 +333,37 @@ console.log(OpenAPIReducer.init(petstore).byOperation('/pet', 'post').reduce());
 console.log(OpenAPIReducer.init(petstore).byPath('/pet').reduce());
 ```
 
+#### Pruner
+
+The `OpenAPIPruner` utility, located in `oas/pruner`, can be used to remove a specific set of tags, paths, operations, or webhooks from an OpenAPI definition. Components and tags that are no longer reachable from the remaining definition are also removed.
+
+Unlike `OpenAPIReducer`, which selects the content to retain, `OpenAPIPruner` is useful when you know which content to remove.
```
erunion:
```suggestion
Unlike `oas/reducer`, which selects the content to retain, the pruner is useful when you know which content to remove.
```

packages/oas/README.md
```diff
@@ -334,6 +333,37 @@ console.log(OpenAPIReducer.init(petstore).byOperation('/pet', 'post').reduce());
...
+console.log(OpenAPIPruner.init(petstore).removePath('/pet').prune());
+```
+
+Referenced Path Items in paths and webhooks are retained intact for operation-level removals because they cannot be partially transformed without rewriting their shared target. Removing an entire referenced path or webhook removes the Path Item and any components that are no longer reachable when it is not used elsewhere. The pruner rejects removals that would break a `$ref` from a surviving operation.
+
+> [!IMPORTANT]
+> The transformer currently has two reference-handling limitations:
+>
+> - Referenced Path Items are preserved intact rather than resolved, so their operations cannot be transformed individually.
+> - The complete dependency set of operations and Path Items retained transitively through cross-operation references is not expanded before paths and webhooks are transformed.
```
erunion: Probably okay to remove this since it's already noted in the paragraph above.

packages/oas/test/__datasets__/pruner-tag-filter-common-parameters.json
erunion, line 1: Can you add `pruner` into this filename somewhere?

packages/oas/test/lib/openapi-transformer/operation-selection.test.ts
```diff
@@ -26,12 +27,13 @@ describe('OperationSelection', () => {
     expect(selection.matches('/pets', 'delete')).toBe(false);
   });
 
-  it('can preserve an entire selection when adding individual operations', () => {
+  it('preserves an all-operation selection when adding individual operations', () => {
```
erunion, line 30: This test needs a falsy assertion to ensure it's not matching data it shouldn't.

packages/oas/test/reducer/index.test.ts
```diff
@@ -68,6 +70,36 @@ describe('OpenAPIReducer', () => {
       });
     });
 
+    it('should intersect tag and operation filters', async () => {
+      const reduced = OpenAPIReducer.init(petstore as OASDocument)
+        .byTag('Store')
+        .byOperation('/store/order', 'post')
```
erunion, line 97: Unrelated to this specific line but did you want to add a `.byOperationId` method to these to prune or reduce by that too?

erunion (changes requested):

packages/oas/test/lib/openapi-transformer/operation-selection.test.ts
HugoHSun: Added in the base PR https://github.com/readmeio/oas/pull/1195/changes#diff-4d57f20e4a64f0902a9fa88068ca691d4048aabffbe1445e161eb95a75998666R39

packages/oas/src/lib/transformer/index.ts
```diff
@@ -641,31 +722,35 @@ export class OpenAPITransformer {
...
-      // If this path no longer has any methods, delete it.
-      if (!Object.keys(this.definition.paths?.[path] || {}).length) {
+      // If filtering removed every operation from this path then remove its path-level properties
+      // as well. We leave path items that were authored without operations untouched.
+      if (
+        ((removedOperation || (this.mode === 'reduce' && excludePathItem)) && !this.hasOperations(pathItem)) ||
+        (this.mode === 'reduce' && !Object.keys(this.definition.paths?.[path] || {}).length)
+      ) {
         delete this.definition.paths?.[path];
       }
```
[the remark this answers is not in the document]
HugoHSun: This looks like a false positive. `pathItem` and `webhook` are live references to objects in `this.definition`, so deleting their operations updates what `hasOperations()` sees. When the last operation is removed, the entire container, including its common parameters, is deleted before unused components are cleaned up.

[This test](https://github.com/readmeio/oas/blob/1dcf220d5cd770a5e2d5a28df0a18e95f85bbff9/packages/oas/test/pruner/index.test.ts#L104-L115) covers the exact path and webhook scenario. It validates the output and asserts that both containers and their parameter components are removed. I reran it against the current head and it passes.

packages/oas/test/reducer/index.test.ts
HugoHSun: Yep added

packages/oas/test/reducer/index.test.ts
```diff
@@ -388,6 +436,43 @@ describe('OpenAPIReducer', () => {
...
   });
 
+  describe('.byOperationId()', () => {
+    it('should reduce by exact authored operation IDs and intersect with other filters', async () => {
+      const reduced = OpenAPIReducer.init(petstore as OASDocument)
+        .byTag('store')
+        .byPath('/store/order')
+        .byOperationId('placeOrder')
+        .byOperationId('getPetById')
+        .reduce();
```
erunion, line 446: Can you add a test that exclusively tests **only** `.byOperationId()`? From this test it's not clear if it's working because you're reducing by the past and tag that `placeOrder` is a part of.

.changeset/green-plums-prune.md
```diff
@@ -0,0 +1,5 @@
+---
+"oas": minor
+---
+
+Add an `OpenAPIPruner` utility for removing selected tags, paths, operations, operation IDs, and webhooks while pruning unreachable components and tags. Add operation ID selection to `OpenAPIReducer`, including generated IDs for operations without an authored `operationId`. The shared transformer also retains operations discovered through cross-operation references and removes metadata from containers when all their operations are filtered.
```
erunion, line 5:
```suggestion
Add an `oas/pruner` utility for removing selected tags, paths, operations, operation IDs, and webhooks while pruning unreachable components and tags. An `.byOperationId()` selector has also been added to `oas/reducer` to support reducing an OpenAPI definition by a given `operationId`(s).
```

erunion (approved):
Two very small requests but otherwise LGTM.

PR #1195, opened 2026-08-25
refactor(oas): extract OpenAPI transformer

HugoHSun:
| 🧱 Base for [#1194](https://github.com/readmeio/oas/pull/1194) | 🚥 Part of [CX-3766](https://linear.app/readme-io/issue/CX-3766/hosted-downloadable-oas-should-omit-x-internal-hidden-endpoints) |
| :--- | :--- |

## 🧰 Changes

Moves the existing `OpenAPIReducer` selection and transformation logic into internal `OperationSelection` and `OpenAPITransformer` classes. This gives [#1194](https://github.com/readmeio/oas/pull/1194) a shared foundation for adding the inverse pruning behavior without duplicating the reducer's reference and dependency handling.

> [!NOTE]
> This is a behavior-preserving refactor. The `OpenAPIReducer` public API and output semantics remain unchanged, so this PR does not include a changeset.

| Aspect | Refactor |
| --- | --- |
| Selection | Moves case-insensitive path, operation, and webhook selection into `OperationSelection`. |
| Transformation | Moves reference traversal, operation filtering, and unused component and tag cleanup into `OpenAPITransformer`. |
| Public API | Keeps `OpenAPIReducer.init()`, `byTag()`, `byPath()`, `byOperation()`, `byWebhook()`, and `reduce()` unchanged. |

## 🧬 QA & Testing

| Aspect | Verified behavior |
| --- | --- |
| Reducer compatibility | The existing reducer and selection suites pass without changing their expected output. |
| Types and packaging | The OAS package typecheck, build, and package export validation pass. |
| Code quality | The changed files pass Oxlint, Oxfmt, and diff validation. |


packages/oas/src/lib/openapi-transformer/operation-selection.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Select an operation within a path or webhook.
+   *
+   * When every operation for the key is already selected, `narrow` replaces that selection with
+   * the supplied operation while `union` preserves the existing all-operation selection.
+   *
+   * @param key Path or webhook key that contains the operation.
+   * @param method HTTP method of the operation to select.
+   * @param strategy How to combine the operation with an existing all-operation selection.
+   */
+  addOperation(key: string, method: string, strategy: 'narrow' | 'union' = 'narrow'): void {
```
erunion: Where is this `strategy` parameter being supplied?

packages/oas/src/lib/transformer/index.ts
erunion, line 1: Rename this file to `index.ts` and the directory it's in to `transformer`.

packages/oas/src/reducer/index.ts
```diff
@@ -144,638 +82,12 @@ export class OpenAPIReducer {
...
-
+    this.transformer.selectWebhook(webhookName, method);
     return this;
   }
 
-  /**
-   * Reduce the current OpenAPI definition down to the configured filters.
-   *
-   */
+  /** Reduce the current OpenAPI definition down to the configured filters. */
```
erunion: Can you retain the existing docblock formatting that was present?

```suggestion
  /** 
   * Reduce the current OpenAPI definition down to the configured filters.
   * 
   */
```

packages/oas/test/lib/openapi-transformer/operation-selection.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('can preserve an entire selection when adding individual operations', () => {
+    const selection = new OperationSelection();
+
+    selection.addAll('/pets');
+    selection.addOperation('/pets', 'get', 'union');
+
+    expect(selection.matches('/pets', 'get')).toBe(true);
+    expect(selection.matches('/pets', 'post')).toBe(true);
+  });
```
erunion: Can you add a falsy assertion to this test?

packages/oas/src/reducer/index.ts
```diff
@@ -1,65 +1,12 @@
...
-  private webhooksToReduceBy: Record<string, '*' | string[]> = {};
-
-  private hasTagsToReduceBy: boolean = false;
-  private hasPathsToReduceBy: boolean = false;
-  private hasWebhooksToReduceBy: boolean = false;
+  private transformer: OpenAPITransformer;
 
   private constructor(definition: OASDocument) {
-    this.definition = structuredClone(definition);
+    this.transformer = new OpenAPITransformer(definition);
```
erunion: Instead of creating a new instance of this class and assigning it into the `Reducer` instance, can `Reducer` just extend the transformer?

erunion (changes requested):

packages/oas/src/lib/openapi-transformer/operation-selection.ts
HugoHSun: It was later decided to remove this parameter and add a `clear` operation here. `OpenAPITransformer` would narrow it when in `reduce` mode. Have cleaned up the history to show it here

https://github.com/readmeio/oas/blob/8fcd134d3e4044f2a7aa03862135294048fd1aea/packages/oas/src/lib/transformer/index.ts#L103-L109

packages/oas/src/reducer/index.ts
HugoHSun: Changed to inheritance with protected methods in `OpenAPITransformer`

packages/oas/test/reducer/index.test.ts
```diff
@@ -29,6 +29,27 @@ describe('OpenAPIReducer', () => {
...
+  it('should narrow whole path and webhook selections to individual operations', async () => {
+    const reducedPath = OpenAPIReducer.init(petstore as OASDocument)
+      .byPath('/store/order/{orderId}')
+      .byOperation('/store/order/{orderId}', 'get')
+      .reduce();
+
+    const reducedWebhook = OpenAPIReducer.init(webhooks as OASDocument)
+      .byWebhook('newPet')
+      .byWebhook('newPet', 'post')
+      .reduce();
```
erunion, line 41: What behavior is this test case asserting? Is it testing reducing by the entire path/webhook, or just the single operation?

erunion (changes requested):

packages/oas/test/reducer/index.test.ts
HugoHSun: It's to make the existing chaining behaviour explicit: selecting a whole path then only a specific operation under it would narrow it to the specific operation when reducing

erunion (approved):

